PR #14, opened 2015-03-09
YARN-3126. FairScheduler: queue's usedResource is always more than the maxResource limit.

Xia-Hu:
This change add a check, for whether queue's usedResource may run over its maxResource Limit after a new resource allocation.


elek:

PR #21, opened 2015-06-10
[HADOOP-12081] Support Hadoop on zLinux - fix JAAS authentication issue

a-roberts:
Currently the 64 bit check in security/UserGroupInformation.java uses os.arch and checks for "64". s390x is returned on IBM's z platform: s390x is 64 bit. Without this change, if we try to use HDFS with Spark, we get a fatal error (unable to login as we can't find a login class).
This address fixes said issue by identifying s390x as a 64 bit platform and thus allowing Spark to run on zLinux. A simple fix with very big implications!

I've tested this builds and works as expected on zLinux.


aajisaka:
I've committed the patch in https://issues.apache.org/jira/browse/HADOOP-12081 to trunk and branch-2. Would you close this issue?

PR #30, opened 2015-08-25
update Shell.java to avoid OOM

huahuiyang:
avoid out of memory(such as NodeManager OOM), in case running a command which has verbose error log continuously


elek:

PR #36, opened 2015-10-10
Later hadoop use flat buffer

hash-X:


elek:

PR #39, opened 2015-10-29
HADOOP-12527. Upgrade Avro dependency to 1.7.7.

ejono:


benmccann:
Also upgrade avro-maven-plugin? https://github.com/apache/hadoop/blob/trunk/pom.xml#L225

ejono:
@benmccann, is that actually necessary for this patch? I don't actually know. Also, that line you point to currently has an open-ended max version, so it's already going to use the latest, right?

benmccann:
No, I don't think it's necessary. It just seems like it'd be good to do at some point

elek:

PR #41, opened 2015-11-02
CheckingTheChanges

Shubh91:
check


hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/permission/AclStatus.java
```diff
@@ -73,6 +75,14 @@ public boolean isStickyBit() {
     return entries;
   }
 
+  /**
+   * Returns the permission set for the path
+   * @return {@link FsPermission} for the path
+   */
+  public FsPermission getPermission() {
```
bk201sama: In hadoop 2.6.0 ,which method can replace it?

elek:

aajisaka:
Invalid. Closing.

PR #42, opened 2015-11-05
AltFileInputStream.java replace FileInputStream.java in apache/hadoop/HDFS

hash-X:
A brief description
Long Stop-The-World GC pauses due to Final Reference processing are observed.
So, Where are those Final Reference come from ?

1 : `Finalizer`
2 : `FileInputStream`

How to solve this problem ?

Here is the detailed description,and I give a solution on this.
https://issues.apache.org/jira/browse/HDFS-8562

FileInputStream have a method of finalize , and it can cause GC pause for a long time.In our test,G1 as our GC. So,in AltFileInputStream , no finalize. A new design for a inputstream use in windows and non-windows.


elek:

PR #50, opened 2015-11-22
HADOOP-11601 FileStatus.getBlocksize() >0 for non-empty files

steveloughran:
Enhance FS spec & tests to mandate FileStatus.getBlocksize() >0 for non-empty files


liuml07:
+1

Will commit this week.

PR #53, opened 2015-11-22
HDFS-9263

steveloughran:


hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
```diff
@@ -59,6 +59,22 @@
...
+  /**
+   * system property for test data: {@value}
+   */
+  public static final String SYSPROP_TEST_DATA_DIR = "test.build.data";
+
+  /**
+   * Default path for test data: {@value}
+   */
+  public static final String DEFAULT_TEST_DATA_DIR =
+      "target " + File.pathSeparator + "test" + File.pathSeparator + "data";
```
cnauroth: I think you'll still need to incorporate the fixes I suggested earlier in a JIRA comment: remove the extra space character at the end of the `"target "` string literal, and switch from `File.pathSeparator` to `File.separator`.

PR #58, opened 2015-11-24
HDFS-9459. hadoop-hdfs-native-client fails test build on Windows afte…

cnauroth:
…r transition to ctest.

I verified that the hadoop-hdfs-native-client tests build correctly and run successfully after applying this patch.


aajisaka:
Now HDFS-9459 is fixed.
@cnauroth Would you close this pull request?

PR #61, opened 2015-12-07
YARN-4430 registry security validation can fail when downgrading to insecure would work

steveloughran:
This makes the check -> warn, leaving ZK to decide how to handle the problem


elek:

PR #62, opened 2015-12-08
YARN-4434.NodeManager Disk Checker parameter documentation is not cor…

bwtakacy:
…rect


aajisaka:
Thank you for the pull request. I reviewed the patch (A) and the another patch in YARN-4434 jira (B) and decided to commit the patch (B) because the patch (B) replaces "i.e. the entire disk" with "i.e. 90% of the disk" as well.

aajisaka:
I've committed the patch (B), so would you close this pull request?

bwtakacy:
OK.
I will close this PR.

Thanks!

PR #79, opened 2016-02-20
for archive.org

h4ck3rm1k3:
I now have access to archive.org working. 
See my test case here : https://github.com/h4ck3rm1k3/hadoop-archive-org-bucket-fs

new URI("s3://bucketname").getHost()
is returning null, and I wonder how this ever worked.


h4ck3rm1k3:
I am having some problems with further testing. Closing this PR for now until I fix and test more.

h4ck3rm1k3:
https://github.com/h4ck3rm1k3/hadoop-archive-org-bucket-fs/issues/1 this is the problem that I found. Needs more work. Need to create test cases inside of hadoop for this.

PR #81, opened 2016-02-29
Shell.java: fix absolute path `/bin/ls`

makefu:
This commit uses `ls` from PATH instead of relying on `ls` being stored in `/bin/`. The only file according to the POSIX standard which must be stored in `/bin/` is `sh`.
This fixes issues plaguing distributions like NixOS which dynamically stitch together a PATH.

This fix is loosly related to https://issues.apache.org/jira/browse/HADOOP-11935 and more specifically related to https://mail-archives.apache.org/mod_mbox/hadoop-user/201512.mbox/%3CCAH2nEUgsSeoJTJkD4T8z=D18ECnRgQ3Qvz861gU5+NPSsNgT=A@mail.gmail.com%3E


makefu:
hi all, any comments on my PR?

PR #82, opened 2016-03-04
Disable setsid in case of Security Exceptions

costin:
If spawning an external process is not permitted by the JVM security manager, disable setsid.


elek:

PR #84, opened 2016-03-15
Update netty to 3.7.1.Final and netty-all to 4.0.34.Final

ceefour:


steveloughran:
Please can you file a separate JIRA for these two changes, link to HADOOP-9991 and include justification. Version updates are a traumatic issue in Hadoop and done fairly reluctantly.

FWIW, updating netty-all from beta to final makes sense just from a release perspective; updating the other jetty less so

ceefour:
@steveloughran This pull request now only updates netty-all (minor version update).

steveloughran:
sorry, I should have been clearer
-create a new Hadoop JIRA for this, mark it as a dependency of HADOOP-9991. thanks

ceefour:
@steveloughran created HADOOP-12927

elek:

PR #86, opened 2016-03-26
HADOOP-12916

xiaoyuyao:
Create a github PR for code review of v04 patch on the JIRA.

### DIFF
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java
@@ -1626,6 +1626,10 @@ public class Configuration implements Iterable<Map.Entry<String,String>>,
       return defaultValue;
     }
     vStr = vStr.trim();
+    return getTimeDurationHelper(name, vStr, unit);
+  }
+
+  private long getTimeDurationHelper(String name, String vStr, TimeUnit unit) {
     ParsedTimeDuration vUnit = ParsedTimeDuration.unitFor(vStr);
     if (null == vUnit) {
       LOG.warn("No unit for " + name + "(" + vStr + ") assuming " + unit);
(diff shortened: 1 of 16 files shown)


hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java
```diff
@@ -1636,6 +1640,15 @@ public long getTimeDuration(String name, long defaultValue, TimeUnit unit) {
     return unit.convert(Long.parseLong(vStr), vUnit.unit());
   }
 
+  public long[] getTimeDurations(String name, TimeUnit unit) {
```
arp7: Consider returning a List.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/ipc/CallQueueManager.java
```diff
@@ -49,25 +57,68 @@
...
   }
 
+  private static <T extends RpcScheduler> T createScheduler(
+      Class<T> theClass, int priorityLevels, String ns, Configuration conf) {
+    // Used for custom, configurable scheduler
+    try {
+      Constructor<T> ctor = theClass.getDeclaredConstructor(int.class,
+          String.class, Configuration.class);
+      return ctor.newInstance(priorityLevels, ns, conf);
+    } catch (RuntimeException e) {
```
arp7: See HDFS-9478 which is fixing exception handling when constructing callqueue instances. We could use a similar fix for createScheduler.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java
xiaoyuyao: Good point. I will address that in the next patch.

elek:

aajisaka:
This issue has been fixed. Closing.

PR #89, opened 2016-04-03
HADOOP-12984. Add GenericTestUtils.getTestDir method and use it for emporary directory in tests

vinayakumarb:
Rebased.

### DIFF
--- a/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/TestFileUtil.java
+++ b/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/TestFileUtil.java
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.fs;
 
+import org.apache.hadoop.test.GenericTestUtils;
 import org.junit.*;
 
 import java.io.File;
@@ -55,9 +56,7 @@ import static org.mockito.Mockito.when;
 public class TestFileUtil {
   private static final Log LOG = LogFactory.getLog(TestFileUtil.class);
 
-  private static final String TEST_ROOT_DIR = System.getProperty(
-      "test.build.data", "/tmp") + "/fu";
-  private static final File TEST_DIR = new File(TEST_ROOT_DIR);
+  private static final File TEST_DIR = GenericTestUtils.getTestDir("fu");
   private static final String FILE = "x";
   private static final String LINK = "y";
   private static final String DIR = "dir";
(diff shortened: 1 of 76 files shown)


hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/server/namenode/TestDecommissioningStatus.java
```diff
@@ -87,7 +87,7 @@ public static void setUp() throws Exception {
     // Set up the hosts/exclude files.
     localFileSys = FileSystem.getLocal(conf);
     Path workingDir = localFileSys.getWorkingDirectory();
-    dir = new Path(workingDir, "build/test/data/work-dir/decommission");
+    dir = new Path(workingDir, "target/test/data" + "work-dir/decommission");
```
aajisaka: I'm thinking the string concatenation by `+` is unnecessarily.

aajisaka:
Thank you for rebasing! Mostly looks good to me. Two comments:
* Can we separate the patch into common change and hdfs change? That way the patch becomes smaller and the review becomes easier.
* I'm thinking we should replace not only `test.build.data` but also `test.build.dir` with `GenericTestUtils.getTestDir`. I'm okay with doing this by separate jira(s).

vinayakumarb:
Rebased,
and separated the common changes and hdfs changes.
Pushed only common changes in this PR.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/TestFileUtil.java
```diff
@@ -55,9 +56,11 @@
 public class TestFileUtil {
   private static final Log LOG = LogFactory.getLog(TestFileUtil.class);
 
-  private static final String TEST_ROOT_DIR = System.getProperty(
-      "test.build.data", "/tmp") + "/fu";
+  private static final String TEST_ROOT_DIR =
+      new File(GenericTestUtils.getTestDir(), "fu").getAbsolutePath();
   private static final File TEST_DIR = new File(TEST_ROOT_DIR);
+  private static final String  cacheDir = System.getProperty("test.cache.data",
+      new File(TEST_DIR, "cache").getAbsolutePath());
```
aajisaka: * The code creates String by `File.getAbsolutePath()` and then creates File by `new File(String)`. It can be simplified by the following and `TEST_ROOT_DIR` can be removed.
```
 private static final File TEST_DIR = new File(GenericTestUtils.getTestDir(), "fu");
```

* `cacheDir` is unused.

aajisaka:
Thank you for the update! I'm +1 if the above comments are addressed.

PR #90, opened 2016-04-04
HDFS-10256. Use GenericTestUtils.getTestDir method in tests for temporary directories

vinayakumarb:
Separated the HDFS changes from HADOOP-12984


hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/fs/TestWebHdfsFileContextMainOperations.java
```diff
@@ -63,7 +64,7 @@ public Path getDefaultWorkingDirectory() {
 
   @Override
   protected FileContextTestHelper createFileContextHelper() {
-    return new FileContextTestHelper("/tmp/TestWebHdfsFileContextMainOperations");
+    return new FileContextTestHelper();
```
aajisaka: I found some other occurrences of `/tmp/xxx`, which is used as the root dir of `FileContextTestHelper`. Would you replace them? I'm thinking the following is fine:
```
return new FileContextTestHelper(GenericTestUtils.getTempPath("TestWebHdfsFileContextMainOperations"));
```
I'm okay if the replace is done in separate jira(s).

aajisaka:
Mostly looks good to me. Would you fix the checkstyle warnings? I'm +1 if that is addressed.
(I commented here because ASF JIRA is down now)

vinayakumarb:
Merged to trunk.

PR #92, opened 2016-04-22
HDFS-9895. Push up DataNode#conf to base class

xiaobingo:
Please kindly review the patch v001, see also https://issues.apache.org/jira/browse/HDFS-9895.

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DNConf.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DNConf.java
@@ -70,7 +70,6 @@ import org.apache.hadoop.security.SaslPropertiesResolver;
  */
 @InterfaceAudience.Private
 public class DNConf {
-  final Configuration conf;
   final int socketTimeout;
   final int socketWriteTimeout;
   final int socketKeepaliveTimeout;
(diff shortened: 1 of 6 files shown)


hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DNConf.java
```diff
@@ -113,71 +112,71 @@
...
   private final boolean allowNonLocalLazyPersist;
-
+  private final DataNode dn;
   private final int volFailuresTolerated;
   private final int volsConfigured;
 
-  public DNConf(Configuration conf) {
-    this.conf = conf;
-    socketTimeout = conf.getInt(DFS_CLIENT_SOCKET_TIMEOUT_KEY,
+  public DNConf(final DataNode dn) {
```
arp7: The dn.getConf() object is not referenced outside the constructor so you can just pass a reference to that object. Also DNConf need not keep a reference to the dn. I think you can just revert all changes to this file.

hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DataNode.java
```diff
@@ -1239,9 +1236,8 @@ void startDataNode(Configuration conf,
     synchronized (this) {
       this.dataDirs = dataDirs;
     }
-    this.conf = conf;
-    this.dnConf = new DNConf(conf);
-    checkSecureConfig(dnConf, conf, resources);
+    this.dnConf = new DNConf(this);
```
arp7: If we revert changes to DNConf we can just replace this with `this.dnConf = new DNConf(getConf())`.

elek:

aajisaka:
https://issues.apache.org/jira/browse/HDFS-9895 has been fixed. Closing.

PR #94, opened 2016-05-10
HDFS-10382 In WebHDFS numeric usernames do not work with DataNode

ramtinb:
In WebHDFS for cat operation, we have 2 sequential HTTP requests.
The first HTTP request is handled by NN and the second one by DN.
Unlike the NN, the DN is not using the suggested domain pattern from the configuration!


elek:

PR #99, opened 2016-06-13
HADOOP-13192. org.apache.hadoop.util.LineReader  match recordDelimiter has a bug

zhudebin:
fix bug HADOOP-13192
org.apache.hadoop.util.LineReader  match recordDelimiter has a bug


aajisaka:
I've committed this. Would you close this pull request?

zhudebin:
ok

PR #103, opened 2016-06-22
HADOOP-12345,HADOOP-11823 fixes

pradeep1288:
I have already filed the HADOOP-12345 and HADOOP-11823 jira. 

The fix for HADOOP-12345 is to correctly compute the credential length which is passed as part of the NFS request. We need to round the XDR bytes to a multiple of 4 if the hostname in the credential is not a multiple of 4.

The fix for HADOOP-11823 is when RPC returns a denied reply, the code should not check for a verifier. It is a bug as it doesn't match the RPC protocol. (See Page 33 from NFS Illustrated book).
 

This is my first time contributing to Hadoop.


pradeep1288:
I will separate out the changes for HADOOP-12345 and HADOOP-11823 and submit two pull requests

PR #104, opened 2016-06-23
HADOOP-12345: Compute the correct credential length

pradeep1288:
I had to discard my earlier pull request as that included fix for the jira HADOOP-11823, hence creating a separate one for each of them.

The fix here addresses the correct credential length to be computed. We need to round of the machine name length to the next multiple of 4, else using such a credential in the NFS RPC request will result in GARBAGE_ARGS from the NFS server. See RFC-5531, page 24 and page 8


pradeep1288:
This has been checked in

PR #108, opened 2016-07-01
HADOOP-13320. Fix arguments check in the WordCount v2.0 in Documentation

pippobaudos:


templedf:
Looks like the right fix, but can you simplify the logic a little, i.e. {{(len != 2) && (len != 4))}}?

pippobaudos:
Hi @templedf I have updated the commit. Now should be the simplest expression to read...

pippobaudos:
Thanks @templedf  I have updated the pull request following the suggestion

PR #113, opened 2016-07-11
[HADOOP-13075] Adding support for SSE-KMS and SSE-C

fedecz:


hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Constants.java
```diff
@@ -142,12 +142,8 @@ private Constants() {
   public static final String SERVER_SIDE_ENCRYPTION_ALGORITHM =
       "fs.s3a.server-side-encryption-algorithm";
 
-  /**
-   * The standard encryption algorithm AWS supports.
-   * Different implementations may support others (or none).
-   */
-  public static final String SERVER_SIDE_ENCRYPTION_AES256 =
```
steveloughran: why was this cut?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -1674,6 +1693,111 @@ public void progressChanged(ProgressEvent progressEvent) {
...
+  }
+
+  protected void setSSECIfRequired(GetObjectMetadataRequest request) {
+    if (StringUtils.isNotBlank(serverSideEncryptionAlgorithm)){
+      if(S3AEncryptionMethods.SSE_C.getMethod()
+          .equals(serverSideEncryptionAlgorithm)) {
+        if (StringUtils.isNotBlank(serverSideEncryptionKey)) {
+          //at the moment, only supports copy using the same key
+          request.setSSECustomerKey(
+              new SSECustomerKey(serverSideEncryptionKey)
```
steveloughran: this is three chained conditions which could be merged through `&&`

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
```diff
@@ -98,24 +101,33 @@
...
       S3AInstrumentation instrumentation,
       long readahead,
       S3AInputPolicy inputPolicy) {
-    Preconditions.checkArgument(StringUtils.isNotEmpty(bucket), "No Bucket");
-    Preconditions.checkArgument(StringUtils.isNotEmpty(key), "No Key");
-    Preconditions.checkArgument(contentLength >= 0 , "Negative content length");
-    this.bucket = bucket;
-    this.key = key;
+    Preconditions.checkNotNull(s3Attributes);
+    Preconditions.checkArgument(
```
steveloughran: could you move the preconditions checks into `S3ObjectAttributes` and invoke them from both output streams? That'd reduce duplicate code and perhaps aid future maintenance

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/TestS3AEncryption.java
```diff
@@ -27,23 +27,24 @@
 
 import java.io.IOException;
 
-import static org.apache.hadoop.fs.contract.ContractTestUtils.*;
-import static org.apache.hadoop.fs.s3a.S3ATestUtils.*;
+import static org.apache.hadoop.fs.contract.ContractTestUtils.dataset;
```
steveloughran: we're ok with .* on static imports here, so you can just skip this bit of the patch

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
```diff
@@ -940,6 +950,10 @@ the DNS TTL of a JVM is "infinity".
 To work with AWS better, set the DNS time-to-live of an application which
 works with S3 to something lower. See [AWS documentation](http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/java-dg-jvm-ttl.html).
 
+*internal.S3V4AuthErrorRetryStrategy (S3V4AuthErrorRetryStrategy.java:buildRetryParams(117)) - Attempting to re-send the request to...*
```
steveloughran: looks like an accidental paste in of a bit of HADOOP-13224's doc changes

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Constants.java
fedecz: As you said, that constant is only used in that test which I did change. I changed it to abstract and created 3 different implementations: one for SSE-S3, SSE-KMS and SSE-C. Basically I'm running all the tests in TestS3AEncryption, but with different encryption algorithms depending on the concrete class.
Yes, it builds and all test pass.

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
fedecz: Nope, this is the section _Other Issues_ in the documentation, so I wanted to document if the user was having that warning, he/she should specify the endpoint in the config. I guess I could set a title to describe it better instead of just pasting the warning.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
fedecz: true, but not all of them can be merged. I'm relying in else clauses as well depending on some of the conditions being false. I'll try to rewrite it though and will see how it looks.

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
steveloughran: Well it's being covered in HADOOP-13224, so it's best to pull it here and review that patch instead

fedecz: I don't see anything related to this patch in that ticket's patch, are you sure that's the one? I'm looking at the attached patch in HADOOP-13224

steveloughran: sorry, wrong JIRA. https://issues.apache.org/jira/browse/HADOOP-13324

elad:
Hello, what's the status of this patch? Would be really great if we could use S3 SSE-C with Hadoop on EMR. :)

steveloughran:
@elad best to discuss that on the JIRA itself; FWIW, EMR's S3 connector is Amazon's codebase, so what we do in the ASF doesn't have any impact there. We are striving to be better :-)

Status: we've gone and broken it in Hadoop branch-2 by replacing {{S3AFastOutputStream}} with {{S3ABlockOutputStream}}; 

@fedecz —could you bring this up to sync with branch-2 and I'll add reviewing it to my todo list. thanks

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Constants.java
steveloughran: we can't pull values out of public interfaces like this, even if its not supported. They'd need to be tagged as deprecated

elek:

aajisaka:
HADOOP-13075 has been fixed. Closing.

PR #114, opened 2016-07-11
MAPREDUCE-6730 Use StandardCharsets instead of String overload

SahilKang:
Use String.getBytes(StandardCharsets.UTF_8) instead of String.getBytes(String).


aajisaka:
Hi @SahilKang, thank you for reporting this issue and creating the patch.
* Would you fix the checkstyle warnings? https://builds.apache.org/job/PreCommit-MAPREDUCE-Build/6651/artifact/patchprocess/diff-checkstyle-hadoop-mapreduce-project_hadoop-mapreduce-client_hadoop-mapreduce-client-core.txt
* Would you fix org.apache.hadoop.mapred.TextOutputFormat as well?

SahilKang:
@aajisaka, this latest commit should fix the two checkstyle warnings, and I'll send another patch soon with the analogous changes made to org.apache.hadoop.mapred.TextOutputFormat.

aajisaka:
+1 pending Jenkins. Thanks @SahilKang for the update.

PR #124, opened 2016-08-31
YARN-5605. Preempt containers (all on one node) to meet the requirement of starved applications.

kambatla:
- Remove existing preemption code.
- Deleted preemption tests, to be reinstated later.
- Identify starved applications in the update thread when preemption is enabled.
- For each starved application, identify containers to be preempted on a single node to make room for this request.
- Preempt these containers after necessary warning.


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/AppSchedulingInfo.java
```diff
@@ -535,6 +535,23 @@ public synchronized Resource getResource(SchedulerRequestKey schedulerKey) {
...
+   * In the initial implementation, we just pick any {@link ResourceRequest}
+   * corresponding to the highest priority.
+   *
+   * @return next {@link ResourceRequest} to allocate resources for.
+   */
+  @Unstable
+  public synchronized ResourceRequest getNextResourceRequest() {
+    for (ResourceRequest rr:
+        resourceRequestMap.get(schedulerKeys.first()).values()) {
+      return rr;
```
templedf: Can we please have one exit point?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSAppAttempt.java
```diff
@@ -442,25 +448,33 @@ public synchronized void resetAllowedLocalityLevel(
...
   }
 
-  // related methods
-  public void addPreemption(RMContainer container, long time) {
-    assert preemptionMap.get(container) == null;
-    preemptionMap.put(container, time);
-    Resources.addTo(preemptedResources, container.getAllocatedResource());
+  @Override
+  public FSLeafQueue getQueue() {
+    return (FSLeafQueue)super.getQueue();
```
templedf: Seems like this should be more defensive, i.e. check for the type before casting.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSAppAttempt.java
```diff
@@ -478,6 +492,31 @@ public void clearPreemptedResources() {
...
+      if (!queue.isPreemptable()) {
+        return false;
+      }
+    }
+
+    // Check if the app's allocation will be over its fairshare even
+    // after preempting this container
+    return (Resources.fitsIn(container.getAllocatedResource(),
+        Resources.subtract(getResourceUsage(), getFairShare())));
+  }
```
templedf: Indentation

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -45,16 +45,19 @@
...
 import org.apache.hadoop.yarn.util.resource.Resources;
 
+import static org.apache.hadoop.yarn.util.resource.Resources.none;
+
 @Private
 @Unstable
 public class FSLeafQueue extends FSQueue {
   private static final Log LOG = LogFactory.getLog(
       FSLeafQueue.class.getName());
+  private FairScheduler scheduler;
```
templedf: Maybe put the scheduler into the context?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -64,19 +67,18 @@
   
   // Variables used for preemption
   private long lastTimeAtMinShare;
-  private long lastTimeAtFairShareThreshold;
-  
+
```
templedf: Spurious newline

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -223,17 +225,76 @@ public void setPolicy(SchedulingPolicy policy)
...
+  private void identifyStarvedApplications() {
+    // First identify starved applications and track total amount of
+    // starvation (in resources)
+    Resource fairShareStarvation = Resources.clone(none());
+    TreeSet<FSAppAttempt> appsWithDemand = fetchAppsWithDemand();
+    for (FSAppAttempt app : appsWithDemand) {
+      Resource appStarvation = app.fairShareStarvation();
+      if (Resources.equals(Resources.none(), appStarvation))  {
+        break;
+      } else {
```
templedf: Drop the else.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -223,17 +225,76 @@ public void setPolicy(SchedulingPolicy policy)
...
+
+    // Compute extent of minshare starvation
+    Resource minShareStarvation = minShareStarvation();
+
+    // Compute minshare starvation that is not subsumed by fairshare starvation
+    Resources.subtractFrom(minShareStarvation, fairShareStarvation);
+
+    // Keep adding apps to the starved list until the unmet demand goes over
+    // the remaining minshare
+    for (FSAppAttempt app : appsWithDemand) {
```
templedf: If minShareStarvation in 0, you're doing a bunch of looping for no reason.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -316,26 +377,12 @@ public Resource assignContainer(FSSchedulerNode node) {
...
-          pendingForResourceApps.add(app);
-        }
-      }
-    } finally {
-      readLock.unlock();
-    }
-    for (FSAppAttempt sched : pendingForResourceApps) {
+    for (FSAppAttempt sched : fetchAppsWithDemand()) {
       if (SchedulerAppUtils.isPlaceBlacklisted(sched, node, LOG)) {
         continue;
```
templedf: It would be nice to replace this _continue_ by wrapping the next few lines in the _if_.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -557,28 +599,33 @@ private boolean preemptContainerPreCheck() {
         getFairShare());
   }
 
-  /**
-   * Is a queue being starved for its min share.
-   */
-  @VisibleForTesting
-  boolean isStarvedForMinShare() {
-    return isStarved(getMinShare());
+  private Resource minShareStarvation() {
```
templedf: This seems convoluted.  why not:

boolean starved = false;
Resource starvation = none();
if (Resources.greaterThan(policy.getResourceCalculator(),
    scheduler.getClusterResource(), getDemand(), getMinShare())) {
  starvation = Resources.subtract(getMinShare(), getResourceUsage());
  starved = Resources.greaterThan(policy.getResourceCalculator(),
    scheduler.getClusterResource(), starvation, none());
}

Nets out to the same thing, but the logic isn't as hard (for me) to follow.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -557,28 +599,33 @@ private boolean preemptContainerPreCheck() {
...
+    long now = scheduler.getClock().getTime();
+    if (!starved) {
+      setLastTimeAtMinShare(now);
+    }
+
+    if (starved &&
+        (now - lastTimeAtMinShare > getMinSharePreemptionTimeout())) {
+      return starvation;
+    } else {
+      return Resources.clone(Resources.none());
```
templedf: Single exit point, please

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSPreemptionThread.java
```diff
@@ -0,0 +1,173 @@
...
+    while (!Thread.interrupted()) {
+      FSAppAttempt starvedApp;
+      try{
+        starvedApp = context.getStarvedApps().take();
+        if (Resources.none().equals(starvedApp.getStarvation())) {
+          continue;
+        }
+      } catch (InterruptedException e) {
+        LOG.info("Preemption thread interrupted! Exiting.");
+        return;
```
templedf: You should call interrupt();

templedf: You could also replace this return with starvedApp = null, and then put a guard at the beginning of identifyContainersToPreempt(), which is arguably a good thing to do in any case.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSParentQueue.java
```diff
@@ -80,13 +80,13 @@ public void removeChildQueue(FSQueue child) {
   }
 
   @Override
-  public void recomputeShares() {
+  public void updateInternal(boolean checkStarvation) {
     readLock.lock();
     try {
       policy.computeShares(childQueues, getFairShare());
       for (FSQueue childQueue : childQueues) {
         childQueue.getMetrics().setFairShare(childQueue.getFairShare());
```
templedf: Seems like this line should be pushed down into FSChildQueue.updateInternal()

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -223,17 +225,76 @@ public void setPolicy(SchedulingPolicy policy)
...
+   * Fairshare is defined for both the queue and the applications.
+   *
+   * If this queue is starved due to minshare, we need to identify the most
+   * deserving apps if they themselves are not starved due to fairshare.
+   *
+   * If this queue is starving due to fairshare, there must be at least
+   * one application that is starved. And, even if the queue is not
+   * starved due to fairshare, there might still be starved applications.
+   */
+  private void identifyStarvedApplications() {
```
templedf: I don't like this name.  It feels to me like an identifyX() should return something.  I'd rather have a name that says what it does, e.g. updateStarvedApplications().

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -223,17 +225,76 @@ public void setPolicy(SchedulingPolicy policy)
...
+    // First identify starved applications and track total amount of
+    // starvation (in resources)
+    Resource fairShareStarvation = Resources.clone(none());
+    TreeSet<FSAppAttempt> appsWithDemand = fetchAppsWithDemand();
+    for (FSAppAttempt app : appsWithDemand) {
+      Resource appStarvation = app.fairShareStarvation();
+      if (Resources.equals(Resources.none(), appStarvation))  {
+        break;
+      } else {
+        context.getStarvedApps().addStarvedApp(app);
```
templedf: Feels like FSContext should have a wrapper method for addStarvedApp().

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -223,17 +225,76 @@ public void setPolicy(SchedulingPolicy policy)
...
+   * If this queue is starving due to fairshare, there must be at least
+   * one application that is starved. And, even if the queue is not
+   * starved due to fairshare, there might still be starved applications.
+   */
+  private void identifyStarvedApplications() {
+    // First identify starved applications and track total amount of
+    // starvation (in resources)
+    Resource fairShareStarvation = Resources.clone(none());
+    TreeSet<FSAppAttempt> appsWithDemand = fetchAppsWithDemand();
+    for (FSAppAttempt app : appsWithDemand) {
```
templedf: Spurious space before the colon

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSPreemptionThread.java
```diff
@@ -0,0 +1,173 @@
...
+import org.apache.hadoop.yarn.api.records.ContainerStatus;
+import org.apache.hadoop.yarn.api.records.Resource;
+import org.apache.hadoop.yarn.api.records.ResourceRequest;
+import org.apache.hadoop.yarn.server.resourcemanager.rmcontainer.RMContainer;
+import org.apache.hadoop.yarn.server.resourcemanager.rmcontainer.RMContainerEventType;
+import org.apache.hadoop.yarn.server.resourcemanager.scheduler.SchedulerUtils;
+import org.apache.hadoop.yarn.util.resource.Resources;
+
+import java.util.ArrayList;
+import java.util.Comparator;
```
templedf: Unused import

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/AppSchedulingInfo.java
kambatla: Adding another variable and breaking out of the for loop seems more complicated that it is worth. Leaving it as is unless you insist.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
kambatla: Did that initially. That bloated up the patch quite a lot, and is somewhat orthogonal. Will file a follow-up JIRA for that work.

kambatla: Filed YARN-5625.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
kambatla: I am not clear on the convention. Always thought there should be a space before and after the colon. Online references seem to use both. Leaving it as is.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
kambatla: The else is required. We are iterating through all apps with unmet demand. Since the list is sorted by fairshare starvation, we could stop iterating when we hit an app that is at or above its fairshare. Added a comment to clarify that.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
kambatla: I was tempted to, but then thought FSContext is just a structure to hold context and should only have fields with getters and setters.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
kambatla: You are right that:

starvation = desiredShare - currentUsage;
starved = starvation > 0;

However, you are missing desiredShare = min(minShare, demand). Essentially, the missing case is when the demand is less than minShare but still more than current allocation. 

That said, I see a minor simplification. Will do that.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
kambatla: Adding a new variable and returning that seems like an overkill. Can't use the same variable or initialize to none() as that requires creating a new object. 

Leaving it as is.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSPreemptionThread.java
```diff
@@ -0,0 +1,173 @@
...
+    setDaemon(true);
+    setName("FSPreemptionThread");
+  }
+
+  public void run() {
+    while (!Thread.interrupted()) {
+      FSAppAttempt starvedApp;
+      try{
+        starvedApp = context.getStarvedApps().take();
+        if (Resources.none().equals(starvedApp.getStarvation())) {
```
kambatla: Good call! Added that.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSPreemptionThread.java
kambatla: FairScheduler#serviceStop calls preemptionThread.interrupt. Are you referring to that?

kambatla: Are you suggesting having the run method return a boolean?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
templedf: You can make it:

~~~~
    if (!starved ||
        (now - lastTimeAtMinShare < getMinSharePreemptionTimeout())) {
      starvation = Resources.clone(Resources.none());

      if (!starved) {
        setLastTimeAtMinShare(now);
      }
    }

    return starvation;
~~~~

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSPreemptionThread.java
templedf: I think I was a little confused.  How about this:

~~~~
      try{
        starvedApp = context.getStarvedApps().take();
        if (!Resources.none().equals(starvedApp.getStarvation())) {
          List<RMContainer> containers = identifyContainersToPreempt(starvedApp);
          if (containers != null) {
            preemptContainers(containers);
          }
        }
      } catch (InterruptedException e) {
        LOG.info("Preemption thread interrupted! Exiting.");
        interrupt();
      }
~~~~

It does some extra work inside the _try_, but the logic is much simpler.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
templedf: I think you should do whatever makes the code cleanest and easiest to maintain.  I don't think making the context a glorified hash map helps you in any notable way here.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/AppSchedulingInfo.java
templedf: It's stylistic, and there's no guideline about multiple exit points that I know of, so I won't push it.  I don't think this form is very future-safe, though.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
templedf: It might be clearer if you swapped the _if_ and _else_.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
kambatla: Actually, this can be simplified further. Improvement in updated patch.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSStarvedApps.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  private static class StarvationComparator implements
+      Comparator<FSAppAttempt>, Serializable {
+    private static final long serialVersionUID = 1;
+
+    @Override
+    public int compare(FSAppAttempt app1, FSAppAttempt app2) {
+      return Resources.fitsIn(app1.getStarvation(), app2.getStarvation())
+          ? -1 : 1;
```
templedf: Missed this before.  I'd much rather see:

~~~~
int ret = 1;

if (Resources.fitsIn(app1.getStarvation(), app2.getStarvation())) {
  ret = -1;
}

return ret;
~~~~

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -74,7 +74,8 @@
   private final ActiveUsersManager activeUsersManager;
   public static final List<FSQueue> EMPTY_LIST = Collections.emptyList();
 
-  public FSLeafQueue(String name, FairScheduler scheduler, FSParentQueue parent) {
+  public FSLeafQueue(
+      String name, FairScheduler scheduler, FSParentQueue parent) {
```
templedf: I'd prefer the line break after the first or second arg.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSStarvedApps.java
kambatla: I don't understand the reasons, but fixing it anyway.

PR #125, opened 2016-09-08
HADOOP-13560 S3A to support huge file writes and operations -with tests

steveloughran:
Adds 

## Scale tests for S3A huge file support; 
- always running at the MB size (maybe best to make optional)
 -configurable to bigger sizes in the auth-keys XML or in the build `-Dfs.s3a.scale.test.huge.filesize=1000`
- limited to upload, seek, read, rename, delete. The JUnit test cases are explicltly set up to run in order here.

## New scalable output stream for writing, `S3ABlockOutputStream`

-always saves in incremental blocks as writes proceed, block size == partition size.
-supports Fast output stream memory buffer code (for regression testing)
-supports a back end which buffers blocks in files, using RR disk allocation. As such, write/read bandwidth is limited to aggregate HDD bandwidth.
-adding extra failure resilience as testing throws up failure conditions (network timeouts, no-response from server on multipart commit, etc).
-adding instrumentation, including using callbacks from AWS SDK to update gauges and counters (in progress)

What we have here is essentially something that can replace the classic "save to file, upload at the end" stream and the fast "store it all in RAM and hope there's space" stream. It should offer incremental upload for faster output of larger files compared the classic file stream, with the scaleability the fast one lacks. And the instrumentation to show what's happening.


hadoop-tools/hadoop-aws/pom.xml
```diff
@@ -183,6 +199,8 @@
                     <include>**/ITestS3AFileSystemContract.java</include>
                     <include>**/ITestS3AMiniYarnCluster.java</include>
                     <include>**/ITest*Root*.java</include>
+                    <include>**/ITestS3AFileContextStatistics.java</include>
```
steveloughran: moved this line down as it was failing sometimes

PR #126, opened 2016-09-08
HADOOP-13341: Deprecate HADOOP_SERVERNAME_OPTS; replace with (command)_(subcommand)_OPTS

aw-was-here:


hadoop-common-project/hadoop-common/src/site/markdown/UnixShellGuide.md
```diff
@@ -24,14 +24,26 @@ Apache Hadoop has many environment variables that control various aspects of the
...
+
+It is also possible to set options on a per subcommand basis.  This allows for one to create special options for particular cases.  The first part of the pattern is the command being used, but all uppercase.  The second part of the command is the subcommand being used.  Then finally followed by the string `_OPT`.
+
+For example, to configure `mapred distcp` to use a 2GB heap, one would use:
+
+```bash
+MAPRED_DISTCP_OPTS="-Xmx2g"
+```
+
+These options will appear *after* `HADOOP_CLIENT_OPTS` during execution and will generally take precedence.
```
steveloughran: might be good to add an example here. e.g what the final options of distcp are going to be. Will there be two -Xmx commands? if so, which wins? Because I suspect that's a JVM decision

aw-was-here: If there is an Xmx in HADOOP_CLIENT_OPTS and an Xmx in MAPRED_DISTCP_OPTS, then the mapred distcp final HADOOP_OPTS will definitely have two Xmx flags.  After HADOOP-13365, we'll be in a position to potentially de-dupe user provided settings like we do for other things.  But until de-dupe, you're correct that it's a JVM decision.  In the past, that decision has been last one wins and I doubt Oracle could change it if they wanted to at this point without major ramifications.

PR #130, opened 2016-09-23
HADOOP-13560 block output streams

steveloughran:
Merge commit of latest code. 

Docs, XML configs up to speed
scale tests only run with a -Pscale option.
some props can be configured in POM


hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
```diff
@@ -1250,6 +1569,144 @@ can be used:
 Using the explicit endpoint for the region is recommended for speed and the
 ability to use the V4 signing API.
 
+
+## "Timeout waiting for connection from pool" when writing to S3A
```
cnauroth: I tried an `mvn site` build, and it looks like the new troubleshooting sections still aren't nested correctly.  I believe it should be `###` instead of `##`.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ABlockOutputStream.java
```diff
@@ -0,0 +1,699 @@
...
+    this.statistics = statistics;
+    this.writeOperationHelper = writeOperationHelper;
+    Preconditions.checkArgument(blockSize >= Constants.MULTIPART_MIN_SIZE,
+        "Block size is too small: %d", blockSize);
+    this.executorService = MoreExecutors.listeningDecorator(executorService);
+    this.multiPartUpload = null;
+    this.progressListener = (progress instanceof ProgressListener) ?
+        (ProgressListener) progress
+        : new ProgressableListener(progress);
+    LOG.debug("Initialized S3ABlockOutputStream for {}" +
```
cnauroth: I think `activeBlock` is always `null` when this log statement executes.

hadoop-common-project/hadoop-common/src/main/resources/core-default.xml
```diff
@@ -1093,12 +1101,48 @@
 </property>
 
 <property>
-  <name>fs.s3a.fast.upload</name>
+  <name>fs.s3a.block.output</name>
```
cnauroth: Is this revision missing the changes to restore/un-deprecate `fs.s3a.fast.upload`?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ADataBlocks.java
```diff
@@ -0,0 +1,816 @@
...
+      return dataSize != null ? dataSize : buffer.size();
+    }
+
+    @Override
+    InputStream startUpload() throws IOException {
+      super.startUpload();
+      dataSize = buffer.size();
+      ByteArrayInputStream bufferData = new ByteArrayInputStream(
+          buffer.toByteArray());
+      buffer.reset();
```
steveloughran: OK

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ADataBlocks.java
```diff
@@ -0,0 +1,816 @@
...
+        // it is here.
+      }
+    }
+
+    /**
+     * Flush operation will flush to disk.
+     * @throws IOException IOE raised on FileOutputStream
+     */
+    @Override
+    void flush() throws IOException {
```
steveloughran: good catch

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
steveloughran: did that everywhere; updated. Also in the troubleshooting s3a memory, just pointed back to the thread tuning entry.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ABlockOutputStream.java
steveloughran: correct! swapped order of log and action

hadoop-common-project/hadoop-common/src/main/resources/core-default.xml
steveloughran: afraid so, that bit had missed the push as I forgot to --force the patch up at the end of the day. Will push it up with all the comments here, after another test run

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ADataBlocks.java
```diff
@@ -0,0 +1,811 @@
...
+    public void close() throws IOException {
+    }
+  }
+
+  /**
+   * This represents a block being uploaded.
+   */
+  static abstract class DataBlock implements Closeable {
+
+    private volatile DestState state = Writing;
```
pieterreuse: I'm nitpicking here, but wouldn't it make more sense to define DestState here instead of on line 272? Moving that line here would improve code readability imo but wouldn't change any behaviour.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ABlockOutputStream.java
```diff
@@ -0,0 +1,699 @@
...
+  /**
+   * An S3A output stream which uploads partitions in a separate pool of
+   * threads; different {@link S3ADataBlocks.BlockFactory}
+   * instances can control where data is buffered.
+   *
+   * @param fs S3AFilesystem
+   * @param key S3 object to work on.
+   * @param executorService the executor service to use to schedule work
+   * @param progress report progress in order to prevent timeouts. If
+   * this class implements {@code ProgressListener} then it will be
```
pieterreuse: This method is passed an object, not a class. You probably meant "If this _object_ implements ..."

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ABlockOutputStream.java
```diff
@@ -0,0 +1,699 @@
...
+    LOG.debug("Initialized S3ABlockOutputStream for {}" +
+        " output to {}", writeOperationHelper, activeBlock);
+  }
+
+  /**
+   * Demand create a destination block.
+   * @return the active block; null if there isn't one.
+   * @throws IOException on any failure to create
+   */
+  private synchronized S3ADataBlocks.DataBlock maybeCreateBlock()
```
pieterreuse: The lazy creation in this method is nice, but the "maybe" in its name gives a false impression of arbitrariness involved. "createBlockIfNeeded" might be a better naming option.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ADataBlocks.java
steveloughran: done

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ABlockOutputStream.java
steveloughran: renamed

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
```diff
@@ -881,40 +881,362 @@ Seoul
...
+and, if disk space is small (for example, tiny EC2 VMs), there may not
+be much disk space to buffer with.
+
+The ByteBuffers are created in the memory of the JVM, but not in the Java Heap itself.
+The amount of data which can be buffered is
+limited by the Java runtime, the operating system, and, for YARN applications,
+the amount of memory requested for each container.
+
+The slower the write bandwidth to S3, the greater the risk of running out
+of memory.
```
thodemoor: Memory usage is bounded to ...

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
```diff
@@ -881,40 +881,362 @@ Seoul
...
+    Both "array" and "bytebuffer" will consume memory in a single stream up to the number
+    of blocks set by:
+
+        fs.s3a.multipart.size * fs.s3a.fast.upload.active.blocks.
+
+    If using either of these mechanisms, keep this value low
+
+    The total number of threads performing work across all threads is set by
+    fs.s3a.threads.max, with fs.s3a.max.total.tasks values setting the number of queued
+    work items.
```
thodemoor: idem as in pom.xml

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
```diff
@@ -881,40 +881,362 @@ Seoul
...
+When `fs.s3a.fast.upload.buffer` is set to `array`, all data is buffered
+in byte arrays in the JVM's heap prior to upload.
+This *may* be faster than buffering to disk.
+
+This `array` option is similar to the in-memory-only stream offered in
+Hadoop 2.7 with `fs.s3a.fast.upload=true`
+
+The amount of data which can be buffered is limited by the available
+size of the JVM heap heap. The slower the write bandwidth to S3, the greater
+the risk of heap overflows.
```
thodemoor: idem

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
```diff
@@ -881,40 +881,362 @@ Seoul
...
+</property>
+  
+<property>
+  <name>fs.s3a.fast.upload.buffer</name>
+  <value>array</value>
+</property>
+
+```
+#### <a name="s3a_fast_upload_thread_tuning"></a>S3A Fast Upload Thread Tuning
+
```
thodemoor: As a (probably better) alternative to my other comments, we could explain the bound on the memory consumption here once and link to it.

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
```diff
@@ -881,40 +881,362 @@ Seoul
...
+#### <a name="s3a_multipart_purge"></a>Cleaning up After Incremental Upload Failures: `fs.s3a.multipart.purge`
+
+
+If an incremental streaming operation is interrupted, there may be 
+intermediate partitions uploaded to S3 —data which will be billed for.
+
+These charges can be reduced by enabling `fs.s3a.multipart.purge`, 
+and setting a purge time in seconds, such as 86400 seconds —24 hours, after
+which the S3 service automatically deletes outstanding multipart
```
thodemoor: To me, the wording here gives the impression this is a server-side operation but the purging happens on the client by listing all uploads and then sending a delete call with the ones to be purged. Consequently, this can cause a (slight) delay when instantiating an s3a FS instance and there are lots of active uploads (to purge).

hadoop-common-project/hadoop-common/src/main/resources/core-default.xml
```diff
@@ -1095,10 +1102,50 @@
...
+    Both "array" and "bytebuffer" will consume memory in a single stream up to the number
+    of blocks set by:
+
+        fs.s3a.multipart.size * fs.s3a.fast.upload.active.blocks.
+
+    If using either of these mechanisms, keep this value low
+
+    The total number of threads performing work across all threads is set by
+    fs.s3a.threads.max, with fs.s3a.max.total.tasks values setting the number of queued
+    work items.
```
steveloughran: you know, now that you can have a queue per stream, it could be set to something
bigger. This is something we could look at in the docs, leaving out of the XML so as
to have a single topic. This phrase here describes the number of active threads, which
is different —and will be more so once there's other work (COPY, DELETE) going on there.

So: wont change here

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -118,21 +126,37 @@
...
+  private S3ADataBlocks.BlockFactory blockFactory;
+  private int blockOutputActiveBlocks;
+
+  /*
+   * Register Deprecated options.
+   */
+  static {
+    Configuration.addDeprecations(new Configuration.DeprecationDelta[]{
+        new Configuration.DeprecationDelta("fs.s3a.threads.core",
+            null,
```
pieterreuse: I'm not familiar with DeprecationDelta's, but this _null_ value gave rise to a nullpointerexception on **all** unit tests when fs.s3a.threads.core was in my config. Replacing this _null_ with _""_ (empty string) resolved my issue, but I'm not 100% sure that is the right thing to do here.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ABlockOutputStream.java
```diff
@@ -0,0 +1,699 @@
...
+    S3ADataBlocks.DataBlock block = createBlockIfNeeded();
+    int written = block.write(source, offset, len);
+    int remainingCapacity = block.remainingCapacity();
+    if (written < len) {
+      // not everything was written —the block has run out
+      // of capacity
+      // Trigger an upload then process the remainder.
+      LOG.debug("writing more data than block has capacity -triggering upload");
+      uploadCurrentBlock();
+      // tail recursion is mildly expensive, but given buffer sizes must be MB.
```
steveloughran: OK. I've set that limit in Constants and will log @ error if the #of blocks exceeds it. We'll see what happens.

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
steveloughran: ..again, not changing it in either place,  as once renames() parallelize, life gets more complex

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ABlockOutputStream.java
thodemoor: We can . With the min part size of 5MB you need a 50GB upload to test this. Will take a while vs. AWS. We can test this cheaply, but of course vs our S3-clone, but at least that will test the log @ error.
@pieterreuse please add this to our testplan

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
steveloughran: adding link to the s3a_fast_upload_thread_tuning section

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
steveloughran: yep.

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
steveloughran: Does it? Never knew that. I'd thought it was server side. Will change. Also, we could make that an async operation; it's not needed to bring up the FS.

thodemoor: The grunt work is done in `com.amazonaws.services.s3.transfer.TransferManager#abortMultipartUploads`

thodemoor: And yes making async is again a very good idea here.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
steveloughran: I've just cut that section entirely. That's  harsh, but, well, it the fast output stream was always marked as experimental ... we've learned from the experiment and are now changing behaviour here, which is something we can look at covering in the release notes. I'll add that to the JIRA.

pieterreuse: That indeed fixes the problems I had, thx for looking into this.

mojodna:
@steveloughran I'm trying this as part of 3.0.0-alpha2 (it's exactly what I was looking for after running into the same OOM problems) and wondering when it cleans up the disk-cached blocks.

I'm generating a ~50GB file on an instance with ~6GB free when the process starts. My expectation is that local copies of the blocks would be deleted after those parts finish uploading, but I'm seeing more than 15 blocks in `/tmp` (and none of them have been deleted thus far).

I can't confirm that any parts have finished uploading, though I suspect they have.

I see that `DiskBlock` deletes temporary files when closed, but is it closed after the block has finished uploading or when the entire file has been fully written to the FS?

steveloughran:
Can you comment on that in a JIRA, not a PR? Thanks

steveloughran:
That's https://issues.apache.org/jira/secure/Dashboard.jspa ; project HADOOP, component fs/s3 /  

They should be deleted as soon as the upload completes; the `close()` call that the AWS httpclient makes on the input stream triggers the deletion. Though there aren't tests for it, as I recall.

mojodna:
Done: https://issues.apache.org/jira/browse/HADOOP-14028

PR #131, opened 2016-09-26
HADOOP-13655 document object store use with fs shell and distcp	

steveloughran:
Patch of filesystem shell & distcp docs to cover object stores. Also updated some references in filesystem/index.md which were out of date

### DIFF
--- a/hadoop-common-project/hadoop-common/src/site/markdown/FileSystemShell.md
+++ b/hadoop-common-project/hadoop-common/src/site/markdown/FileSystemShell.md
@@ -53,9 +53,13 @@ Returns 0 on success and 1 on error.
 cat
 ---
 
-Usage: `hadoop fs -cat URI [URI ...]`
+Usage: `hadoop fs -cat [-ignoreCrc] URI [URI ...]`
 
 Copies source paths to stdout.
+
+Options
+
+* The `-ignoreCrc` option disables checkshum verification.
 
 Example:
 
@@ -116,11 +120,16 @@ copyFromLocal
 
 Usage: `hadoop fs -copyFromLocal <localsrc> URI`
 
-Similar to put command, except that the source is restricted to a local file reference.
+Similar to the `fs -put` command, except that the source is restricted to a local file reference.
 
 Options:
 
-* The -f option will overwrite the destination if it already exists.
+* `-p` : Preserves access and modification times, ownership and the permissions.
+(assuming the permissions can be propagated across filesystems)
+* `-f` : Overwrites the destination if it already exists.
+* `-l` : Allow DataNode to lazily persist the file to disk, Forces a replication
+ factor of 1. This flag will result in reduced durability. Use with care.
+* `-d` : Skip creation of temporary file with the suffix `._COPYING_`.
 
 copyToLocal
 -----------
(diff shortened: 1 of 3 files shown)


hadoop-common-project/hadoop-common/src/site/markdown/FileSystemShell.md
```diff
@@ -315,7 +324,11 @@ Returns 0 on success and -1 on error.
 
 Options:
 
-The -f option will overwrite the destination if it already exists.
+* `-p` : Preserves access and modification times, ownership and the permissions.
+(assuming the permissions can be propagated across filesystems)
+* `-f` : Overwrites the destination if it already exists.
+* `-ignorecrc` : Skip CRC checks on the file(s) downloaded.
+* `crc`: write CRC checksums for the files downloaded.
```
steveloughran: fixed

hadoop-common-project/hadoop-common/src/site/markdown/FileSystemShell.md
```diff
@@ -729,3 +757,278 @@ usage
...
+
+```bash
+# Upload a file from the cluster filesystem
+hadoop fs -put -d /datasets/example.orc s3a://bucket/datasets/
+
+# Upload a file from the local filesystem
+hadoop fs -copyFromLocal -d -f ~/datasets/devices.orc s3a://bucket/datasets/
+
+# create a file from stdin
+echo "hello" | hadoop fs -put -d -f - wasb://yourcontainer@youraccount.blob.core.windows.net/hello.txt
```
steveloughran: actually no! I couldn't get the put command to read off stdin according to the (Existing) docs, so I looked in the source code and saw that you needed a special "-" as the source, otherwise the s3a url :// was being treated as a source and the command failing as the destination was missing.  It does make sense —and is consistent with other tools, just not documented completely.

I'd mentioned this in the "put" section, but I've expanded the comment here and added the - to the DNF form of the put command.

hadoop-tools/hadoop-distcp/src/site/markdown/DistCp.md.vm
```diff
@@ -470,6 +470,105 @@ $H3 SSL Configurations for HSFTP sources
...
+**Notes**
+
+* The `-atomic` option causes a rename of the temporary data, so significantly
+increases the time to commit work at the end of the operation. Furthermore,
+as Object Stores other than (optionally) `wasb://` do not offer atomic renames of directories
+the `-atomic` operation doesn't actually deliver what is promised. *Avoid*.
+
+* The `-append` option is not supported.
+
+* The `-diff` option is not supported
```
liuml07: The `-diff/-rdiff` option is not supported

Yes there is an `rdiff` options that is just added.

hadoop-tools/hadoop-distcp/src/site/markdown/DistCp.md.vm
```diff
@@ -470,6 +470,105 @@ $H3 SSL Configurations for HSFTP sources
...
+be it a heap overflow or a YARN container termination.
+This is particularly common if the network bandwidth
+between the cluster and the object store is limited (such as when working
+with remote object stores). It is best to disable/avoid such options and
+rely on disk buffering.
+
+* Copy operations within a single object store still take place in the Hadoop cluster
+—even when the object store implements a more efficient COPY operation internally
+
+    That is, an operation such as
```
liuml07: The indention is unnecessary?

hadoop-tools/hadoop-distcp/src/site/markdown/DistCp.md.vm
steveloughran: ok

hadoop-tools/hadoop-distcp/src/site/markdown/DistCp.md.vm
steveloughran: it is needed to indent the paragraph and code sample under the bullet point.

PR #134, opened 2016-10-06
YARN-5025. Container move (relocation) between nodes

szape:
Support for relocating containers has become a must-have requirement for most multi-service applications, since the inevitable concept-drifts make SLAs hard to be satisfied. The relocation and co-location of services (long running containers) can help to reduce bottlenecks in a multi-service cluster, especially where data-intensive, streaming applications interfere.


elek:

PR #136, opened 2016-10-07
HADOOP-12774 use UGI.currentUser for user and group of s3a objects

steveloughran:
This patch grabs the UGI current user shortname in the FS initialize call, then uses that as the user and group for all filestatus instances generated.

```
$ ./hadoop fs -ls s3a://hwdev-steve-ireland/
Found 1 items
drwxrwxrwx   - stevel stevel          0 2016-10-07 17:29 s3a://hwdev-steve-ireland/tests
```

### DIFF
--- a/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileStatus.java
+++ b/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileStatus.java
@@ -32,18 +32,24 @@ import org.apache.hadoop.fs.Path;
 @InterfaceStability.Evolving
 public class S3AFileStatus extends FileStatus {
   private boolean isEmptyDirectory;
+  private final String owner;
 
   // Directories
-  public S3AFileStatus(boolean isdir, boolean isemptydir, Path path) {
+  public S3AFileStatus(boolean isdir,
+      boolean isemptydir,
+      Path path,
+      String owner) {
     super(0, isdir, 1, 0, 0, path);
     isEmptyDirectory = isemptydir;
+    this.owner = owner;
   }
 
   // Files
   public S3AFileStatus(long length, long modification_time, Path path,
-      long blockSize) {
+      long blockSize, String owner) {
     super(length, false, 1, blockSize, modification_time, path);
     isEmptyDirectory = false;
+    this.owner = owner;
   }
 
   public boolean isEmptyDirectory() {
@@ -52,7 +58,16 @@ public class S3AFileStatus extends FileStatus {
 
   @Override
   public String getOwner() {
-    return System.getProperty("user.name");
+    return owner;
+  }
+
+  /**
+   * The group of an S3A entry is the same as the owner
+   * @return the owner.
+   */
+  @Override
+  public String getGroup() {
+    return owner;
   }
 
   /** Compare if this object is equal to another object.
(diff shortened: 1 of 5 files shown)


steveloughran:
Patch 002: create a fake user, create an FS and verify that the user and owner on the root path is that username

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileStatus.java
```diff
@@ -32,18 +32,24 @@
 @InterfaceStability.Evolving
 public class S3AFileStatus extends FileStatus {
   private boolean isEmptyDirectory;
+  private final String owner;
```
cnauroth: I think we can achieve this change without adding a member variable in the subclass.  (See more specific notes to follow.)  Removing this member variable would reduce memory footprint.  Admittedly, the memory cost is probably not significant, but as we start thinking about the possibility of caching `FileStatus` instances client-side for things like S3Guard, then the per-instance memory cost of each `FileStatus` could become more significant.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileStatus.java
```diff
@@ -32,18 +32,24 @@
...
 
   // Directories
-  public S3AFileStatus(boolean isdir, boolean isemptydir, Path path) {
+  public S3AFileStatus(boolean isdir,
+      boolean isemptydir,
+      Path path,
+      String owner) {
     super(0, isdir, 1, 0, 0, path);
     isEmptyDirectory = isemptydir;
+    this.owner = owner;
```
cnauroth: Consider removing the member variable and changing this line of code to:

    this.setOwner(owner);
    this.setGroup(owner);

(These are `protected` methods in the base class.)

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileStatus.java
```diff
@@ -52,7 +58,16 @@ public boolean isEmptyDirectory() {
 
   @Override
   public String getOwner() {
```
cnauroth: If we implement the above comments, then we can completely remove the override of `getOwner` here.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileStatus.java
```diff
@@ -52,7 +58,16 @@ public boolean isEmptyDirectory() {
...
-    return System.getProperty("user.name");
+    return owner;
+  }
+
+  /**
+   * The group of an S3A entry is the same as the owner
+   * @return the owner.
+   */
+  @Override
+  public String getGroup() {
```
cnauroth: The override of `getGroup` could be removed too.

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/ITestS3AConfiguration.java
```diff
@@ -42,8 +42,11 @@
 
 import java.io.File;
 import java.net.URI;
+import java.security.PrivilegedAction;
```
cnauroth: Unused import?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileStatus.java
```diff
@@ -32,18 +32,24 @@
...
+    this.owner = owner;
   }
 
   // Files
   public S3AFileStatus(long length, long modification_time, Path path,
-      long blockSize) {
+      long blockSize, String owner) {
     super(length, false, 1, blockSize, modification_time, path);
     isEmptyDirectory = false;
+    this.owner = owner;
```
cnauroth: Similar to the earlier comment:

    this.setOwner(owner);
    this.setGroup(owner);

PR #143, opened 2016-10-18
YARN-4597: initial commit

xslogic:


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+    }
+  }
+
+  private void startAllocatedContainer(Container container) {
+    LOG.info("Starting container [" + container.getContainerId()+ "]");
+    scheduledToRunContainers.put(container.getContainerId(), container);
+    this.utilizationManager.addContainerResources(container);
+    if (container.getContainerTokenIdentifier().getExecutionType() ==
+        ExecutionType.OPPORTUNISTIC) {
+      this.opportunisticContainersStatus.setOpportMemoryUsed(
```
kambatla: Method name has "used", while we actually refer to allocated. Should we name this differently? 

Or, would it make more sense to add another boolean to OpportunisticContainersStatus to capture whether used refers to allocated/utilization?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+
+  // Used to keep track of containers that have been marked to be killed
+  // to make room for a guaranteed container.
+  private final Map<ContainerId, Container> oppContainersMarkedForKill =
+      new HashMap<>();
+
+  // Containers launched by the Scheduler will take a while to actually
+  // move to the RUNNING state, but should still be fair game for killing
+  // by the scheduler to make room for guaranteed containers.
+  private final LinkedHashMap<ContainerId, Container> scheduledToRunContainers =
```
kambatla: Also, should this be a stack? So, containers can be killed in a LIFO manner?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+      this.opportunisticContainersStatus.setRunningOpportContainers(
+          this.opportunisticContainersStatus.getRunningOpportContainers()
+              + 1);
+    }
+    container.sendLaunchEvent();
+  }
+
+  private List<Container> pickOpportunisticContainersToKill(
+      ContainerId containerToStartId) {
+    // The additional opportunistic containers that need to be killed for the
```
kambatla: Again, why are these additional?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+    // given container to start.
+    List<Container> extraOpportContainersToKill = new ArrayList<>();
```
kambatla: Same comment about extra...

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+    EventHandler<ContainerSchedulerEvent> {
+
+  private static final Logger LOG =
+      LoggerFactory.getLogger(ContainerScheduler.class);
+
+  private final Context context;
+  private final int maxOppQueueLength;
+
+  // Queue of Guaranteed Containers waiting for resources to run
+  private final LinkedHashMap<ContainerId, Container>
```
kambatla: Why do these need to be a map, why not a queue? Is it because there could be multiple Container objects corresponding to one ContainerId? 

If yes, would it make sense to implement equals and hashCode on Container to ensure we check only ContainerId?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+      queuedGuaranteedContainers = new LinkedHashMap<>();
+  // Queue of Opportunistic Containers waiting for resources to run
+  private final LinkedHashMap<ContainerId, Container>
```
kambatla: Same comment about it being a queue here..

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+    // Use a descending iterator to kill more recently started containers.
+    Iterator<Container> reverseContainerIterator =
+        new LinkedList<>(scheduledToRunContainers.values()).descendingIterator();
+    while(reverseContainerIterator.hasNext() &&
+        !hasSufficientResources(resourcesToFreeUp)) {
+      Container runningCont = reverseContainerIterator.next();
+      if (runningCont.getContainerTokenIdentifier().getExecutionType() ==
+          ExecutionType.OPPORTUNISTIC) {
+
+        if (oppContainersMarkedForKill.containsKey(
```
kambatla: How about removing all containers that are already marked before we get to this..

If runningOppContainers were a stack and oppContainersToKill were a queue, you could just do 
`
for (Container container : runningOppContainers.removeAll(oppContainersToKill)) {
   killContainer(container);
   if (hasSufficientResources()) {
        break;
   }
}`

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+          continue;
+        }
+        extraOpportContainersToKill.add(runningCont);
+        ResourceUtilizationManager.decreaseResourceUtilization(
+            getContainersMonitor(), resourcesToFreeUp,
+            runningCont.getResource());
+      }
+    }
+    if (!hasSufficientResources(resourcesToFreeUp)) {
+      LOG.warn("There are no sufficient resources to start guaranteed [{}]" +
```
kambatla: Shouldn't be this higher than WARN? 

This seems like a pretty bad case that we should never run into. Should we take more drastic measures? Crash the NM especially if failfast is true?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+    for (Container container : queuedGuaranteedContainers.values()) {
+      ResourceUtilizationManager.increaseResourceUtilization(
+          getContainersMonitor(), resourceAllocationToFreeUp,
+          container.getResource());
+      if (container.getContainerId().equals(containerToStartId)) {
+        break;
+      }
+    }
+
+    // These Resources have already been freed, due to demand from an
```
kambatla: Rephrase? "These resources are being freed, likely at the behest of another guaranteed container."?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+
+    // Subtract the overall node resources.
+    getContainersMonitor().subtractNodeResourcesFromResourceUtilization(
+        resourceAllocationToFreeUp);
+    return resourceAllocationToFreeUp;
+  }
+
+  public void updateQueuingLimit(ContainerQueuingLimit limit) {
+    this.queuingLimit.setMaxQueueLength(limit.getMaxQueueLength());
+    // TODO: Include wait time as well once it is implemented
```
kambatla: Annotate the TODO with a JIRA?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+  private void shedQueuedOpportunisticContainers() {
+    int numAllowed = this.queuingLimit.getMaxQueueLength();
+    Iterator<Container> containerIter =
+        queuedOpportunisticContainers.values().iterator();
+    while (containerIter.hasNext()) {
+      Container container = containerIter.next();
+      if (numAllowed <= 0) {
+        container.sendKillEvent(
+            ContainerExitStatus.KILLED_BY_CONTAINER_SCHEDULER,
+            "Container Killed to make room for Guaranteed Container.");
```
kambatla: Shouldn't this diagnostic message be similar to the log below?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerSchedulerEvent.java
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.yarn.server.nodemanager.containermanager.scheduler;
+
+import org.apache.hadoop.yarn.event.AbstractEvent;
+import org.apache.hadoop.yarn.server.nodemanager.containermanager.container
+    .Container;
```
kambatla: One line for imports?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-common/src/main/resources/yarn-default.xml
```diff
@@ -1000,10 +1000,10 @@
   </property>
 
   <property>
-    <description>Enable Queuing of OPPORTUNISTIC containers on the
+    <description>Max numbed of OPPORTUNISTIC containers to queue at the
```
kambatla: s/numbed/number

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/ContainerManager.java
```diff
@@ -26,6 +26,8 @@
 import org.apache.hadoop.yarn.server.nodemanager.ContainerManagerEvent;
 import org.apache.hadoop.yarn.server.nodemanager.containermanager.monitor
     .ContainersMonitor;
+import org.apache.hadoop.yarn.server.nodemanager.containermanager.scheduler
+    .ContainerScheduler;
```
kambatla: imports should be on a single line?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/ContainerManagerImpl.java
```diff
@@ -311,6 +318,10 @@ protected void createAMRMProxyService(Configuration conf) {
     }
   }
 
+  protected ContainerScheduler createContainerScheduler(Context cntxt) {
```
kambatla: Is this method exposed for testing? If yes, should it be marked @VisibleForTesting?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/monitor/ContainersMonitorImpl.java
```diff
@@ -743,6 +739,8 @@ private void changeContainerResource(
       LOG.warn("Container " + containerId.toString() + "does not exist");
       return;
     }
+    // TODO: Route this through the ContainerScheduler to
```
kambatla: Annotate TODO with JIRA?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ResourceUtilizationManager.java
```diff
@@ -0,0 +1,163 @@
...
+ * This class abstracts out how a container contributes to Resource Utilization.
+ * It is used by the {@link ContainerScheduler} to determine which
+ * OPPORTUNISTIC containers to be killed to make room for a GUARANTEED
+ * container.
+ * It currently equates resource utilization with the total resource allocated
+ * to the container. Another implementation might choose to use the actual
+ * resource utilization.
+ */
+
+public class ResourceUtilizationManager {
```
kambatla: Should this be called ResourceUtilizationTracker?

Also, would it make more sense for this to be an interface with an allocation-based implementation? Later, one could add a usage-based implementation and may be integrate tightly with ContainersMonitor?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ResourceUtilizationManager.java
```diff
@@ -0,0 +1,163 @@
...
+      return false;
+    }
+    return true;
+  }
+
+  public ContainersMonitor getContainersMonitor() {
+    return this.scheduler.getContainersMonitor();
+  }
+
+  public static void increaseResourceUtilization(
```
kambatla: Shouldn't this be a non-static method on ResourceUtilization instead?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ResourceUtilizationManager.java
```diff
@@ -0,0 +1,163 @@
...
+      ContainersMonitor containersMonitor, ResourceUtilization resourceAlloc,
+      Resource resource) {
+    float vCores = (float) resource.getVirtualCores() /
+        containersMonitor.getVCoresAllocatedForContainers();
+    int vmem = (int) (resource.getMemorySize()
+        * containersMonitor.getVmemRatio());
+    resourceAlloc.addTo((int)resource.getMemorySize(), vmem, vCores);
+  }
+
+  public static void decreaseResourceUtilization(
```
kambatla: Shouldn't this be a non-static method on ResourceUtilization instead?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ResourceUtilizationManager.java
```diff
@@ -0,0 +1,163 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.yarn.server.nodemanager.containermanager.scheduler;
+
+import org.apache.hadoop.yarn.api.records.ExecutionType;
```
kambatla: Some unused imports.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+              container.getContainerId());
+        } catch (IOException e) {
+          LOG.warn("Could not store container state into store..", e);
+        }
+      }
+    }
+  }
+
+  private void killOpportunisticContainers(Container container) {
+    List<Container> extraOpportContainersToKill =
```
kkaranasos: I had named that extra because some containers might already have been marked for killing, and here you will keep the additional ones.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
kkaranasos: I would suggest to not turn it into a stack, because eventually we might want to kill not based on arrival time, but on job progress, etc.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: I chose a LinkedHashMap, since it is essentially an indexed Queue. We might need to check if a container exists in the queue or remove a specific container from the queue (for eg. when AM asks to kill a queued container).
Also, like you noted, we would also need to implement hashCode/equals for Container...

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The ContainerScheduler manages a collection of runnable containers. It
+ * ensures that a container is launched only if all it launch criteria are
```
xslogic: Will do

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: Ditto :)

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: I agree with @kkaranasos, lets keep it as it is... Regarding the LIFO order of killing, I am using a descending iterator so candidates are examined in LIFO order.

It was earlier 'runningContainers'. But based of @jian-he's feedback, we felt this is better, since the containers need not be immediately running at the point it is added to this.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+    if (maxOppQueueLength <= 0) {
+      startAllocatedContainer(container);
+      return;
+    }
+    if (queuedGuaranteedContainers.isEmpty() &&
+        queuedOpportunisticContainers.isEmpty() &&
+        this.utilizationManager.hasResourcesAvailable(container)) {
+      startAllocatedContainer(container);
+    } else {
+      LOG.info("No available resources for container {} to start its execution "
```
xslogic: So, I felt this should be INFO, since it should happen less frequently, and based on experience from production logs, it is better know immediately why a container did not run (and not have to re-run in DEBUG mode)

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+      boolean isQueued = true;
+      if (container.getContainerTokenIdentifier().getExecutionType() ==
+          ExecutionType.GUARANTEED) {
+        queuedGuaranteedContainers.put(container.getContainerId(), container);
+        // Kill running opportunistic containers to make space for
+        // guaranteed container.
+        killOpportunisticContainers(container);
+      } else {
+        if (queuedOpportunisticContainers.size() <= maxOppQueueLength) {
+          LOG.info("Opportunistic container {} will be queued at the NM.",
```
xslogic: Same as above...

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+              container.getContainerId());
+          queuedOpportunisticContainers.put(
+              container.getContainerId(), container);
+        } else {
+          isQueued = false;
+          LOG.info("Opportunistic container [{}] will not be queued at the NM" +
```
xslogic: Ditto.. as above.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,393 @@
...
+      LOG.info(
+          "Opportunistic container {} will be killed in order to start the "
+              + "execution of guaranteed container {}.",
+          contToKill.getContainerId(), container.getContainerId());
+    }
+  }
+
+  private void startAllocatedContainer(Container container) {
+    LOG.info("Starting container [" + container.getContainerId()+ "]");
+    scheduledToRunContainers.put(container.getContainerId(), container);
```
xslogic: I agree it is not technically needed. But, being a scheduler, I felt it would come in handy to have a data-structure of running containers.
One other use-case I forsee is if the container is updated (if we implement container promotion, for instance), we would have to move it from one list to another, which is cumbersome.
I vote to keep it as it is for the time-being and move it if required.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: I agree totally, but for the current patch, I was just hoping to bring parity with YARN-2995 (which introduced the OpportunisticContainerStatus) and YARN-2883 (which introduced the QueuingContainerManager).
Can we maybe do the above in a separate patch? Also, given that in this patch, for all practical purposes, used == allocated from the perspective of the Scheduler.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: Will fix the comment..

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: In this case, the 'extra' is needed. It signifies all the container above the already marked 'oppContainersToKill' that need to be killed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: Actually that message might be a bit misleading. It should be re-worded to 'There are no sufficient resources at the moment to start guaranteed...'. You can only reach here if kill signals have been sent to Opp. containers, but the completed container event hasn't been received by the Scheduler yet. At which point the guaranteed container will start.
It is never possible to allocate a guaranteed container on node that cannot start it.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: Done..

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: The Log.info line is printed immediately, which is why it has 'will be killed'.
The diagnostic line is part of the Container status message and is in past-tense and does not need the containerid.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerSchedulerEvent.java
xslogic: My IDE is very obstinate about it :) Can we leave it as it is.. especially since the checkstyle anyway ignores it ?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-common/src/main/resources/yarn-default.xml
xslogic: done..

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/ContainerManager.java
xslogic: as mentioned earlier.. can we leave it as it is, since it is ignored by checkstyle ?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/ContainerManagerImpl.java
xslogic: will do.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/monitor/ContainersMonitorImpl.java
xslogic: will do: YARN-5860

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ResourceUtilizationManager.java
xslogic: will do..

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ResourceUtilizationManager.java
xslogic: I agree about the rename... will do..
But lets make it an interface when we add the alternate implementation ?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ResourceUtilizationManager.java
xslogic: I sort of agree, but since it depends certain information from the ContainersMonitor (which we have to pass in to the method), I thought we should maybe keep it as static.
If you don't mind, lets keep it a it is for this patch.. and refactor when we get a more concrete argument.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -295,29 +306,29 @@ private void startAllocatedContainer(Container container) {
     // Go over the running opportunistic containers.
     // Use a descending iterator to kill more recently started containers.
     Iterator<Container> reverseContainerIterator =
```
kambatla: Nit: Should we call this lifoIterator?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ResourceUtilizationTracker.java
```diff
@@ -0,0 +1,161 @@
...
+ * This class abstracts out how a container contributes to Resource Utilization.
+ * It is used by the {@link ContainerScheduler} to determine which
+ * OPPORTUNISTIC containers to be killed to make room for a GUARANTEED
+ * container.
+ * It currently equates resource utilization with the total resource allocated
+ * to the container. Another implementation might choose to use the actual
+ * resource utilization.
+ */
+
+public class ResourceUtilizationTracker {
```
kambatla: Didn't we think this should be an interface with the default implementation being the one that considers allocation - a.k.a AllocationBased{ResourceUtilization}Tracker

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
```diff
@@ -0,0 +1,411 @@
...
+    // The opportunistic containers that need to be killed for the
+    // given container to start.
+    List<Container> extraOpportContainersToKill = new ArrayList<>();
+    // Track resources that need to be freed.
+    ResourceUtilization resourcesToFreeUp = resourcesToFreeUp(
+        containerToStartId);
+
+    // Go over the running opportunistic containers.
+    // Use a descending iterator to kill more recently started containers.
+    Iterator<Container> reverseContainerIterator =
```
kambatla: Nit: should we call this lifoIterator so it is easier to understand?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ResourceUtilizationTracker.java
xslogic: Done.... fixed in latest commit

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: Done.. fixed in latest commit

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/java/org/apache/hadoop/yarn/server/nodemanager/containermanager/scheduler/ContainerScheduler.java
xslogic: Done... fixed in latest commit

elek:

xslogic:
Hmm.. anything specific to verify ? It was committed about an year and half
ago.
And has been part of 2.9.x and 3.x.x releases.

Cheers
-Arun

elek:
Sorry for the noise, It was just accidentally picked by a jenkins job which didn't work as expected.
But as it's already committed: I will close this PR if no objections....

aajisaka:
YARN-4597 has been resolved. Closing.

PR #146, opened 2016-10-26
HADOOP-6801. io.sort.mb and io.sort.factor were renamed and moved to …

QwertyManiac:
…mapreduce but are still in CommonConfigurationKeysPublic.java and used in SequenceFile.java


QwertyManiac:
Rebased the patch.

PR #147, opened 2016-10-26
HADOOP-1381. The distance between sync blocks in SequenceFiles should be configurable

QwertyManiac:
- Allowed the sync interval size to be configurable via a new parameter in the Writer argument API
- New default sync interval size is 200 KiB
- Added tests with varied sync interval sizes, and kept a test for the older default as-is for compatibility checks
- Moved some older Reader API usage in modified tests to support newer constructions


QwertyManiac:
Fixed via 07825f2b49384dbec92bfae87ea661cef9ffab49

PR #149, opened 2016-10-30
YARN-5800: Deleted LinuxContainerExecutor comment from yarn-default.xml

HorizonNet:
https://issues.apache.org/jira/browse/YARN-5800
- Removed the comment about LinuxContainerExecutor from the yarn-default.xml file


HorizonNet:
This one was already applied to trunk via [this commit](https://github.com/apache/hadoop/commit/773c60bd7bd00651dc3016799b424b9bd2233eb3).

PR #152, opened 2016-11-07
HDFS-11115 Remove bytes2String and string2Bytes

SahilKang:
Since StandardCharsets makes converting between (utf-8) bytes and strings
trivial, let's remove the methods:
    - org.apache.hadoop.hdfs.DFSUtilClient.bytes2String
    - org.apache.hadoop.hdfs.DFSUtilClient.string2Bytes


elek:

PR #153, opened 2016-11-07
HDFS-11114. Support for running async disk checks in DataNode.

arp7:
Interface for running async checks on a resource.

The implementation ThrottledAsyncChecker supports throttling and result-caching. DataNode changes to use it will be done in another Jira.

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/checker/AsyncChecker.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/checker/AsyncChecker.java
@@ -58,5 +58,6 @@ public interface AsyncChecker<K, V> {
    *
    * @throws InterruptedException
    */
-  void join(long timeout, TimeUnit timeUnit) throws InterruptedException;
+  void shutdownAndWait(long timeout, TimeUnit timeUnit)
+      throws InterruptedException;
 }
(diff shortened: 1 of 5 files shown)


hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/checker/AsyncChecker.java
```diff
@@ -0,0 +1,62 @@
...
+   * Cancel all executing checks and wait for them to complete.
+   * First attempts a graceful cancellation, then cancels forcefully.
+   * Waits for the supplied timeout after both attempts.
+   *
+   * See {@link ExecutorService#awaitTermination} for a description of
+   * the parameters.
+   *
+   * @throws InterruptedException
+   */
+  void join(long timeout, TimeUnit timeUnit) throws InterruptedException;
```
anuengineer: Just trying to understand this a little better, From the signature and implementation this function looks more a awaitTermination in executorService. That is this function will wait for a while and cancel and task if the timeout occurs, in that case would you consider calling this await or awaitTermination. Java "join" seems to imply a wait without timeouts. Just making sure that the intended was indeed a shutdown/await pattern.

arp7: Thanks for taking a look @anuengineer. The method covers both shutdown and awaitTermination semantics. I could call it shutdownAndAwaitTermination() to make it clearer.

arp7:
Renamed it to shutdownAndWait.

PR #155, opened 2016-11-09
HDFS-11119. Support for parallel checking of StorageLocations on DataNode startup

arp7:
Introduce a StorageLocationChecker class that can parallelize checking StorageLocations. It also detects stalled checks and flags such volumes as failed. The DataNode will use this class in the next Jira.


anuengineer:
+1, LGTM

PR #157, opened 2016-11-10
HADOOP-13600. S3a rename() to copy files in a directory in parallel

sahilTakiar:
* Creates two separate `TransferManager`s - one for file uploads and one for file copies; this way they can be both configured separately; users may want to set the # of parallel uploads to a lower value than the # of parallel copies because uploads require actual I/O while copies do not
* Main modifications are to the `innerRename` method
* Instead of renaming a directory file by file, use the copy `TranferManager` to upload them in parallel
* Copies are submitted to the `TransferManager` and then tracked using the returned `Copy` object
* Deletes are handled via a `BlockingQueue` - once each copy is complete it adds a key to the queue, once `MAX_ENTRIES_TO_DELETE` keys need to be deleted, then the `removeKeys` method is invoked
* A separate thread is spawned to read from the delete queue and issue the delete requests
* A `ProgressListener` is used to track when a copy has been completed, once a copy finishes, the key to delete is added to the delete queue
* Some other re-factoring to make the above possible

### DIFF
--- a/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Constants.java
+++ b/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Constants.java
@@ -107,11 +107,15 @@ public final class Constants {
   public static final String MAX_PAGING_KEYS = "fs.s3a.paging.maximum";
   public static final int DEFAULT_MAX_PAGING_KEYS = 5000;
 
-  // the maximum number of threads to allow in the pool used by TransferManager
+  // the maximum number of threads to allow in the pool used by TransferManager for uploads
   public static final String MAX_THREADS = "fs.s3a.threads.max";
   public static final int DEFAULT_MAX_THREADS = 10;
 
-  // the time an idle thread waits before terminating
+  // the maximum number of threads to allow in the pool used by TransferManager for copies
+  public static final String COPY_MAX_THREADS = "fs.s3a.copy.threads.max";
+  public static final int DEFAULT_COPY_MAX_THREADS = 25;
+
+  // the time an idle thread waits before terminating for uploads and copies
   public static final String KEEPALIVE_TIME = "fs.s3a.threads.keepalivetime";
   public static final int DEFAULT_KEEPALIVE_TIME = 60;
 
--- a/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/CopyContext.java
+++ b/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/CopyContext.java
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.fs.s3a;
+
+import com.amazonaws.event.ProgressListener;
+import com.amazonaws.services.s3.transfer.Copy;
+
+
+/**
+ * A wrapper around a {@link Copy} object, the source key for the copy, the destination for the copy, the length of the
+ * key copied, and the {@link ProgressListener.ExceptionReporter} for the copy.
+ */
+class CopyContext {
+
+  private final Copy copy;
+  private final String srcKey;
+  private final String destKey;
+  private final long length;
+  private final ProgressListener.ExceptionReporter progressListener;
+
+  CopyContext(Copy copy, String srcKey, String destKey, long length,
+              ProgressListener.ExceptionReporter progressListener) {
+    this.copy = copy;
+    this.srcKey = srcKey;
+    this.destKey = destKey;
+    this.length = length;
+    this.progressListener = progressListener;
+  }
+
+  Copy getCopy() {
+    return copy;
+  }
+
+  String getSrcKey() {
+    return srcKey;
+  }
+
+  String getDestKey() {
+    return destKey;
+  }
+
+  long getLength() {
+    return length;
+  }
+
+  ProgressListener.ExceptionReporter getProgressListener() {
+    return progressListener;
+  }
+}
--- a/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/LazyTransferManager.java
+++ b/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/LazyTransferManager.java
@@ -1,0 +1,117 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.fs.s3a;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.transfer.TransferManager;
+import com.amazonaws.services.s3.transfer.TransferManagerConfiguration;
+
+import org.apache.hadoop.conf.Configuration;
+
+import static org.apache.hadoop.fs.s3a.Constants.COPY_MAX_THREADS;
+import static org.apache.hadoop.fs.s3a.Constants.DEFAULT_COPY_MAX_THREADS;
+import static org.apache.hadoop.fs.s3a.Constants.DEFAULT_KEEPALIVE_TIME;
+import static org.apache.hadoop.fs.s3a.Constants.DEFAULT_MAX_THREADS;
+import static org.apache.hadoop.fs.s3a.Constants.KEEPALIVE_TIME;
+import static org.apache.hadoop.fs.s3a.Constants.MAX_THREADS;
+import static org.apache.hadoop.fs.s3a.S3AUtils.getMaxThreads;
+import static org.apache.hadoop.fs.s3a.S3AUtils.longOption;
+
+
+/**
+ * A wrapper around a {@link TransferManager} that lazily initializes the {@link TransferManager}.
+ */
+class LazyTransferManager {
+
+  private final AmazonS3 s3;
+  private final TransferManagerConfiguration transferConfiguration;
+  private final int maxThreads;
+  private final long keepAliveTime;
+  private final String threadPoolName;
+
+  private TransferManager transferManager;
+  private ExecutorService executorService;
+
+  private LazyTransferManager(AmazonS3 s3, TransferManagerConfiguration transferConfiguration, int maxThreads,
+                      long keepAliveTime, String threadPoolName) {
+    this.s3 = s3;
+    this.transferConfiguration = transferConfiguration;
+    this.maxThreads = maxThreads;
+    this.keepAliveTime = keepAliveTime;
+    this.threadPoolName = threadPoolName;
+  }
+
+  synchronized TransferManager get() {
+    if (transferManager == null) {
+      transferManager = createTransferManager();
+      return transferManager;
+    }
+    return transferManager;
+  }
+
+  boolean isInitialized() {
+    return transferManager == null;
+  }
+
+  ExecutorService getExecutorService() {
+    return executorService;
+  }
+
+  private TransferManager createTransferManager() {
+    executorService = new ThreadPoolExecutor(
+            maxThreads, Integer.MAX_VALUE,
+            keepAliveTime, TimeUnit.SECONDS,
+            new LinkedBlockingQueue<>(),
+            BlockingThreadPoolExecutorService.newDaemonThreadFactory(
+                    threadPoolName));
+
+    TransferManager transferManager = new TransferManager(s3, executorService);
+    transferManager.setConfiguration(transferConfiguration);
+    return transferManager;
+  }
+
+  static LazyTransferManager createLazyUploadTransferManager(AmazonS3 s3, Configuration conf, long partSize,
+                                                                     long multiPartThreshold) {
+    int maxThreads = getMaxThreads(conf, MAX_THREADS, DEFAULT_MAX_THREADS);
+    long keepAliveTime = longOption(conf, KEEPALIVE_TIME, DEFAULT_KEEPALIVE_TIME, 0);
+
+    TransferManagerConfiguration transferConfiguration = new TransferManagerConfiguration();
+    transferConfiguration.setMinimumUploadPartSize(partSize);
+    transferConfiguration.setMultipartUploadThreshold(multiPartThreshold);
+
+    return new LazyTransferManager(s3, transferConfiguration, maxThreads, keepAliveTime, "s3a-upload-unbounded");
+  }
+
+  static LazyTransferManager createLazyCopyTransferManager(AmazonS3 s3, Configuration conf, long partSize,
+                                                                   long multiPartThreshold) {
+    int maxThreads = getMaxThreads(conf, COPY_MAX_THREADS, DEFAULT_COPY_MAX_THREADS);
+    long keepAliveTime = longOption(conf, KEEPALIVE_TIME, DEFAULT_KEEPALIVE_TIME, 0);
+
+    TransferManagerConfiguration transferConfiguration = new TransferManagerConfiguration();
+    transferConfiguration.setMultipartCopyPartSize(partSize);
+    transferConfiguration.setMultipartCopyThreshold(multiPartThreshold);
+
+    return new LazyTransferManager(s3, transferConfiguration, maxThreads, keepAliveTime, "s3a-copy-unbounded");
+  }
+}
--- a/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
+++ b/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
@@ -34,9 +34,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.Objects;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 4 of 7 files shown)


hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -207,11 +218,17 @@ public StorageStatistics provide() {
...
-      threadPoolExecutor = BlockingThreadPoolExecutorService.newInstance(
+      uploadThreadPoolExecutor = BlockingThreadPoolExecutorService.newInstance(
           maxThreads,
           maxThreads + totalTasks,
           keepAliveTime, TimeUnit.SECONDS,
-          "s3a-transfer-shared");
+          "s3a-upload-shared");
+
+      copyThreadPoolExecutor = BlockingThreadPoolExecutorService.newInstance(
+          maxThreads,
```
thodemoor: COPY is server-side (no data transfer) and is thus generally much less resource-intensive and much quicker than PUT (the smaller your bandwidth to S3, the bigger the difference becomes). So I think the `maxThreads` for the copyThreadpool could be (much) higher than for uploadThreadpool and should thus be configurable separately.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Constants.java
```diff
@@ -99,17 +99,25 @@ private Constants() {
...
 
-  // the maximum number of threads to allow in the pool used by TransferManager
-  public static final String MAX_THREADS = "fs.s3a.threads.max";
-  public static final int DEFAULT_MAX_THREADS = 10;
+  // the maximum number of threads to allow in the pool used by TransferManager for uploads
+  public static final String UPLOAD_MAX_THREADS = "fs.s3a.threads.max";
+  public static final int UPLOAD_DEFAULT_MAX_THREADS = 10;
+
+  // the maximum number of threads to allow in the pool used by TransferManager for copies
+  public static final String COPY_MAX_THREADS = "fs.s3a.threads.max";
```
thodemoor: fs.s3a.copy.threads.max?

sahilTakiar: Whoops, fixed. Leaving the parameters for the upload `TransferManager` the same since changing the name would be a backwards incompatible change.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Constants.java
```diff
@@ -107,11 +107,15 @@ private Constants() {
   public static final String MAX_PAGING_KEYS = "fs.s3a.paging.maximum";
   public static final int DEFAULT_MAX_PAGING_KEYS = 5000;
 
-  // the maximum number of threads to allow in the pool used by TransferManager
-  public static final String MAX_THREADS = "fs.s3a.threads.max";
-  public static final int DEFAULT_MAX_THREADS = 10;
+  // the maximum number of threads to allow in the pool used by TransferManager for uploads
+  public static final String UPLOAD_MAX_THREADS = "fs.s3a.threads.max";
+  public static final int DEFAULT_UPLOAD_MAX_THREADS = 10;
```
steveloughran: not going to rename constants because it will break anyone using them. Class is tagged {@InterfaceAudience.Public/evolving}

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -269,7 +273,8 @@ public StorageStatistics provide() {
       }
       useListV1 = (listVersion == 1);
 
-      initTransferManager();
+      this.uploads = createUploadTransferManager(conf);
+      this.copies = createCopyTransferManager(conf);
```
steveloughran: I'd actually like this to be on demand. That way, if not needed, no overhead of creating it. init times of s3a are already high

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -306,6 +311,58 @@ public StorageStatistics provide() {
...
+            new TransferManagerConfiguration();
+    transferConfiguration.setMinimumUploadPartSize(partSize);
+    transferConfiguration.setMultipartUploadThreshold(multiPartThreshold);
+
+    TransferManager uploads = new TransferManager(s3, uploadUnboundedThreadPool);
+    uploads.setConfiguration(transferConfiguration);
+    return uploads;
+  }
+
+  private TransferManager createCopyTransferManager(Configuration conf) {
```
steveloughran: this is almost the same as the previous one apart from the conf fields and names; should be a common method

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -896,45 +941,108 @@ private boolean innerRename(Path source, Path dest)
...
+      BlockingQueue<Optional<DeleteObjectsRequest.KeyVersion>> deleteQueue = new ArrayBlockingQueue<>(
+              MAX_ENTRIES_TO_DELETE * 1.5);
+
+      // Used to track if the delete thread was gracefully shutdown
+      boolean deleteFutureComplete = false;
+      FutureTask<Void> deleteFuture = null;
+
+      try {
+        // Launch a thread that will read from the deleteQueue and batch delete any files that have already been copied
+        deleteFuture = new FutureTask<>(new Callable<Void>() {
```
steveloughran: If we target java 8 only, this can be a proper lambda-expression :)

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -143,9 +150,11 @@
   private Listing listing;
   private long partSize;
   private boolean enableMultiObjectsDelete;
-  private TransferManager transfers;
+  private TransferManager uploads;
```
steveloughran: leave the name of this alone unless really, really needed because (a) we may do other things with it in future, (b) reduces the size of the diff hence compatibility with other patches.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -2167,7 +2280,34 @@ public String getCanonicalServiceName() {
    * @throws IOException Other IO problems
    */
   private void copyFile(String srcKey, String dstKey, long size)
-      throws IOException, InterruptedIOException, AmazonClientException {
+      throws IOException, AmazonClientException {
```
steveloughran: leave as before so as to make clear it may be interrupted, unless its mentioned in the javadocs

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Constants.java
sahilTakiar: Fixed

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: Added a separate class called `LazyTransferManager` that lazily initializes a `TransferManager`.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: Re-factored. Most of this was moved into `LazyTransferManager`. I abstracted the determination of # of threads into a common method.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: Changed to lambda expression.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: Fixed

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: Fixed

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/CopyContext.java
```diff
@@ -0,0 +1,34 @@
+package org.apache.hadoop.fs.s3a;
```
ajfabbri: Will need apache license text at top.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/LazyTransferManager.java
```diff
@@ -0,0 +1,63 @@
+package org.apache.hadoop.fs.s3a;
```
ajfabbri: Ditto here.. new files need the apache license text at top

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -891,50 +902,123 @@ private boolean innerRename(Path source, Path dest)
...
+
+      try {
+        // Launch a thread that will read from the deleteQueue and batch delete any files that have already been copied
+        deleteFuture = new FutureTask<>(() -> {
+          while (true) {
+            while (keysToDelete.size() < MAX_ENTRIES_TO_DELETE) {
+              Optional<DeleteObjectsRequest.KeyVersion> key = deleteQueue.take();
+
+              // The thread runs until is is given an EOF message (an Optional#empty())
+              if (key.isPresent()) {
```
ajfabbri: Use of Optional will make this harder to backport to Java 7 branches, no?   Looks like you are using it for an EOF marker only? You could also create a static final KeyVersion instance and use a reference equality comparison to detect that it is the special EOF version?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -891,50 +902,123 @@ private boolean innerRename(Path source, Path dest)
...
+          deleteFuture.get();
+        } catch (ExecutionException | InterruptedException e) {
+          throw new RenameFailedException(srcKey, dstKey, e);
+        }
+        deleteFutureComplete = true;
+      } finally {
+        if (!deleteFutureComplete) {
+          if (deleteFuture != null && !deleteFuture.isDone() && !deleteFuture.isCancelled()) {
+            deleteFuture.cancel(true);
+          }
```
ajfabbri: Wondering if we can move parallel rename logic to a new class.. S3AFileSystem is getting too big.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
steveloughran: I'm making the leap to Java 8 with the committer and retry logic. Doesn't mean we should jump to using Optional just because we can. Put differently: if you are going to just use it as a fancy null, it's overkill. Key benefit is that you and use map & foreach, but there we are crippled by java's checked exceptions stopping us throwing IOEs

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/CopyContext.java
```diff
@@ -0,0 +1,34 @@
...
+
+  Copy getCopy() {
+    return copy;
+  }
+
+  String getSrcKey() {
+    return srcKey;
+  }
+
+  String getDstKey() {
```
steveloughran: no, go on, use "destKey". We can afford the letter e's

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -241,26 +242,17 @@ public StorageStatistics provide() {
                     }
                   });
 
-      int maxThreads = conf.getInt(MAX_THREADS, DEFAULT_MAX_THREADS);
-      if (maxThreads < 2) {
-        LOG.warn(MAX_THREADS + " must be at least 2: forcing to 2.");
-        maxThreads = 2;
-      }
+      int maxThreads = getMaxThreads(conf, MAX_THREADS, DEFAULT_MAX_THREADS);
```
steveloughran: I'm assuming the checks are here for a reason. Unless lazy xfer manager does the uprate, it'll need reinstatement. Doing it in the manager would be best

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -241,26 +242,17 @@ public StorageStatistics provide() {
...
       int totalTasks = intOption(conf,
           MAX_TOTAL_TASKS, DEFAULT_MAX_TOTAL_TASKS, 1);
       long keepAliveTime = longOption(conf, KEEPALIVE_TIME,
           DEFAULT_KEEPALIVE_TIME, 0);
+
```
steveloughran: ideally, keep superfluous edits to a min. I know, we all abuse that...but its good to try

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -303,7 +296,37 @@ public StorageStatistics provide() {
     } catch (AmazonClientException e) {
       throw translateException("initializing ", new Path(name), e);
     }
+  }
+
+  private int getMaxThreads(Configuration conf, String maxThreadsKey, int defaultMaxThreads) {
```
steveloughran: move to S3AUtils

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -303,7 +296,37 @@ public StorageStatistics provide() {
...
+    int maxThreads = conf.getInt(maxThreadsKey, defaultMaxThreads);
+    if (maxThreads < 2) {
+      LOG.warn(maxThreadsKey + " must be at least 2: forcing to 2.");
+      maxThreads = 2;
+    }
+    return maxThreads;
+  }
+
+  private LazyTransferManager createLazyUploadTransferManager(Configuration conf) {
```
steveloughran: this logic could be moved into LazyUploadTransferManager itself, maybe just as a static method

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
steveloughran: mmm. I mostly concur, handing in either the S3aFS or the (expanded) WriteOperationsHelper

steveloughran:
In HADOOP-13786 I'm wrapping every single s3 client call with retry policy, then expanding the inconsistent client to generate more faults (initially throttle, later connection setup/response parsing). I'd really like this work to actually await that, as without it this code isn't going to be resilient to large copies where you are much more likely to hit parallel IO. And we need to make sure there's a good failure policy set up there

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/CopyContext.java
sahilTakiar: Whoops, always forget those. Fixed.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/LazyTransferManager.java
sahilTakiar: Fixed

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: I removed the usage of `Optional`. Using a `private static final DeleteObjectsRequest.KeyVersion END_OF_KEYS_TO_DELETE = new DeleteObjectsRequest.KeyVersion(null, null);` as the EOF instead.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: Moved all the parallel rename logic into a dedicated calls called `ParallelDirectoryRenamer`

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/CopyContext.java
sahilTakiar: Fixed

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: Fixed

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: Done

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
sahilTakiar: Done

sahilTakiar:
Updates:
* Moved the parallel rename logic into a dedicated class called `ParallelDirectoryRenamer`
* A few other bug fixes, the core logic remains the same

@steveloughran your last comment on HADOOP-13786 suggested you may move the retry logic out into a separate patch? Are you planning to do that? If not, do you think this patch requires waiting for all the work in HADOOP-13786 to be completed?

If there are concerns with retry behavior, we could also set the default value of the copy thread pool to be 1, that way this feature is essentially off by default.

Also what do you mean by "isn't going to be resilient to large copies where you are much more likely to hit parallel IO"? What parallel IO are you referring to?

steveloughran:
1. I don't see why this needs to be blocked waiting for all of the '13786 patch to go in
2. but the rename/throttling stuff: yes.

Parallel execution: we've got another transfer manager issuing COPY requests, so more HTTPS requests to an S3 bucket/shard. The more requests to a single shard, the likelier you are to hit failures. Now, I think the xfer manager does handle the 503 replies which come back, but the rest of the FS client doesn't

sahilTakiar:
@steveloughran ok that makes sense. Thanks for the explanation. Let me know if you need any help with pulling the retry logic out.

elek:

harshavardhana:
Is this being actively worked on?

steveloughran:
@harshavardhana I'm doing work on parallel renames in HADOOP-15183; the complication there is that I want to update the S3Guard metastore after each copy and then maybe do the delete immediately after too, or, as discussed on that JIRA, postpone all the deletes until every copy completes.

This is driven mostly by the need to have S3Guard fully resilient to failures during the operation; speedup with scale are also on that todo list though. 

I need to review @sahilTakiar's patch here to see what I can lift from it -it's clear that Sahil  was way ahead of the rest of us in using Java 8 idioms we're only now picking up. I think the ability to abort ongoing uploads is something we'd want to handle a failure -and which some of the progress callbacks here seems to set us up for. 

pr #654 is the work there.

aajisaka:
https://issues.apache.org/jira/browse/HADOOP-13600 is closed as duplicate. Closing this PR as well.

PR #160, opened 2016-11-12
HDFS-10930.07.patch

xiaoyuyao:
Create a PR for easy review.


xiaoyuyao:
Update PR to fix the findbugs and unit test failure.

PR #161, opened 2016-11-14
HADOOP-13817. Add a finite shell command timeout to ShellBasedUnixGroupsMapping.

QwertyManiac:
- Tests required log capture so the LOG variable was elevated in visibility, which required changes in a few test methods unrelated to just this change.
- Timeout is applied to both, the regular groups and the partial group listing check commands.
- Default is left to 0 to not break current behaviour (i.e. to wait forever until groups are resolvable).

### DIFF
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/CommonConfigurationKeysPublic.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/CommonConfigurationKeysPublic.java
@@ -517,6 +517,21 @@ public class CommonConfigurationKeysPublic {
    * <a href="{@docRoot}/../hadoop-project-dist/hadoop-common/core-default.xml">
    * core-default.xml</a>
    */
+  public static final String HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT =
+      "hadoop.security.groups.shell.groups.command.timeout";
+  /**
+   * @see
+   * <a href="{@docRoot}/../hadoop-project-dist/hadoop-common/core-default.xml">
+   * core-default.xml</a>
+   */
+  public static final long
+      HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT_DEFAULT =
+          0L;
+  /**
+   * @see
+   * <a href="{@docRoot}/../hadoop-project-dist/hadoop-common/core-default.xml">
+   * core-default.xml</a>
+   */
   public static final String  HADOOP_SECURITY_AUTHENTICATION =
     "hadoop.security.authentication";
   /**
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/CommonConfigurationKeysPublic.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/CommonConfigurationKeysPublic.java
@@ -518,7 +518,7 @@ public class CommonConfigurationKeysPublic {
    * core-default.xml</a>
    */
   public static final String HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT_SECS =
-      "hadoop.security.groups.shell.command.timeout.secs";
+      "hadoop.security.groups.shell.command.timeout";
   /**
    * @see
    * <a href="{@docRoot}/../hadoop-project-dist/hadoop-common/core-default.xml">
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/CommonConfigurationKeysPublic.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/CommonConfigurationKeysPublic.java
@@ -517,15 +517,15 @@ public class CommonConfigurationKeysPublic {
    * <a href="{@docRoot}/../hadoop-project-dist/hadoop-common/core-default.xml">
    * core-default.xml</a>
    */
-  public static final String HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT =
-      "hadoop.security.groups.shell.groups.command.timeout";
+  public static final String HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT_SECS =
+      "hadoop.security.groups.shell.command.timeout.secs";
   /**
    * @see
    * <a href="{@docRoot}/../hadoop-project-dist/hadoop-common/core-default.xml">
    * core-default.xml</a>
    */
   public static final long
-      HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT_DEFAULT =
+          HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT_SECS_DEFAULT =
           0L;
   /**
    * @see
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
@@ -52,16 +52,19 @@ public class ShellBasedUnixGroupsMapping extends Configured
       LoggerFactory.getLogger(ShellBasedUnixGroupsMapping.class);
 
   private long timeout = 0L;
-  private final List<String> emptyGroupsList = new LinkedList<>();
+  private static final List<String> EMPTY_GROUPS = new LinkedList<>();
 
   @Override
   public void setConf(Configuration conf) {
     super.setConf(conf);
-    timeout = conf.getTimeDuration(
-        CommonConfigurationKeys.HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT_SECS,
-        CommonConfigurationKeys.
-                HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT_SECS_DEFAULT,
-        TimeUnit.SECONDS);
+    if (conf != null) {
+      timeout = conf.getTimeDuration(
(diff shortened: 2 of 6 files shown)


hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
```diff
@@ -37,11 +43,23 @@
...
+  @VisibleForTesting
+  protected static final Logger LOG =
+      LoggerFactory.getLogger(ShellBasedUnixGroupsMapping.class);
+
+  private long timeout = 0L;
+
+  @Override
+  public void setConf(Configuration conf) {
+    super.setConf(conf);
+    timeout = conf.getLong(
```
xiaoyuyao: Can we use Configuration#getTimeDuration() instead of Configuration#getLong to make the newly added configuration key easier to use?

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
```diff
@@ -133,8 +171,26 @@ protected ShellCommandExecutor createGroupIDExecutor(String userName) {
         groups = resolvePartialGroupNames(user, e.getMessage(),
             executor.getOutput());
       } catch (PartialGroupNameException pge) {
-        LOG.warn("unable to return groups for user " + user, pge);
+        LOG.warn("unable to return groups for user {}", user, pge);
+        return new LinkedList<>();
```
xiaoyuyao: Can we use flyweight pattern to minimize memory usage for the empty LinkList?

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
```diff
@@ -133,8 +171,26 @@ protected ShellCommandExecutor createGroupIDExecutor(String userName) {
...
+      if (executor.isTimedOut()) {
+        LOG.warn(
+            "Unable to return groups for user '{}' as shell group lookup " +
+            "command '{}' ran longer than the configured timeout limit of " +
+            "{} seconds.",
+            user,
+            Arrays.asList(executor.getExecString()),
+            timeout
+        );
         return new LinkedList<>();
```
xiaoyuyao: flyweight pattern applies here as well.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
QwertyManiac: Done in added commit.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
QwertyManiac: Done in added commit.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/CommonConfigurationKeysPublic.java
```diff
@@ -517,15 +517,15 @@
    * <a href="{@docRoot}/../hadoop-project-dist/hadoop-common/core-default.xml">
    * core-default.xml</a>
    */
-  public static final String HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT =
-      "hadoop.security.groups.shell.groups.command.timeout";
+  public static final String HADOOP_SECURITY_GROUP_SHELL_COMMAND_TIMEOUT_SECS =
+      "hadoop.security.groups.shell.command.timeout.secs";
```
xiaoyuyao: I think we should keep "hadoop.security.groups.shell.groups.command.timeout". Sorry I was not clear about this when suggesting Configuration#getTimeDuration. It allows admin to specify the time values with suffix like 10s, 1m, 2h. So the ."secs" won't be needed here.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
```diff
@@ -51,14 +52,16 @@
       LoggerFactory.getLogger(ShellBasedUnixGroupsMapping.class);
 
   private long timeout = 0L;
+  private final List<String> emptyGroupsList = new LinkedList<>();
```
xiaoyuyao: This can be static.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/CommonConfigurationKeysPublic.java
QwertyManiac: I'd just felt it would be clearer to have '.secs' in it indicating the minimum unit type despite the parser. I've removed that in the latest commit and added some examples into its doc-text in the xml.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
QwertyManiac: Done in the latest commit.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/security/TestShellBasedUnixGroupsMapping.java
```diff
@@ -22,19 +22,32 @@
...
-  private static final Log LOG =
+  private static final Log TESTLOG =
       LogFactory.getLog(TestShellBasedUnixGroupsMapping.class);
 
+  private final GenericTestUtils.LogCapturer shellMappingLog =
+      GenericTestUtils.LogCapturer.captureLogs(
+          ShellBasedUnixGroupsMapping.LOG);
+
+  private static final boolean WINDOWS =
+      (Shell.osType == Shell.OSType.OS_TYPE_WIN);
```
jojochuang: I think you can just use Shell.WINDOWS instead.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
```diff
@@ -225,8 +287,16 @@ protected ShellCommandExecutor createGroupIDExecutor(String userName) {
...
-        throw new PartialGroupNameException("can't execute the shell command to"
-        + " get the list of group id for user '" + userName + "'", ioe);
+        String message =
+            "Can't execute the shell command to " +
+            "get the list of group id for user '" + userName + "'";
+        if (exec2.isTimedOut()) {
+          message +=
+              " because of the command taking longer than " +
+              "the configured timeout: " + timeout + " seconds";
+          throw new PartialGroupNameException(message);
```
jojochuang: Maybe this line is not needed if it will throw the exception anyway? The only difference is that it will not carry the original exception

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/security/TestShellBasedUnixGroupsMapping.java
QwertyManiac: Thank you, addressed in new commit.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
QwertyManiac: Thanks, addressed in new commit. I just felt the timeout exception may look weird, but I've dropped the line so we can be consistent in exposing the exception at all times.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ShellBasedUnixGroupsMapping.java
```diff
@@ -133,8 +177,26 @@ protected ShellCommandExecutor createGroupIDExecutor(String userName) {
...
+      // If its a shell executor timeout, indicate so in the message
+      // but treat the result as empty instead of throwing it up,
+      // similar to how partial resolution failures are handled above
+      if (executor.isTimedOut()) {
+        LOG.warn(
+            "Unable to return groups for user '{}' as shell group lookup " +
+            "command '{}' ran longer than the configured timeout limit of " +
+            "{} seconds.",
+            user,
+            Arrays.asList(executor.getExecString()),
```
jojochuang: Reviewed the patch again, and I think it's almost ready.
Here's one nit regarding the warning message here: it would print something like "Unable to return groups for user 'foobarnonexistinguser' as shell group lookup command '[sleep, 2]'". But this can be confusing as the command is not run with brackets and comas.

jojochuang: I am +1 pending this and Jenkins precommit build. Somehow Jenkins is never run for your patches.

QwertyManiac: Thank you again @jojochuang. I've added a new change here addressing your comments. I've also uploaded the full patch directly to JIRA to trigger a build test.

QwertyManiac:
Done via e8694deb6ad180449f8ce6c1c8b4f84873c0587a.

PR #162, opened 2016-11-16
HDFS-11148. Update DataNode to use StorageLocationChecker at startup.

arp7:
Change-Id: I664e5c719921ef5e0891bc392f37ee67639a8660

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DataNode.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DataNode.java
@@ -384,7 +384,7 @@ public class DataNode extends ReconfigurableBase
       .availableProcessors();
   private static final double CONGESTION_RATIO = 1.5;
   private DiskBalancer diskBalancer;
-  
+
   @Nullable
   private final StorageLocationChecker storageLocationChecker;
 
@@ -1940,7 +1940,7 @@ public class DataNode extends ReconfigurableBase
     if (storageLocationChecker != null) {
       storageLocationChecker.shutdownAndWait(1, TimeUnit.SECONDS);
     }
-    
+
     if (pauseMonitor != null) {
       pauseMonitor.stop();
     }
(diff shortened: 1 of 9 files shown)


hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DataNode.java
```diff
@@ -1935,6 +1937,10 @@ public void shutdown() {
       }
     }
 
+    if (storageLocationChecker != null) {
+      storageLocationChecker.shutdownAndWait(10, TimeUnit.SECONDS);
```
anuengineer: Something for you to consider, most of the other threads seems to shutdown much quicker, do we need to wait 10 seconds ?

arp7: Good point. We can exit much quicker since a stalled disk check need not block shutdown at all. I'll change it to 1 second.

hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/server/datanode/TestDataDirs.java
```diff
@@ -96,26 +91,4 @@ public void testDataDirParsing() throws Throwable {
     assertThat(locations.get(1).getStorageType(), is(StorageType.DISK));
     assertThat(locations.get(1).getUri(), is(dir1.toURI()));
   }
-
-  @Test(timeout = 30000)
-  public void testDataDirValidation() throws Throwable {
```
arp7: Yes there's new tests in TestStorageLocationChecker added by HDFS-11119 which cover this better with different combinations of failed/healthy storages.

anuengineer:
Arpit, thanks for updating the patch. +1.  LGTM.

elek:

aajisaka:
This issue has been fixed. Closing.

PR #163, opened 2016-11-18
HADOOP-13227 outputstream specification

steveloughran:
WiP specification of output streams.

Alongside the docs, this will include tests and tightening of behaviours of output streams which don't comply with the `java.io.OutputStream` specification or other parts of the specification (i.e. the extra methods are inconsistent with HDFS). Object stores are special; their behaviours will be documented and (unreconcilable) differences with filesystems handled in tests.

Note that HDFS does not currently follow `java.io.OutputStream.write(int)`'s required behaviour: it will *not* fail on a write on a closed stream. This will have to be corrected


elek:

steveloughran:
@elek  that feature doesn't come yet

PR #164, opened 2016-11-22
YARN-5924 - Resource Manager fails to load state with InvalidProtocolBufferException

ameks94:
The solution is to catch "InvalidProtocolBufferException", show warning and remove application's folder that contains invalid data to prevent RM restart failure. 

Additionally, I've added catch for other exceptions that can appear during recovering of the specific application, to avoid RM failure even if the only one application's state can't be loaded.


ameks94:
Update PR to fix the checkstyle and whitespace tests failure.

ameks94:
I realized that current solution is not good (to allow RM's launch even with broken app's data).
It's better to crash RM in case application's file with app's state is broken. This case we can specify more detailed information about which file is broken (Maybe to give the recommendation to remove application's folder with broken data to allow RM to be launched successfully)
Second, the most important part of the fix should be to find the reason of file's crashing and to find the way to prevent file's crash.

PR #170, opened 2016-12-06
HDFS-10930

xiaoyuyao:
HDFS-10930.10


xiaoyuyao:
close as change commit to trunk.

PR #171, opened 2016-12-09
HDFS-11227: Set read timeout for peer

subahugu:
Read timeout is not set with peer in org.apache.hadoop.hdfs.BlockReaderFactory, so BlockReader read doesn't timeout.

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/BlockReaderFactory.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/BlockReaderFactory.java
@@ -471,6 +471,10 @@ public class BlockReaderFactory implements ShortCircuitReplicaCreator {
       if (curPeer == null) break;
       if (curPeer.fromCache) remainingCacheTries--;
       DomainPeer peer = (DomainPeer)curPeer.peer;
+      try {
+        peer.setReadTimeout(conf.socketTimeout);
+      }
+      catch(IOException e){}
       Slot slot = null;
       ShortCircuitCache cache = clientContext.getShortCircuitCache();
       try {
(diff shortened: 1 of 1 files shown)


elek:

hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/BlockReaderFactory.java
```diff
@@ -471,6 +471,10 @@ public ShortCircuitReplicaInfo createShortCircuitReplicaInfo() {
       if (curPeer == null) break;
       if (curPeer.fromCache) remainingCacheTries--;
       DomainPeer peer = (DomainPeer)curPeer.peer;
+      try {
+        peer.setReadTimeout(conf.socketTimeout);
+      }
+      catch(IOException e){}
```
jojochuang: According to the javadoc, it should return a null if it is unable to establish a connection. I think it should also return a null in this case as well.

hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/BlockReaderFactory.java
```diff
@@ -486,6 +490,10 @@ public ShortCircuitReplicaInfo createShortCircuitReplicaInfo() {
           curPeer = nextDomainPeer();
           if (curPeer == null) break;
           peer = (DomainPeer)curPeer.peer;
+          try {
+            peer.setReadTimeout(conf.socketTimeout);
+          }
+          catch(IOException e){}
```
jojochuang: Return null upon exception

hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/BlockReaderFactory.java
```diff
@@ -743,6 +753,10 @@ private BlockReaderPeer nextDomainPeer() {
     if (remainingCacheTries > 0) {
       Peer peer = clientContext.getPeerCache().get(datanode, true);
       if (peer != null) {
+        try {
+          peer.setReadTimeout(conf.socketTimeout);
+        }
+        catch(IOException e){}
```
jojochuang: return null upon exception

PR #172, opened 2016-12-12
HDFS-11234: Made the socket buffer size configurable with the config …

subahugu:
…node fs.hdfs.data.socket.size to be set in core-site.xml. If the node is not found, the default value is -1, so socket buffer size would not be set.


elek:

PR #173, opened 2016-12-15
Details for ADLS connectivity using Client Keys

slachterman:
I did not find the token endpoint as described in the previous version of this document, I found the URL from the ADLS REST API documentation https://docs.microsoft.com/en-us/azure/data-lake-store/data-lake-store-get-started-rest-api

Added dfs.adls.oauth2.access.token.provider.type, I saw errors regarding a missing dfs.adls.oauth2.access.token.provider value without this property set.


elek:

PR #183, opened 2017-01-25
HADOOP-13075 adding sse-c and sse-kms

orngejaket:
Updating fedecz changes to work with changes in trunk.  Refactored methods to be clearer as to purpose and added more integration tests.

### DIFF
--- a/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
+++ b/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
@@ -237,6 +237,12 @@ public class S3AFileSystem extends FileSystem {
         throw new IOException("SSE-C is enable and no encryption key " +
           "is provided.");
       }
+      if(S3AEncryptionMethods.SSE_S3.equals(serverSideEncryptionAlgorithm) &&
+          StringUtils.isNotBlank(S3AUtils.getServerSideEncryptionKey(
+            getConf()))) {
+        throw new IOException("SSE-S3 is configured and an encryption key is " +
+          "provided");
+      }
       LOG.debug("Using encryption {}", serverSideEncryptionAlgorithm);
       inputPolicy = S3AInputPolicy.getPolicy(
           conf.getTrimmed(INPUT_FADVISE, INPUT_FADV_NORMAL));
--- a/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
+++ b/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
@@ -234,7 +234,7 @@ public class S3AFileSystem extends FileSystem {
           conf.getTrimmed(SERVER_SIDE_ENCRYPTION_ALGORITHM));
       if(S3AEncryptionMethods.SSE_C.equals(serverSideEncryptionAlgorithm) &&
           StringUtils.isBlank(S3AUtils.getServerSideEncryptionKey(getConf()))) {
-        throw new IOException("SSE-C is enable and no encryption key " +
+        throw new IOException("SSE-C is enabled and no encryption key " +
           "is provided.");
       }
       if(S3AEncryptionMethods.SSE_S3.equals(serverSideEncryptionAlgorithm) &&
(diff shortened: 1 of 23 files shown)


orngejaket:
Accidently included others commits due to trying to resolve unpushable commit.  Update to the pull request is in e7303f9eff316d3d4329de9011cfd4348ac97a23.  Will correct this tomorrow and resubmit cleaner patch.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -227,8 +230,13 @@ public StorageStatistics provide() {
 
       initMultipartUploads(conf);
 
-      serverSideEncryptionAlgorithm =
-          conf.getTrimmed(SERVER_SIDE_ENCRYPTION_ALGORITHM);
+      serverSideEncryptionAlgorithm = S3AEncryptionMethods.getMethod(
+          conf.getTrimmed(SERVER_SIDE_ENCRYPTION_ALGORITHM));
+      if(S3AEncryptionMethods.SSE_C.equals(serverSideEncryptionAlgorithm) &&
+          StringUtils.isBlank(S3AUtils.getServerSideEncryptionKey(getConf()))) {
+        throw new IOException("SSE-C is enable and no encryption key " +
```
steveloughran: "is enabled"

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -1788,6 +1811,72 @@ public void progressChanged(ProgressEvent progressEvent) {
     }
   }
 
+  protected void setOptionalMultipartUploadRequestParameters(
+      InitiateMultipartUploadRequest req) {
+    if(S3AEncryptionMethods.SSE_KMS.equals(serverSideEncryptionAlgorithm)) {
```
steveloughran: I'd like to see case: statements here; it may help readability

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -1953,6 +2042,9 @@ public String toString() {
           .append(serverSideEncryptionAlgorithm)
           .append('\'');
     }
+    if (S3AUtils.getServerSideEncryptionKey(getConf()) != null) {
```
steveloughran: just cut this

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
```diff
@@ -98,24 +101,30 @@
...
-    this.bucket = bucket;
-    this.key = key;
+  public S3AInputStream(S3ObjectAttributes s3Attributes,
+                        long contentLength,
+                        AmazonS3 client,
+                        FileSystem.Statistics stats,
+                        S3AInstrumentation instrumentation,
+                        long readahead,
+                        S3AInputPolicy inputPolicy) {
+    Preconditions.checkArgument(
```
steveloughran: I'd rather these lines reverted to the previous ones, even if they are > 80 chars long. That is a soft limit on style checks: we break it if it keeps the code looking nice

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ObjectAttributes.java
```diff
@@ -0,0 +1,59 @@
...
+
+package org.apache.hadoop.fs.s3a;
+
+/**
+ * This class is only a holder for bucket, key, SSE Algorithm and SSE key
+ * attributes. It is only used in {@link S3AInputStream}
+ * as a way to reduce parameters being passed
+ * to the constructor of such class.
+ */
+public class S3ObjectAttributes {
```
steveloughran: If this can be made package private, I'd like it to be. Otherwise it must be tagged
{code}
@InterfaceAudience.Private
{code}

steveloughran:
Overall, production code looks good —just some minor tweaks needed. Test code all LGTM, as do docs. As stated on the JIRA, I'd really like tests which verify that sse-c credentials are making it up. Anything we can do to expand coverage of the actual S3 state helps the users, and it helps catch any regressions in our code or the SDK library.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
orngejaket: I assume that I can break that convention elsewhere?   I've got several lines that come in at 82-84 chars.

orngejaket:
> As stated on the JIRA, I'd really like tests which verify that sse-c credentials are making it up. 

The ITestS2AEncryptionSSEC tests currently just check to make sure that md5 of the encryption key is filled in.  I'll change it to do a md5 on the key to verify it.  The AWS UI won't show those files as encrypted in the properties or metadata.  But the Java SDK can see if that one field is filled in with the md5.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -1788,6 +1817,83 @@ public void progressChanged(ProgressEvent progressEvent) {
...
   }
 
+  protected void setOptionalMultipartUploadRequestParameters(
+      InitiateMultipartUploadRequest req) {
+    switch (serverSideEncryptionAlgorithm) {
+    case SSE_KMS:
+      req.setSSEAwsKeyManagementParams(generateSSEAwsKeyParams());
+      break;
+    case SSE_C:
+      if (StringUtils.isNotBlank(S3AUtils.getServerSideEncryptionKey(getConf()))) {
```
steveloughran: S3aUtils is already statically imported; no need to qualify.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
```diff
@@ -98,24 +101,26 @@
...
-    this.bucket = bucket;
-    this.key = key;
+  public S3AInputStream(S3ObjectAttributes s3Attributes,
+                        long contentLength,
+                        AmazonS3 client,
+                        FileSystem.Statistics stats,
+                        S3AInstrumentation instrumentation,
+                        long readahead,
+                        S3AInputPolicy inputPolicy) {
+    Preconditions.checkArgument(StringUtils.isNotEmpty(s3Attributes.getBucket()), "No Bucket");
```
steveloughran: I wouldn't adjust the indentation, because it amplifies the diff and makes merging harder. Lines 105-110 should be the same as the original.

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/ITestS3AEncryptionAlgorithmValidation.java
```diff
@@ -0,0 +1,158 @@
...
+
+  }
+
+  @Test
+  public void testEncryptionAlgorithmSSECWithNoEncryptionKey() throws
+    Throwable {
+    expectedException.expect(IllegalArgumentException.class);
+    expectedException.expectMessage("The value of property " +
+        "fs.s3a.server-side-encryption-key must not be null");
+
```
steveloughran: We're generally moving towards `LambdaTestUtils.intercept()`, because if a closure fails, intercept() will print what came back. Less important for voids though, and on branch-2 & java7, not so compelling.

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/ITestS3AEncryptionAlgorithmValidation.java
```diff
@@ -0,0 +1,158 @@
...
+    assertEquals("wrong filesystem of " + fsURI,
+        contract.getScheme(), fsURI.getScheme());
+    fileSystem.initialize(fsURI, conf);
+  }
+
+  @Test
+  public void testEncryptionAlgorithmSSECWithBlankEncryptionKey() throws
+    Throwable {
+    expectedException.expect(IOException.class);
+    expectedException.expectMessage("SSE-C is enabled and no " +
```
steveloughran: I do like shared constants here, with the constant in the production code, test reading it. Stops the tests being brittle to change in the message. A flaw with expectMessage is that it looks for the whole string, doesn't it; again, we prefer a .contains() as its not brittle to exceptions adding more diags

orngejaket:
Closing because changes are already pushed in.

PR #186, opened 2017-02-04
HDFS-11391 Numeric usernames do no work with WebHDFS FS

pvillard31:
In HDFS-4983, a property has been introduced to configure the pattern
validating name of users interacting with WebHDFS because default
pattern was excluding names starting with numbers.

Problem is that this fix works only for read access. In case of write
access against data node, the default pattern is still applied whatever
the configuration is.

This PR fixes the web handler running in data nodes.


pvillard31:
Committed as part of 8e53f2b9b08560bf4f8e81e697063277dbdc68f9. Closing.

PR #188, opened 2017-02-06
YARN-6151. FS Preemption doesn't filter out queues which cannot be pr…

flyrain:
…eempted.


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSQueue.java
```diff
@@ -236,6 +236,29 @@ public void setFairSharePreemptionThreshold(float fairSharePreemptionThreshold)
   }
 
   /**
+   * Recursively check if the queue can be preempted based on whether the
+   * resource usage is greater than fair share.
+   *
+   * @return true if the queue can be preempted
+   */
+  public boolean canBePreempted() {
+    assert parent != null;
```
kambatla: Why is this necessary?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFairScheduler.java
```diff
@@ -2036,10 +2036,10 @@ public void testPreemptionIsNotDelayedToNextRound() throws Exception {
...
     assertEquals(3277, toPreempt.getMemorySize());
 
-    // verify if the 3 containers required by queueA2 are preempted in the same
+    // verify if the 4 containers required by queueA2 are preempted in the same
     // round
     scheduler.preemptResources(toPreempt);
-    assertEquals(3, scheduler.getSchedulerApp(app1).getPreemptionContainers()
+    assertEquals(4, scheduler.getSchedulerApp(app1).getPreemptionContainers()
         .size());
   }
```
kambatla: Can we add a new test that verifies the exact scenario in the JIRA description?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSQueue.java
```diff
@@ -236,6 +236,29 @@ public void setFairSharePreemptionThreshold(float fairSharePreemptionThreshold)
   }
 
   /**
+   * Recursively check if the queue can be preempted based on whether the
+   * resource usage is greater than fair share.
+   *
+   * @return true if the queue can be preempted
+   */
+  public boolean canBePreempted() {
```
kambatla: Should the check of the allowPreemptionFrom flag also be part of this method?

flyrain: It should be, but allowPreemptionFrom is introduced after 2.8.x.

kambatla: Aah, I keep forgetting branch-2.8 was cut years ago. :(

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFairScheduler.java
```diff
@@ -2043,6 +2043,78 @@ public void testPreemptionIsNotDelayedToNextRound() throws Exception {
...
 
+  @Test
+  public void testPreemptionFilterOutNonPreemptableQueues() throws Exception {
```
kambatla: Can we add this test to TestFairSchedulerPreemption instead?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSQueue.java
```diff
@@ -242,7 +244,9 @@ public void setFairSharePreemptionThreshold(float fairSharePreemptionThreshold)
    * @return true if the queue can be preempted
    */
   public boolean canBePreempted() {
-    assert parent != null;
+    Preconditions.checkNotNull(parent, "Parent queue can't be null since"
```
kambatla: Maybe, we could make this message more clear. "Parent queue is null. Looks like we are checking if root can be preempted."

Alternatively, can we make the if check (parent != null && ...)? That way, else would capture the null case and things should work fine?

flyrain:
Committed.

PR #190, opened 2017-02-09
Huichun/mx net

lhcxx:
https://github.com/apache/hadoop/compare/trunk...lhcxx:huichun-MXNet?expand=1#files_bucket


elek:

PR #195, opened 2017-02-16
HDFS-11421. Make WebHDFS' ACLs RegEx configurable. (harsh)

QwertyManiac:
- Introduced a new config key `dfs.webhdfs.acl.provider.permission.pattern` with value set to existing default ACL regex
- Added pattern-from-config setting points for `AclPermissionParam` to `NameNodeHttpServer`, `WebHdfsFileSystem` and `WebHdfsHandler`, akin to how `UserParam` is set today at these three places
- Added setters/getters for custom pattern to `AclPermissionParam`
- Added a new test for `AclPermissionParam` custom patterns that exercises numeric usernames and group names with `@` characters
- Extended existing `UserParam` configurability tests to cover ACL modification with numeric usernames and group names with `@` characters over a custom pattern
- Ran existing relevant tests, which continue to pass


QwertyManiac:
Done via e24ed47d9a19f34a4dd8d4bad9b5c78ca3dd1c2e on trunk.

PR #196, opened 2017-02-18
YARN-6194. Cluster capacity in SchedulingPolicy is updated only on allocation file reload

flyrain:
This patch passes the ClusterNodeTracker instead of ClusterResource into the DRF policy.

### DIFF
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSContext.java
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSContext.java
@@ -17,6 +17,8 @@
  */
 package org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair;
 
+import org.apache.hadoop.yarn.api.records.Resource;
+
 /**
  * Helper class that holds basic information to be passed around
  * FairScheduler classes. Think of this as a glorified map that holds key
@@ -27,28 +29,37 @@ public class FSContext {
   private boolean preemptionEnabled = false;
   private float preemptionUtilizationThreshold;
   private FSStarvedApps starvedApps;
+  private FairScheduler scheduler;
 
-  public boolean isPreemptionEnabled() {
+  FSContext(FairScheduler scheduler) {
+    this.scheduler = scheduler;
+  }
+
+  boolean isPreemptionEnabled() {
     return preemptionEnabled;
   }
 
-  public void setPreemptionEnabled() {
+  void setPreemptionEnabled() {
     this.preemptionEnabled = true;
     if (starvedApps == null) {
       starvedApps = new FSStarvedApps();
     }
   }
 
-  public FSStarvedApps getStarvedApps() {
+  FSStarvedApps getStarvedApps() {
     return starvedApps;
   }
 
-  public float getPreemptionUtilizationThreshold() {
+  float getPreemptionUtilizationThreshold() {
     return preemptionUtilizationThreshold;
   }
 
-  public void setPreemptionUtilizationThreshold(
+  void setPreemptionUtilizationThreshold(
       float preemptionUtilizationThreshold) {
     this.preemptionUtilizationThreshold = preemptionUtilizationThreshold;
+  }
+
+  public Resource getClusterResource() {
+    return scheduler.getClusterResource();
   }
 }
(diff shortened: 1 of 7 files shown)


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSContext.java
```diff
@@ -27,28 +29,37 @@
...
   }
 
-  public void setPreemptionUtilizationThreshold(
+  void setPreemptionUtilizationThreshold(
       float preemptionUtilizationThreshold) {
     this.preemptionUtilizationThreshold = preemptionUtilizationThreshold;
   }
+
+  public Resource getClusterResource() {
+    return scheduler.getClusterResource();
```
kambatla: This looks okay for now, but this allows a scheduling policy to modify the overall cluster's resources. Making a copy could be expensive, as this is called on every compare call.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/policies/TestDominantResourceFairnessPolicy.java
```diff
@@ -40,7 +43,10 @@
   private Comparator<Schedulable> createComparator(int clusterMem,
       int clusterCpu) {
     DominantResourceFairnessPolicy policy = new DominantResourceFairnessPolicy();
-    policy.initialize(BuilderUtils.newResource(clusterMem, clusterCpu));
+    FSContext fsContext = mock(FSContext.class);
+    when(fsContext.getClusterResource()).
+        thenReturn(BuilderUtils.newResource(clusterMem, clusterCpu));
```
kambatla: Let us use Resources.create instead of BuilterUtils.newInstance.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFairScheduler.java
```diff
@@ -3369,7 +3369,48 @@ public void testBasicDRFWithQueues() throws Exception {
     scheduler.handle(updateEvent);
     Assert.assertEquals(1, app2.getLiveContainers().size());
   }
-  
+
+  @Test
+  public void testDRFWithClusterResourceChanges() throws Exception {
```
kambatla: Using a "real" scheduler with mock nodes seems excessive for this. Can we just mock the scheduler and context? Also, this might be a good test for TestDRF than TestFairScheduler.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/SchedulingPolicy.java
```diff
@@ -92,7 +92,7 @@ public static SchedulingPolicy parse(String policy)
     return getInstance(clazz);
   }
   
-  public void initialize(Resource clusterCapacity) {}
+  public void initialize(FSContext fsContext) {}
```
kambatla: Since this method is in an @Public class, let us add a new method and deprecate the old method.

Let us also add javadoc for both methods.

flyrain:
Thanks Karthik for the review. Push a new commit for your comments.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/SchedulingPolicy.java
```diff
@@ -91,7 +91,23 @@ public static SchedulingPolicy parse(String policy)
     }
     return getInstance(clazz);
   }
-  
+
+  /**
+   * Initialize the scheduling policy with cluster resources. Deprecated since
```
kambatla: Want to add a @deprecated tag in the javadoc and point to what to use?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/SchedulingPolicy.java
```diff
@@ -91,7 +91,23 @@ public static SchedulingPolicy parse(String policy)
...
+   * it doesn't track cluster resource changes.
+   *
+   * @param clusterCapacity cluster resources
+   */
+  @Deprecated
+  public void initialize(Resource clusterCapacity) {}
+
+  /**
+   * Initialize the scheduling policy with a {@link FSContext} object which can
```
kambatla: In the future, different policies could use different information from the FSContext. Maybe, instead of referring to it as the only thing, say something like "FSContext, which has a pointer to the cluster resources among other information.

flyrain:
Committed

flyrain:
Committed

PR #198, opened 2017-03-01
YARN-6261 - Catch user with no group when getting queue from mapping …

pvillard31:
…definition


elek:

PR #201, opened 2017-03-11
YARN-5829 Reserve containers for preemptors

szegedim:
YARN-5829 Reserve containers for preemptors.
The change does the following.
1. When a container is selected for preemption, it stores the preemptor next to the container.
2. When a container is finally completed it creates a reservation for the preemptor application.
3. On the next node update we check first reservations, then preemption reservations. They are satisfied before any other request, even if those have higher priorities. This is used tho avoid stealing preempted containers.
4. It also disables standard reservations because I noticed that the preempted applications reserve nodes for themselves stealing back the preempted resources.
5. If a preemptor is deleted before the preemption or after the preemption, the assignment won't happen at node update.
6. If more containers were preempted to satisfy partial reservations, delete the reservation, if no more resource is needed.


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSAppAttempt.java
```diff
@@ -831,7 +831,11 @@ private Resource assignContainer(
     }
 
     // The desired container won't fit here, so reserve
-    if (isReservable(capability) &&
+    // Important: Do not enable this kind of reservation in case of preemption,
+    // since the preempted application will re-reserve the preempted resources
+    // and prevent the starving application to get them
+    if (!scheduler.getConf().getPreemptionEnabled() &&
```
kambatla: This seems very bad. What happens if preemption is enabled, but no containers are preempted. However, if there are apps requesting large containers that don't fit in the free resources created by smaller containers finishing, these apps end up being starved forever.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,30 +18,52 @@
...
-import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.ConcurrentSkipListMap;
 
 @Private
 @Unstable
 public class FSSchedulerNode extends SchedulerNode {
 
   private static final Log LOG = LogFactory.getLog(FSSchedulerNode.class);
-
+  private static final Comparator<RMContainer> comparator =
```
kambatla: Why do we need this comparator?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,30 +18,52 @@
...
-  private final Set<RMContainer> containersForPreemption =
-      new ConcurrentSkipListSet<>();
+  // Stores preemption list until the container is completed
+  @VisibleForTesting
+  final Map<RMContainer, FSAppAttempt> containersForPreemption =
+      new ConcurrentSkipListMap<>(comparator);
+
+  // Stores preemption list after the container is completed before assigned
+  @VisibleForTesting
+  Map<FSAppAttempt, Resource> reservedApp =
```
kambatla: This map can be final as well.

kambatla: Instead of calling this reservedApp which can be confused with reservedAppSchedulable, how about we call this resourcesPreemptedPerApp?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -118,23 +178,56 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
    *
    * @param containers container to mark
    */
-  void addContainersForPreemption(Collection<RMContainer> containers) {
-    containersForPreemption.addAll(containers);
+  void addContainersForPreemption(Collection<RMContainer> containers,
+                                  FSAppAttempt appAttempt) {
+    for(RMContainer container : containers) {
+      containersForPreemption.put(container, appAttempt);
```
kambatla: We should add the entry to the other map also here. That way, we count allocations prior to the preemptions also against the same app.

kambatla: Actually, if we do that, the first list of containers need not be a map. It could continue to be a set.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -118,23 +178,56 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
...
 
   /**
    * Remove container from the set of containers marked for preemption.
+   * Reserve the preempted resources for the app that requested
+   * the preemption.
    *
    * @param container container to remove
    */
-  void removeContainerForPreemption(RMContainer container) {
+  private synchronized void removeContainerForPreemption(RMContainer container) {
```
kambatla: If we add the entry to the map when we add container, this method doesn't need to be changed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -118,23 +178,56 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
...
+   * @param containerId ID of container to be released.
+   * @param releasedByNode whether the release originates from a node update.
+   */
+  @Override
+  public synchronized void releaseContainer(ContainerId containerId,
+                                            boolean releasedByNode) {
+    RMContainer container = getContainer(containerId);
+    super.releaseContainer(containerId, releasedByNode);
+    if (container != null) {
+      removeContainerForPreemption(container);
```
kambatla: This is nice to see. We should have done this to begin with.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -110,6 +132,44 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
   }
 
   /**
+   * List reserved resources after preemption and assign them to the
+   * appropriate applications in a FIFO order.
+   * @return if any resources were allocated
+   */
+  synchronized boolean assignContainersToPreemptionReservees() {
```
kambatla: I am slightly concerned about this splitting the container allocation code across too many places. The alternative is to do this in FairScheduler but maintain the map here, which is also ugly. Would it make any sense to hold this information in FSContext? 

Just calling out we need to think through this carefully.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
```diff
@@ -0,0 +1,360 @@
...
+import java.util.Collections;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test scheduler node, especially preemption reservations.
```
kambatla: Should the tests related to preemptions be in TestFairSchedulerPreemption instead?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: The unit test failed without it.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: Sure, done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: The reason I brought it here is that we need to run it synchronized to make sure the list does not change. The preemption reservation data belongs here I think. What I could do is that do the assignContainersToPreemptionReservees tasks in fair scheduler and expose the lock to that function.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: java.lang.ClassCastException: org.apache.hadoop.yarn.server.resourcemanager.rmcontainer.RMContainer$$EnhancerByMockitoWithCGLIB$$4607190f cannot be cast to java.lang.Comparable

	at java.util.concurrent.ConcurrentSkipListMap.cpr(ConcurrentSkipListMap.java:655)
	at java.util.concurrent.ConcurrentSkipListMap.doGet(ConcurrentSkipListMap.java:794)
	at java.util.concurrent.ConcurrentSkipListMap.get(ConcurrentSkipListMap.java:1546)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.FSSchedulerNode.removeContainerForPreemption(FSSchedulerNode.java:203)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.FSSchedulerNode.releaseContainer(FSSchedulerNode.java:230)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.TestFSSchedulerNode.testPreemptionToCompletedApp(TestFSSchedulerNode.java:302)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:309)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:160)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:51)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:237)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
szegedim: Only, if all the preemption classes are used by the test. This is a real unit test for the class FSSchedulerNode. In this case the unit test is restricted to mocks and only FSSchedulerNode is a real class. All in all, I see this as a test of a limited subset of preemption functionality restricted to FSSchedulerNode, so it should be in a class named TestFSSchedulerNode. If you want me to add additional functional tests to TestFairSchedulerPreemption, I am happy to do that.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: Thanks!

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: Fixed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSAppAttempt.java
szegedim: I improved this by giving higher priority to preempted containers than reservation. Child queues may pass containers to each other this way resolving fairness starvation. Please review the updated change.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -131,10 +203,58 @@ void addContainersForPreemption(Collection<RMContainer> containers) {
...
+   * @param rmContainer Allocated container
+   * @param launchedOnNode True if the container has been launched
+   */
+  @Override
+  protected synchronized void allocateContainer(RMContainer rmContainer,
+                                                boolean launchedOnNode) {
+    super.allocateContainer(rmContainer, launchedOnNode);
+    Resource allocated = rmContainer.getAllocatedResource();
+    if (!Resources.isNone(allocated)) {
+      for (FSAppAttempt app: resourcesPreemptedPerApp.keySet()) {
```
kambatla: Given this is a map, why aren't we just doing a map.get()?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -131,10 +203,58 @@ void addContainersForPreemption(Collection<RMContainer> containers) {
...
+        if (app.getApplicationAttemptId().equals(
+            rmContainer.getApplicationAttemptId())) {
+          Resource reserved = resourcesPreemptedPerApp.get(app);
+          Resource fulfilled = Resources.componentwiseMin(reserved, allocated);
```
kambatla: This is not necessary if we use subtractFromNonNegative in the following two statements.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
```diff
@@ -93,15 +91,8 @@
...
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Comparator;
-import java.util.EnumSet;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
kambatla: Same comment as above. Let us avoid using wildcard imports

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
```diff
@@ -987,9 +978,19 @@ void attemptScheduling(FSSchedulerNode node) {
...
-
+      // 1. Ensure containers are assigned to the apps that preempted
+      // 2. Check for reserved applications
+      // 3. Schedule if there are no reservations
+
+      // Apps may wait for preempted containers
+      // We have to satisfy these first to avoid cases, when we preempt
+      // a container for A from B and C gets the preempted containers,
+      // when C does not qualify for preemption itself.
+      for (FSAppAttempt app : node.getPreemptionList()) {
```
kambatla: We don't seem to handle the case where an app is to be allocated multiple containers from based on the preempted resources.

Would it help to store the apps in a TreeMap in FSSchedulerNode and have a method that returns the next preempted app to allocate to? If none exist, return null? Also, if this were to work, we might have to revert back to the behavior in an earlier patch: add app/resources to the map only after the container is actually preempted?

kambatla: Also, we are not counting allocated resources/containers against the assignMultiple. We probably should so that we limit the number of non-preemption, non-reserved containers/resources appropriately.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSPreemptionThread.java
```diff
@@ -113,11 +113,6 @@ public void run() {
         List<FSSchedulerNode> potentialNodes = scheduler.getNodeTracker()
             .getNodesByResourceName(rr.getResourceName());
         for (FSSchedulerNode node : potentialNodes) {
-          // TODO (YARN-5829): Attempt to reserve the node for starved app.
-          if (isNodeAlreadyReserved(node, starvedApp)) {
-            continue;
-          }
-
```
szegedim: Indeed, preemption reservation has higher priority.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSPreemptionThread.java
```diff
@@ -164,7 +160,10 @@ private PreemptableContainers identifyContainersToPreemptOnNode(
     containersToCheck.removeAll(node.getContainersForPreemption());
 
     // Initialize potential with unallocated resources
-    Resource potential = Resources.clone(node.getUnallocatedResource());
+    Resource potential = Resources.clone(
+        Resources.componentwiseMax(Resources.none(),
```
szegedim: Yes that is nicer. Fixed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,35 +18,66 @@
...
 import org.apache.hadoop.yarn.server.resourcemanager.rmcontainer.RMContainer;
 import org.apache.hadoop.yarn.server.resourcemanager.rmnode.RMNode;
 import org.apache.hadoop.yarn.server.resourcemanager.scheduler.SchedulerApplicationAttempt;
 import org.apache.hadoop.yarn.server.scheduler.SchedulerRequestKey;
 import org.apache.hadoop.yarn.server.resourcemanager.scheduler.SchedulerNode;
+import org.apache.hadoop.yarn.util.resource.Resources;
 
-import java.util.Collection;
-import java.util.Set;
+import java.util.*;
```
szegedim: Fixed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,35 +18,66 @@
...
 import java.util.concurrent.ConcurrentSkipListSet;
 
 @Private
 @Unstable
 public class FSSchedulerNode extends SchedulerNode {
 
   private static final Log LOG = LogFactory.getLog(FSSchedulerNode.class);
+  private static final Comparator<RMContainer> comparator =
```
szegedim: RMContainer is an interface, it cannot implement a method.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,35 +18,66 @@
...
+    public int compare(RMContainer o1, RMContainer o2) {
+      return Long.compare(o1.getContainerId().getContainerId(),
+          o2.getContainerId().getContainerId());
+    }
+  };
 
   private FSAppAttempt reservedAppSchedulable;
-  private final Set<RMContainer> containersForPreemption =
-      new ConcurrentSkipListSet<>();
+  // Stores preemption list until the container is completed
```
szegedim: All right.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,35 +18,66 @@
...
+  @VisibleForTesting
+  final Set<RMContainer> containersForPreemption =
+      new ConcurrentSkipListSet<>(comparator);
+  // Stores preemption list after the container is completed before assigned
+  @VisibleForTesting
+  final Map<FSAppAttempt, Resource>
+      resourcesPreemptedPerApp = new LinkedHashMap<>();
```
szegedim: All right.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,35 +18,66 @@
...
+  Resource totalResourcesPreempted = Resource.newInstance(0, 0);
```
szegedim: All right.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,35 +18,66 @@
...
   }
 
+  /**
+   * Total amount of reserved resources including reservations and preempted
+   * containers.
+   * @return total resources reserved
+   */
+  Resource getTotalReserved() {
+    Resource totalReserved = getReservedContainer() != null ?
+        Resources.clone(getReservedContainer().getAllocatedResource()) :
```
szegedim: All right.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,35 +18,66 @@
...
+        Resource.newInstance(0, 0);
```
szegedim: Sorry, what is the point cloning none? I think the current one is more straightforward.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -131,10 +203,58 @@ void addContainersForPreemption(Collection<RMContainer> containers) {
 
   /**
    * Remove container from the set of containers marked for preemption.
+   * Reserve the preempted resources for the app that requested
+   * the preemption.
    *
    * @param container container to remove
    */
   void removeContainerForPreemption(RMContainer container) {
```
szegedim: Good catch. Fixed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: It is not indexed by ID but the app.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: Not really, because we would subtract a bigger value from total than from reserved, leaving the state of the map and it's totals inconsistent.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
szegedim: All right.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
szegedim: > Would it help to store the apps in a TreeMap in FSSchedulerNode and have a method that > returns the next preempted app to allocate to? If none exist, return null? Also, if this were > to work, we might have to revert back to the behavior in an earlier patch: add app/resources to the map only after the container is actually preempted?
This approach has a clear advantage, that it allocates any newly appeared container to the preemptors. TreeMap might slow down search a little bit. Let's try to continue with this approach but multiple allocations enabled.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
kambatla: A for loop seems a little wasteful. Would it make any sense to keep another map so this can just be a lookup?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
```diff
@@ -0,0 +1,376 @@
...
+    // Preemption occurs
+    schedulerNode.releaseContainer(containers.get(0).getContainerId(), true);
+    allocateContainers(schedulerNode);
+    assertEquals("Container should be allocated",
+        schedulerNode.getTotalResource(),
+        schedulerNode.getAllocatedResource());
+
+    // Release all containers
+    for (int i = 1; i < containerNum; ++i) {
+      schedulerNode.releaseContainer(containers.get(i).getContainerId(), true);
```
kambatla: Can't you go through the container list for this?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
```diff
@@ -0,0 +1,376 @@
...
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test scheduler node, especially preemption reservations.
+ */
+public class TestFSSchedulerNode {
+  private long containerNum = 0;
```
kambatla: Actually, can't you just use containers.size() instead?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
```diff
@@ -0,0 +1,376 @@
...
+    schedulerNode.releaseContainer(containers.get(1).getContainerId(), true);
+
+    allocateContainers(schedulerNode);
+    assertEquals("Container should be allocated",
+        schedulerNode.getTotalResource(),
+        schedulerNode.getAllocatedResource());
+
+    // Release all containers
+    for (int i = 3; i < containerNum; ++i) {
+      schedulerNode.releaseContainer(containers.get(i).getContainerId(), true);
```
kambatla: Same comment as above.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
kambatla: And, shouldn't you start i with 0?

kambatla: This seems to be in many tests. How about a helper to release all containers?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFairSchedulerPreemption.java
```diff
@@ -294,11 +294,29 @@ private void verifyPreemption(int numStarvedAppContainers)
...
     // Verify the preempted containers are assigned to starvingApp
     assertEquals("Starved app is not assigned the right # of containers",
         numStarvedAppContainers, starvingApp.getLiveContainers().size());
+
+    // Verify the node is not reserved for the starvingApp anymore
+    for (RMNode rmNode : rmNodes) {
+      FSSchedulerNode node = (FSSchedulerNode)
+          scheduler.getNodeTracker().getNode(rmNode.getNodeID());
+      if (node.getContainersForPreemption().size() > 0) {
+        assertTrue(!node.getPreemptionList().keySet().contains(starvingApp));
```
kambatla: assertFalse?

kambatla: assert message.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFairSchedulerPreemption.java
```diff
@@ -294,11 +294,29 @@ private void verifyPreemption(int numStarvedAppContainers)
         8 - 2 * numStarvedAppContainers,
         greedyApp.getQueue().getMetrics().getAggregatePreemptedContainers());
 
+    // Verify the node is reserved for the starvingApp
+    for (RMNode rmNode : rmNodes) {
+      FSSchedulerNode node = (FSSchedulerNode)
+          scheduler.getNodeTracker().getNode(rmNode.getNodeID());
+      if (node.getContainersForPreemption().size() > 0) {
+        assertTrue(node.getPreemptionList().keySet().contains(starvingApp));
```
kambatla: assert message.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
szegedim: All right. Fixed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -38,15 +47,45 @@
 public class FSSchedulerNode extends SchedulerNode {
 
   private static final Log LOG = LogFactory.getLog(FSSchedulerNode.class);
+  private static final Comparator<RMContainer> comparator =
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -18,18 +18,27 @@
...
 import org.apache.hadoop.yarn.server.resourcemanager.scheduler.SchedulerApplicationAttempt;
 import org.apache.hadoop.yarn.server.scheduler.SchedulerRequestKey;
 import org.apache.hadoop.yarn.server.resourcemanager.scheduler.SchedulerNode;
+import org.apache.hadoop.yarn.util.resource.Resources;
 
 import java.util.Collection;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -38,15 +47,45 @@
...
+  };
 
   private FSAppAttempt reservedAppSchedulable;
-  private final Set<RMContainer> containersForPreemption =
-      new ConcurrentSkipListSet<>();
+  // Stores list of containers still to be preempted
+  @VisibleForTesting
+  final Set<RMContainer> containersForPreemption =
+      new ConcurrentSkipListSet<>(comparator);
+  // Stores amount of respources preempted and reserved for each app
```
szegedim: Fixed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -38,15 +47,45 @@
...
 
+  /**
+   * Total amount of reserved resources including reservations and preempted
+   * containers.
+   * @return total resources reserved
+   */
+  Resource getTotalReserved() {
+    Resource totalReserved = getReservedContainer() != null
+        ? Resources.clone(getReservedContainer().getAllocatedResource())
+        : Resource.newInstance(0, 0);
```
szegedim: Fixed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -110,16 +149,57 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
   }
 
   /**
+   * List reserved resources after preemption and assign them to the
+   * appropriate applications in a FIFO order.
+   * It is a good practice to call {@link #cleanupPreemptionList()}
```
szegedim: Fixed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -110,16 +149,57 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
...
+   */
+  synchronized LinkedHashMap<FSAppAttempt, Resource> getPreemptionList() {
+    cleanupPreemptionList();
+    return new LinkedHashMap<>(resourcesPreemptedForApp);
+  }
+
+  /**
+   * Remove apps that have their preemption requests fulfilled
+   */
+  synchronized void cleanupPreemptionList() {
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -110,16 +149,57 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
...
+    Iterator<FSAppAttempt> iterator =
+        resourcesPreemptedForApp.keySet().iterator();
+    while (iterator.hasNext()) {
+      FSAppAttempt app = iterator.next();
+      boolean removeApp = false;
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -110,16 +149,57 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
...
+      if (app.isStopped() || Resources.equals(
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -110,16 +149,57 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
...
+        iterator.remove();
+      }
+    }
+  }
+
+  /**
    * Mark {@code containers} as being considered for preemption so they are
    * not considered again. A call to this requires a corresponding call to
-   * {@link #removeContainerForPreemption} to ensure we do not mark a
+   * removeContainer for preemption to ensure we do not mark a
```
szegedim: Fixed.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
```diff
@@ -968,6 +966,32 @@ private boolean shouldContinueAssigning(int containers,
...
 
+  /**
+   * Assign preempted containers to the applications that have reserved
+   * resources for preempted containers.
+   * @param node Node to check
+   * @return assignment has occurred
+   */
+  static boolean assignPreemptedContainers(FSSchedulerNode node ) {
+    boolean assignedAny = false;
+    node.cleanupPreemptionList();
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
```diff
@@ -968,6 +966,32 @@ private boolean shouldContinueAssigning(int containers,
...
+    for (Entry<FSAppAttempt, Resource> entry :
+        node.getPreemptionList().entrySet()) {
+      FSAppAttempt app = entry.getKey();
+      Resource reserved = Resources.clone(entry.getValue());
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
```diff
@@ -968,6 +966,32 @@ private boolean shouldContinueAssigning(int containers,
...
+      while (!app.isStopped() && !Resources.isNone(reserved)) {
+        Resource assigned = app.assignContainer(node);
+        if (Resources.isNone(assigned)) {
+          break;
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
```diff
@@ -968,6 +966,32 @@ private boolean shouldContinueAssigning(int containers,
...
+        }
+        assignedAny = true;
+        Resources.subtractFromNonNegative(reserved, assigned);
+      }
+    }
+    node.cleanupPreemptionList();
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSAppAttempt.java
```diff
@@ -646,7 +646,6 @@ private Container createContainer(FSSchedulerNode node, Resource capability,
   private boolean reserve(Resource perAllocationResource, FSSchedulerNode node,
       Container reservedContainer, NodeType type,
       SchedulerRequestKey schedulerKey) {
-
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
```diff
@@ -0,0 +1,376 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test scheduler node, especially preemption reservations.
+ */
+public class TestFSSchedulerNode {
+  private long containerNum = 0;
+  private ArrayList<RMContainer> containers = new ArrayList<>();
```
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
```diff
@@ -0,0 +1,376 @@
...
+    when(id.getContainerId()).thenReturn(containerNum);
+    when(containerInner.getResource()).
+        thenReturn(Resources.clone(request));
+    when(containerInner.getId()).thenReturn(id);
+    when(containerInner.getExecutionType()).
+        thenReturn(ExecutionType.GUARANTEED);
+    when(container.getApplicationAttemptId()).thenReturn(appAttemptId);
+    when(container.getContainerId()).thenReturn(id);
+    when(container.getContainer()).thenReturn(containerInner);
+    when(container.getExecutionType()).thenReturn(ExecutionType.GUARANTEED);
```
szegedim: That is for the inner class. Sorry, what do you mean? When?

szegedim: I do have a when in all these cases.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
```diff
@@ -0,0 +1,376 @@
...
+        thenReturn(Resources.clone(request));
+    containers.add(container);
+    containerNum++;
+    return container;
+  }
+
+  private void saturateCluster(FSSchedulerNode schedulerNode) {
+    while (!Resources.isNone(schedulerNode.getUnallocatedResource())) {
+      createDefaultContainer();
+      schedulerNode.allocateContainer(containers.get((int)containerNum - 1));
```
szegedim: I changed to int.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
szegedim: It is not 0 because I release the rest of the containers that were not preempted. That is, I would like to see the actual schedulerNode.releaseContainer call next to the preemption in the same function to see what is actually happening. Another function would hide the ability to see all releases together.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSSchedulerNode.java
szegedim: Same response :-)

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFairSchedulerPreemption.java
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFairSchedulerPreemption.java
szegedim: Done.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/rmcontainer/RMContainer.java
```diff
@@ -42,7 +42,8 @@
  * when resources are being reserved to fill space for a future container 
  * allocation.
  */
-public interface RMContainer extends EventHandler<RMContainerEvent> {
+public interface RMContainer extends EventHandler<RMContainerEvent>,
```
kambatla: We need a corresponding change in RMContainerImpl. It should suffice to implement only RMContainer now. .

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -110,16 +140,56 @@ synchronized FSAppAttempt getReservedAppSchedulable() {
...
+        iterator.remove();
+      }
+    }
+  }
+
+  /**
    * Mark {@code containers} as being considered for preemption so they are
    * not considered again. A call to this requires a corresponding call to
-   * {@link #removeContainerForPreemption} to ensure we do not mark a
+   * {@code releaseContainer} for preemption to ensure we do not mark a
```
kambatla: Should we drop "for preemption" altogether?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -130,11 +200,47 @@ void addContainersForPreemption(Collection<RMContainer> containers) {
...
+   * application.
+   * @param rmContainer Allocated container
+   * @param launchedOnNode True if the container has been launched
+   */
+  @Override
+  protected synchronized void allocateContainer(RMContainer rmContainer,
+                                                boolean launchedOnNode) {
+    super.allocateContainer(rmContainer, launchedOnNode);
+    Resource allocated = rmContainer.getAllocatedResource();
+    if (!Resources.isNone(allocated)) {
```
kambatla: Will we have an empty allocation? If not, should we log an error when we do.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSSchedulerNode.java
```diff
@@ -130,11 +200,47 @@ void addContainersForPreemption(Collection<RMContainer> containers) {
...
+      FSAppAttempt app =
+          appIdToAppMap.get(rmContainer.getApplicationAttemptId());
+      if (app != null) {
```
kambatla: For readability, can we can add a comment either before/after the null check or in the else block.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
```diff
@@ -987,12 +1010,18 @@ void attemptScheduling(FSSchedulerNode node) {
...
+
+      // Apps may wait for preempted containers
+      // We have to satisfy these first to avoid cases, when we preempt
+      // a container for A from B and C gets the preempted containers,
+      // when C does not qualify for preemption itself.
+      boolean validReservation;
+      validReservation = assignPreemptedContainers(node);
       FSAppAttempt reservedAppSchedulable = node.getReservedAppSchedulable();
-      if (reservedAppSchedulable != null) {
+      if (!validReservation && reservedAppSchedulable != null) {
```
kambatla: In lieu of future changes around assigning as many containers as directed by assign-multiple settings, can we not gate further allocations on the result of assignPreemptedContainers.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSAppAttempt.java
```diff
@@ -647,7 +647,12 @@ private boolean reserve(Resource perAllocationResource, FSSchedulerNode node,
       Container reservedContainer, NodeType type,
       SchedulerRequestKey schedulerKey) {
 
-    if (!reservationExceedsThreshold(node, type)) {
+    RMContainer nodeReservedContainer = node.getReservedContainer();
+    boolean reservableForThisApp = nodeReservedContainer == null ||
+        nodeReservedContainer.getApplicationAttemptId()
+            .equals(getApplicationAttemptId());
+    if (!reservationExceedsThreshold(node, type) &&
+        reservableForThisApp) {
```
kambatla: Let us flip the order here, because reservableForThisApp is cheaper to check.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairScheduler.java
```diff
@@ -987,10 +1010,16 @@ void attemptScheduling(FSSchedulerNode node) {
...
+      // 1. Ensure containers are assigned to the apps that preempted
+      // 2. Check for reserved applications
+      // 3. Schedule if there are no reservations
+
+      // Apps may wait for preempted containers
+      // We have to satisfy these first to avoid cases, when we preempt
+      // a container for A from B and C gets the preempted containers,
+      // when C does not qualify for preemption itself.
       boolean validReservation = false;
+      assignPreemptedContainers(node);
```
kambatla: Call to assignPreemptedContainers should go before declaring validReservation.

PR #206, opened 2017-03-24
HADOOP-14235. S3A Path does not understand colon (:) when globbing

kazuyukitanimura:
Hi @steveloughran @liuml07 

I explained the issue at https://issues.apache.org/jira/browse/HADOOP-14235

This pull request fixes the issue and does not break other things as far as I know. (I also ran the unit tests).

Probably, #204 should also fix this issue. This pull request is for a short-term solution in case anyone is interested.


kazuyukitanimura:
Closing PR as stated in Jira

PR #207, opened 2017-03-25
HADOOP-14237. S3A Support Shared Instance Profile Credentials Across All Instances

kazuyukitanimura:
Hi @steveloughran 

Yet another patch that I made a few months back.
I explained the issue at https://issues.apache.org/jira/browse/HADOOP-14237

This pull request is aiming for more open discussions rather than a complete solution. It would be great if you could offer your thoughts.

### DIFF
--- a/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/SharedInstanceProfileCredentialsProvider.java
+++ b/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/SharedInstanceProfileCredentialsProvider.java
@@ -18,10 +18,17 @@
 
 package org.apache.hadoop.fs.s3a;
 
-import com.amazonaws.auth.InstanceProfileCredentialsProvider;
+import com.amazonaws.auth.*;
+import com.amazonaws.AmazonClientException;
+import com.amazonaws.AmazonServiceException;
 
+import org.apache.commons.lang.*;
 import org.apache.hadoop.classification.InterfaceAudience;
 import org.apache.hadoop.classification.InterfaceStability;
+import org.apache.hadoop.conf.*;
+import org.apache.hadoop.fs.*;
+
+import java.io.*;
 
 /**
  * A subclass of {@link InstanceProfileCredentialsProvider} that enforces
@@ -49,6 +56,12 @@ public final class SharedInstanceProfileCredentialsProvider
   private static final SharedInstanceProfileCredentialsProvider INSTANCE =
       new SharedInstanceProfileCredentialsProvider();
 
+  private static final Path s3crednetialPath = new Path("hdfs:///tmp/s3credential.txt");
+
+  private volatile AWSCredentials credentials;
+
+  private static final int maxRetries = 5;
+
   /**
    * Returns the singleton instance.
    *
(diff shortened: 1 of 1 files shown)


hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/SharedInstanceProfileCredentialsProvider.java
```diff
@@ -58,6 +71,84 @@ public static SharedInstanceProfileCredentialsProvider getInstance() {
...
+    for (int retry = 0; retry < maxRetries; retry++) {
+      try {
+        AWSCredentials newCredentials = super.getCredentials();
+        // if this new credentials is different from HDFS write back
+        if (credentials == null || (!newCredentials.getAWSSecretKey().equals(credentials.getAWSSecretKey()))) {
+          credentials = newCredentials;
+          writeCredentialsToHDFS(credentials);
+        }
+        break;
+      } catch (Exception e) {
```
steveloughran: I't use our normal Retry logic here, consider some sleep  + jitter if it really is caused by throttling

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/SharedInstanceProfileCredentialsProvider.java
```diff
@@ -58,6 +71,84 @@ public static SharedInstanceProfileCredentialsProvider getInstance() {
...
+  private void writeCredentialsToHDFS(AWSCredentials credentials) {
+    try {
+      // Simulate atomic write by creating a new s3credential file with random string suffix and rename to s3crednetialPath
+      Path newS3crednetialPath = new Path(s3crednetialPath.toUri() + RandomStringUtils.randomAlphanumeric(8));
+      FileSystem fs = FileSystem.get(new Configuration());
+      BufferedWriter br = new BufferedWriter(new OutputStreamWriter(fs.create(newS3crednetialPath, true)));
+      String accessKey = credentials.getAWSAccessKeyId();
+      String secretKey = credentials.getAWSSecretKey();
+      String token = "";
+      if (credentials instanceof BasicSessionCredentials) {
```
steveloughran: I would only allow session credentials to persist, so as to reduce risk of leakage of persistent secrets

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/SharedInstanceProfileCredentialsProvider.java
```diff
@@ -58,6 +71,84 @@ public static SharedInstanceProfileCredentialsProvider getInstance() {
...
+      br.write(secretKey);
+      br.newLine();
+      br.write(token);
+      br.newLine();
+      br.close();
+      fs.delete(s3crednetialPath, false);
+      fs.rename(newS3crednetialPath, s3crednetialPath);
+    } catch (Exception e) {
+      // ignore write errors
+      // throw new AmazonServiceException("Failed writing S3 credentials from HDFS " + e.getStackTrace());
```
steveloughran: log @ debug at the very least

elek:

PR #208, opened 2017-03-25
HADOOP-14239. S3A Retry Multiple S3 Key Deletion

kazuyukitanimura:
Hi @steveloughran @liuml07 

Sorry for sending may requests.

I explained the problem here https://issues.apache.org/jira/browse/HADOOP-14239

This pull requests recursively retries to delete only S3 keys that are previously failed to delete during the multiple object deletion because aws-java-sdk retry does not help. If it still fails, it will fall back to the single deletion.


kazuyukitanimura:
Closing PR as stated in Jira

PR #219, opened 2017-05-08
YARN-6457  use existing conf object as a resource for sslConf object in WebApps for the builder to use in HttpServer2

sanjaypujare:
use the passed conf object as a resource in local sslConf so it can be overridden


elek:

aajisaka:
https://issues.apache.org/jira/browse/YARN-6457 has been fixed. Closing.

PR #239, opened 2017-06-19
YARN-6674 Add memory cgroup settings for opportunistic containers

szegedim:
YARN-6674 Add memory cgroup settings for opportunistic containers


szegedim:
I created #240

PR #246, opened 2017-07-07
HADOOP-14539. Move commons logging APIs over to slf4j in hadoop-common.

wenxinhe:


hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/SecurityUtil.java
```diff
@@ -53,6 +51,8 @@
 
 
 //this will need to be replaced someday when there is a suitable replacement
+import org.slf4j.Logger;
```
aajisaka: Would you move the new imports above the comment "this will need to ...."?

wenxinhe: OK, fix it in d04928e.

PR #252, opened 2017-07-24
replication example revised

yonatan-py:
the placing of the third replica is confusing, is it put on the same rack as the first replica, or are the second and third put on a different replica.


elek:

aajisaka:
-1

2nd replica is written on the different rack from the 1st replica in the normal case. Please see BlockPlacementPolicyDefault.java for detail.

aajisaka:
Closing. If you disagree, please feel free to reopen this PR.

PR #254, opened 2017-07-25
Liumihust patch add null pointer exception in FsDatasetImpl.checkBlock()

liumihust:
Additional PR for the the same bug


elek:

PR #255, opened 2017-07-27
[Parallel Block Copy for DiskBalancer]Update DiskBalancer.java

liumihust:
The default plan executing is in sequential order,which means each step is executed one by one, because one volume can't be changed by two or more step at the same time.
However, We find a way to execute the plan in parallel,which is not only  much more efficient but also safe.The detail is here:
https://github.com/liumihust/Parallel-Block-Copy-HDFS


elek:

PR #259, opened 2017-08-03
MAPREDUCE-6931. Remove TestDFSIO "Total Throughput" calculation.

dennishuo:
Previously it failed to convert ms to seconds and thus reports aggregate
throughput as 1/1000x actual numbers. Also, make all the bytes-to-mb
and milliseconds-to-seconds conversions consistent in the reporting
messages to help avoid this type of error in the future.

### DIFF
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-common/src/main/resources/yarn-default.xml
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-common/src/main/resources/yarn-default.xml
@@ -1688,7 +1688,7 @@
   <property>
     <description>Property to enable docker user remapping</description>
     <name>yarn.nodemanager.runtime.linux.docker.enable-userremapping.allowed</name>
-    <value>false</value>
+    <value>true</value>
   </property>
 
   <property>


hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/src/test/java/org/apache/hadoop/fs/TestDFSIO.java
```diff
@@ -855,7 +855,7 @@ public int run(String[] args) throws IOException {
       long tStart = System.currentTimeMillis();
       sequentialTest(fs, testType, nrBytes, nrFiles);
       long execTime = System.currentTimeMillis() - tStart;
-      String resultLine = "Seq Test exec time sec: " + (float)execTime / 1000;
+      String resultLine = "Seq Test exec time sec: " + msToSecs(execTime);
```
clehene: Perhaps rename variable to includes units too? E.g. `execTimeMs` ?

dennishuo: Good suggestion, done.

PR #261, opened 2017-08-11
YARN-2162. add ability to optionally configure maxResources in terms …

flyrain:
…of percentage

### DIFF
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationConfiguration.java
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationConfiguration.java
@@ -40,15 +40,13 @@ import com.google.common.annotations.VisibleForTesting;
 public class AllocationConfiguration extends ReservationSchedulerConfiguration {
   private static final AccessControlList EVERYBODY_ACL = new AccessControlList("*");
   private static final AccessControlList NOBODY_ACL = new AccessControlList(" ");
-  private static final ResourceCalculator RESOURCE_CALCULATOR =
-      new DefaultResourceCalculator();
   // Minimum resource allocation for each queue
   private final Map<String, Resource> minQueueResources;
   // Maximum amount of resources per queue
   @VisibleForTesting
-  final Map<String, Resource> maxQueueResources;
+  final Map<String, ConfigurableResource> maxQueueResources;
   // Maximum amount of resources for each queue's ad hoc children
-  private final Map<String, Resource> maxChildQueueResources;
+  private final Map<String, ConfigurableResource> maxChildQueueResources;
   // Sharing weights for each queue
   private final Map<String, Float> queueWeights;
   
@@ -60,7 +58,7 @@ public class AllocationConfiguration extends ReservationSchedulerConfiguration {
   final Map<String, Integer> userMaxApps;
   private final int userMaxAppsDefault;
   private final int queueMaxAppsDefault;
-  private final Resource queueMaxResourcesDefault;
+  private final ConfigurableResource queueMaxResourcesDefault;
 
   // Maximum resource share for each leaf queue that can be used to run AMs
   final Map<String, Float> queueMaxAMShares;
@@ -109,14 +107,14 @@ public class AllocationConfiguration extends ReservationSchedulerConfiguration {
   private final Set<String> nonPreemptableQueues;
 
   public AllocationConfiguration(Map<String, Resource> minQueueResources,
-      Map<String, Resource> maxQueueResources,
-      Map<String, Resource> maxChildQueueResources,
+      Map<String, ConfigurableResource> maxQueueResources,
+      Map<String, ConfigurableResource> maxChildQueueResources,
       Map<String, Integer> queueMaxApps,
       Map<String, Integer> userMaxApps,
       Map<String, Float> queueWeights,
       Map<String, Float> queueMaxAMShares, int userMaxAppsDefault,
       int queueMaxAppsDefault,
-      Resource queueMaxResourcesDefault,
+      ConfigurableResource queueMaxResourcesDefault,
       float queueMaxAMShareDefault,
       Map<String, SchedulingPolicy> schedulingPolicies,
       SchedulingPolicy defaultSchedulingPolicy,
@@ -165,7 +163,7 @@ public class AllocationConfiguration extends ReservationSchedulerConfiguration {
     queueMaxAMShares = new HashMap<>();
     userMaxAppsDefault = Integer.MAX_VALUE;
     queueMaxAppsDefault = Integer.MAX_VALUE;
-    queueMaxResourcesDefault = Resources.unbounded();
+    queueMaxResourcesDefault = new ConfigurableResource(Resources.unbounded());
     queueMaxAMShareDefault = 0.5f;
     queueAcls = new HashMap<>();
     resAcls = new HashMap<>();
@@ -291,26 +289,17 @@ public class AllocationConfiguration extends ReservationSchedulerConfiguration {
 
   /**
    * Get the maximum resource allocation for the given queue. If the max in not
-   * set, return the larger of the min and the default max.
+   * set, return the default max.
    *
    * @param queue the target queue's name
    * @return the max allocation on this queue
    */
-  @VisibleForTesting
-  Resource getMaxResources(String queue) {
-    Resource maxQueueResource = maxQueueResources.get(queue);
+  @VisibleForTesting ConfigurableResource getMaxResources(String queue) {
+    ConfigurableResource maxQueueResource = maxQueueResources.get(queue);
     if (maxQueueResource == null) {
-      Resource minQueueResource = minQueueResources.get(queue);
-      if (minQueueResource != null &&
-          Resources.greaterThan(RESOURCE_CALCULATOR, Resources.unbounded(),
-          minQueueResource, queueMaxResourcesDefault)) {
-        return minQueueResource;
-      } else {
-        return queueMaxResourcesDefault;
-      }
-    } else {
-      return maxQueueResource;
+      maxQueueResource = queueMaxResourcesDefault;
     }
+    return maxQueueResource;
   }
   
   /**
@@ -319,8 +308,7 @@ public class AllocationConfiguration extends ReservationSchedulerConfiguration {
    * @param queue the target queue's name
    * @return the max allocation on this queue or null if not set
    */
-  @VisibleForTesting
-  Resource getMaxChildResources(String queue) {
+  @VisibleForTesting ConfigurableResource getMaxChildResources(String queue) {
     return maxChildQueueResources.get(queue);
   }
 
@@ -418,9 +406,9 @@ public class AllocationConfiguration extends ReservationSchedulerConfiguration {
     queue.setMaxChildQueueResource(getMaxChildResources(name));
 
     // Set queue metrics.
-    queue.getMetrics().setMinShare(getMinResources(name));
-    queue.getMetrics().setMaxShare(getMaxResources(name));
-    queue.getMetrics().setMaxApps(getQueueMaxApps(name));
+    queue.getMetrics().setMinShare(queue.getMinShare());
+    queue.getMetrics().setMaxShare(queue.getMaxShare());
+    queue.getMetrics().setMaxApps(queue.getMaxRunningApps());
     queue.getMetrics().setSchedulingPolicy(getSchedulingPolicy(name).getName());
   }
 }
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationConfiguration.java
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationConfiguration.java
@@ -292,7 +292,8 @@ public class AllocationConfiguration extends ReservationSchedulerConfiguration {
    * @param queue the target queue's name
    * @return the max allocation on this queue
    */
-  @VisibleForTesting ConfigurableResource getMaxResources(String queue) {
+  @VisibleForTesting
+  ConfigurableResource getMaxResources(String queue) {
     ConfigurableResource maxQueueResource = maxQueueResources.get(queue);
     if (maxQueueResource == null) {
       maxQueueResource = queueMaxResourcesDefault;
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationConfiguration.java
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationConfiguration.java
@@ -31,8 +31,6 @@ import org.apache.hadoop.yarn.api.records.Resource;
 import org.apache.hadoop.yarn.security.AccessType;
 import org.apache.hadoop.yarn.server.resourcemanager.reservation.ReservationSchedulerConfiguration;
 import org.apache.hadoop.yarn.server.resourcemanager.scheduler.SchedulerUtils;
-import org.apache.hadoop.yarn.util.resource.DefaultResourceCalculator;
-import org.apache.hadoop.yarn.util.resource.ResourceCalculator;
 import org.apache.hadoop.yarn.util.resource.Resources;
 
 import com.google.common.annotations.VisibleForTesting;
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationFileLoaderService.java
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationFileLoaderService.java
@@ -226,8 +226,8 @@ public class AllocationFileLoaderService extends AbstractService {
     // Create some temporary hashmaps to hold the new allocs, and we only save
     // them in our fields if we have parsed the entire allocs file successfully.
     Map<String, Resource> minQueueResources = new HashMap<>();
-    Map<String, Resource> maxQueueResources = new HashMap<>();
-    Map<String, Resource> maxChildQueueResources = new HashMap<>();
+    Map<String, ConfigurableResource> maxQueueResources = new HashMap<>();
+    Map<String, ConfigurableResource> maxChildQueueResources = new HashMap<>();
     Map<String, Integer> queueMaxApps = new HashMap<>();
     Map<String, Integer> userMaxApps = new HashMap<>();
     Map<String, Float> queueMaxAMShares = new HashMap<>();
@@ -244,7 +244,8 @@ public class AllocationFileLoaderService extends AbstractService {
     Set<String> nonPreemptableQueues = new HashSet<>();
     int userMaxAppsDefault = Integer.MAX_VALUE;
     int queueMaxAppsDefault = Integer.MAX_VALUE;
-    Resource queueMaxResourcesDefault = Resources.unbounded();
+    ConfigurableResource queueMaxResourcesDefault =
+        new ConfigurableResource(Resources.unbounded());
     float queueMaxAMShareDefault = 0.5f;
     long defaultFairSharePreemptionTimeout = Long.MAX_VALUE;
     long defaultMinSharePreemptionTimeout = Long.MAX_VALUE;
@@ -305,7 +306,7 @@ public class AllocationFileLoaderService extends AbstractService {
           }
         } else if ("queueMaxResourcesDefault".equals(element.getTagName())) {
           String text = ((Text)element.getFirstChild()).getData().trim();
-          Resource val =
+          ConfigurableResource val =
               FairSchedulerConfiguration.parseResourceConfigValue(text);
           queueMaxResourcesDefault = val;
         } else if ("userMaxAppsDefault".equals(element.getTagName())) {
@@ -448,8 +449,8 @@ public class AllocationFileLoaderService extends AbstractService {
    */
   private void loadQueue(String parentName, Element element,
       Map<String, Resource> minQueueResources,
-      Map<String, Resource> maxQueueResources,
-      Map<String, Resource> maxChildQueueResources,
+      Map<String, ConfigurableResource> maxQueueResources,
+      Map<String, ConfigurableResource> maxChildQueueResources,
       Map<String, Integer> queueMaxApps,
       Map<String, Integer> userMaxApps,
       Map<String, Float> queueMaxAMShares,
@@ -496,17 +497,17 @@ public class AllocationFileLoaderService extends AbstractService {
       Element field = (Element) fieldNode;
       if ("minResources".equals(field.getTagName())) {
         String text = ((Text)field.getFirstChild()).getData().trim();
-        Resource val =
+        ConfigurableResource val =
             FairSchedulerConfiguration.parseResourceConfigValue(text);
-        minQueueResources.put(queueName, val);
+        minQueueResources.put(queueName, val.getResource());
       } else if ("maxResources".equals(field.getTagName())) {
         String text = ((Text)field.getFirstChild()).getData().trim();
-        Resource val =
+        ConfigurableResource val =
             FairSchedulerConfiguration.parseResourceConfigValue(text);
         maxQueueResources.put(queueName, val);
       } else if ("maxChildResources".equals(field.getTagName())) {
         String text = ((Text)field.getFirstChild()).getData().trim();
-        Resource val =
+        ConfigurableResource val =
             FairSchedulerConfiguration.parseResourceConfigValue(text);
         maxChildQueueResources.put(queueName, val);
       } else if ("maxRunningApps".equals(field.getTagName())) {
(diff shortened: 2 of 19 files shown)


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationFileLoaderService.java
```diff
@@ -605,13 +606,15 @@ private void loadQueue(String parentName, Element element,
 
     queueAcls.put(queueName, acls);
     resAcls.put(queueName, racls);
-    if (maxQueueResources.containsKey(queueName) &&
-        minQueueResources.containsKey(queueName)
-        && !Resources.fitsIn(minQueueResources.get(queueName),
-            maxQueueResources.get(queueName))) {
+    ResourceConfiguration maxResourceConf = maxQueueResources.get(queueName);
+    if (maxResourceConf != null &&
+        !maxResourceConf.isPercentage() &&
```
kambatla: A warning is necessary even when the config is percentage-based. I would argue the warning is more important in this case than the other.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
```diff
@@ -343,10 +343,10 @@ public void updateDemand() {
       readLock.unlock();
     }
     // Cap demand to maxShare to limit allocation to maxShare
-    demand = Resources.componentwiseMin(tmpDemand, maxShare);
+    demand = Resources.componentwiseMin(tmpDemand, getMaxShare());
```
kambatla: Nice. Have we verified we don't access maxShare directly anywhere else? How about other classes that have similar field?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSQueue.java
```diff
@@ -158,26 +158,33 @@ public Resource getMinShare() {
...
     this.maxShare = maxShare;
   }
 
+  @Override
+  public Resource getMaxShare() {
+    Resource maxResource = maxShare.getResource(scheduler.getClusterResource());
+    // Set max resource to min resource if min resource is greater than max
+    // resource
+    if(Resources.greaterThan(scheduler.getResourceCalculator(),
+        scheduler.getClusterResource(), minShare, maxResource)) {
```
kambatla: Is the indentation off by one space?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestAllocationFileLoaderService.java
```diff
@@ -235,21 +235,21 @@ public void testAllocationFileParsing() throws Exception {
...
 
     assertEquals(Resources.createResource(2048, 10),
-        queueConf.getMaxResources("root.queueA"));
+        queueConf.getMaxResources("root.queueA").getResource());
     assertEquals(Resources.createResource(5120, 110),
-        queueConf.getMaxResources("root.queueB"));
-    assertEquals(Resources.createResource(5120, 0),
-        queueConf.getMaxResources("root.queueC"));
+        queueConf.getMaxResources("root.queueB").getResource());
+    assertEquals(Resources.createResource(4096, 100),
```
kambatla: Why are we having to change this value here? Shouldn't the behavior be the same?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSParentQueue.java
```diff
@@ -42,6 +43,8 @@ public void setUp() throws Exception {
     AllocationConfiguration allocConf = new AllocationConfiguration(conf);
     when(scheduler.getAllocationConfiguration()).thenReturn(allocConf);
     when(scheduler.getConf()).thenReturn(conf);
+    when(scheduler.getResourceCalculator()).thenReturn(
```
kambatla: Why do we need to set the resource-calculator now and not before?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-site/src/site/markdown/FairScheduler.md
```diff
@@ -85,9 +85,9 @@ The allocation file must be in XML format. The format contains five types of ele
...
 
-    * **maxResources**: maximum resources a queue is allowed, in the form "X mb, Y vcores". A queue will never be assigned a container that would put its aggregate usage over this limit.
+    * **maxResources**: maximum resources a queue is allowed, in the form "X mb, Y vcores". A queue will never be assigned a container that would put its aggregate usage over this limit. Set it to a percentage of all cluster resources in form of "50%" or "50% memory, 60% vcore".
```
kambatla: How about:

**maxResources**: maximum resources a queue is allocated, expressed either in absolute values (X mb, Y vcores) or as a percentage of the cluster resources (X% memory, Y% cpu). A queue will not be assigned a container that would put its aggregate usage over this limit.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-site/src/site/markdown/FairScheduler.md
```diff
@@ -85,9 +85,9 @@ The allocation file must be in XML format. The format contains five types of ele
...
 
-    * **maxChildResources**: maximum resources an ad hoc child queue is allowed, in the form "X mb, Y vcores". Any ad hoc queue that is a direct child of a queue with this property set will have it's maxResources property set accordingly.
+    * **maxChildResources**: maximum resources an ad hoc child queue is allowed, in the form "X mb, Y vcores". Any ad hoc queue that is a direct child of a queue with this property set will have it's maxResources property set accordingly. Set it to a percentage of all cluster resources in form of "50%" or "50% memory, 60% vcore".
```
kambatla: Similar to the above config?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairSchedulerConfiguration.java
```diff
@@ -287,27 +289,69 @@ public float getReservableNodes() {
...
+    HashMap<ResourceType, Double> resourcePercentage = new HashMap<>();
+    String[] strings = val.split(",");
+    if (strings.length == 1) {
+      double percentage = findPercentage(strings[0], "");
+      for (ResourceType resourceType: ResourceType.values()) {
+        resourcePercentage.put(resourceType, percentage/100);
+      }
+    } else {
+      double memPercentage = findPercentage(val, "memory");
+      double vcorePercentage = findPercentage(val, "vcore");
```
kambatla: vcore is a unit of cpu. If we are using memory instead of mb, we should use cpu instead of vcore.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationConfiguration.java
```diff
@@ -321,8 +310,7 @@ Resource getMaxResources(String queue) {
    * @param queue the target queue's name
    * @return the max allocation on this queue or null if not set
    */
-  @VisibleForTesting
-  Resource getMaxChildResources(String queue) {
+  @VisibleForTesting ResourceConfiguration getMaxChildResources(String queue) {
```
flyrain: Will do.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSLeafQueue.java
flyrain: Yes. It is verified. Besides, FSQueue#maxShare is private now.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSQueue.java
flyrain: Didn't see the formatting issue, but will reformat in next patch anyway to make sure it is right.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationConfiguration.java
```diff
@@ -293,26 +293,15 @@ Resource getMinResources(String queue) {
 
   /**
    * Get the maximum resource allocation for the given queue. If the max in not
-   * set, return the larger of the min and the default max.
+   * set, return the default max.
```
flyrain: I changed it since this is not the right place to do that. I move this behavior to FSQueue#getMaxShare. The reason is allocation file reload and cluster resource changing are two independent events. It is not right to calculate based on cluster resource while refresh the allocation file.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationFileLoaderService.java
```diff
@@ -495,17 +496,17 @@ private void loadQueue(String parentName, Element element,
       Element field = (Element) fieldNode;
       if ("minResources".equals(field.getTagName())) {
         String text = ((Text)field.getFirstChild()).getData().trim();
-        Resource val =
+        ResourceConfiguration val =
```
flyrain: The major usage of min share is sorting schedulables and preemption. I didn't see many cons. In the other side,
the needs for min share as percentage are limited. It shouldn't a common case that a job/queue says to RM that no matter how much resource you have, I need at least 10% of them. Make min share as percentage isn't necessary as well. Hence, we could just leave it alone.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationFileLoaderService.java
flyrain: Add the warning in method FSQueue#getMaxShare()

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairSchedulerConfiguration.java
flyrain: Will do in next patch.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/ResourceConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.HashMap;
+
+/**
+ * A ResourceConfiguration represents an entity that is used to configuration
+ * resources, such as maximum resources of a queue. It can be percentage of
+ * cluster resources or an absolute value.
+ */
+@Private
+@Unstable
+public class ResourceConfiguration {
```
flyrain: Will change to ConfigurableResource in next patch.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/ResourceConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+  private final boolean isPercentage;
```
flyrain: I prefer a boolean. We need to check the state of the object sometimes. Without a boolean, we are gonna to rely on value of absolute resource and/or percentage, which is not a nice way, e.g., what if absolute resource is set to null.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/ResourceConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+  private final Resource resource;
+  private final HashMap<ResourceType, Double> percentages;
```
flyrain: Will change to an array in next patch.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/ResourceConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public boolean isPercentage() {
```
flyrain: We need to check the state of the object sometimes. Without a boolean, we are gonna to rely on value of absolute resource and/or percentage, which is not a nice way, e.g., what if absolute resource is set to null.

flyrain: Remove it in next patch.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/ResourceConfiguration.java
flyrain: Remove the boolean in next patch

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestAllocationFileLoaderService.java
flyrain: The logic to set max share to min share if max share is less than min share has been moved to FSQueue#getMaxShare

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/TestFSParentQueue.java
flyrain: It throws an exception like this if we don't do this:
```
java.lang.NullPointerException
	at org.apache.hadoop.yarn.util.resource.Resources.greaterThan(Resources.java:310)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.FSQueue.getMaxShare(FSQueue.java:171)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.AllocationConfiguration.initFSQueue(AllocationConfiguration.java:409)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.FSQueue.reinit(FSQueue.java:111)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.FSQueue.<init>(FSQueue.java:96)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.FSParentQueue.<init>(FSParentQueue.java:60)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.QueueManager.initialize(QueueManager.java:77)
	at org.apache.hadoop.yarn.server.resourcemanager.scheduler.fair.TestFSParentQueue.setUp(TestFSParentQueue.java:58)
```
This code path doesn't exist before my patch.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/ResourceConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+    this.resource = null;
+  }
+
+  public ResourceConfiguration(Resource resource) {
+    isPercentage = false;
+    this.percentages = null;
+    this.resource = resource;
+  }
+
+  public Resource getResource(Resource clusterResource) {
```
flyrain: I evaluated two ways you suggested. Seems neither of them is better than the other. I rewrite the getResource a little bit. Let me know what you think. Thanks.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/ConfigurableResource.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public ConfigurableResource(Resource resource) {
+    this.percentages = null;
+    this.resource = resource;
+  }
+
+  public Resource getResource(Resource clusterResource) {
+    if (percentages != null && clusterResource != null) {
+      long memory = (long) (clusterResource.getMemorySize() * percentages[0]);
```
kambatla: Is this inline with the ResourceTypes work? Do we want someone from that world, say Daniel, to look at this patch?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSQueue.java
```diff
@@ -158,26 +158,37 @@ public Resource getMinShare() {
...
   }
 
+  @Override
+  public Resource getMaxShare() {
+    Resource maxResource = maxShare.getResource(scheduler.getClusterResource());
+
+    // Set max resource to min resource if min resource is greater than max
+    // resource
+    if(Resources.greaterThan(scheduler.getResourceCalculator(),
+        scheduler.getClusterResource(), minShare, maxResource)) {
```
kambatla: This should likely by componentWiseMax. See Daniel's patch on YARN-6964.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairSchedulerConfiguration.java
```diff
@@ -287,27 +289,69 @@ public float getReservableNodes() {
...
+    String[] strings = val.split(",");
+    if (strings.length == 1) {
+      double percentage = findPercentage(strings[0], "");
+      for (int i = 0 ; i < ResourceType.values().length ; i++) {
+        resourcePercentage[i] = percentage/100;
+      }
+    } else {
+      double memPercentage = findPercentage(val, "memory");
+      double vcorePercentage = findPercentage(val, "cpu");
+      resourcePercentage[0] = memPercentage/100;
```
kambatla: With ResourceTypes coming in, I feel this has to be handled better.  Loop over and use that index instead?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/AllocationConfiguration.java
```diff
@@ -290,26 +290,15 @@ Resource getMinResources(String queue) {
...
-      } else {
-        return queueMaxResourcesDefault;
-      }
-    } else {
-      return maxQueueResource;
-    }
+  @VisibleForTesting ConfigurableResource getMaxResources(String queue) {
+    ConfigurableResource maxQueueResource = maxQueueResources.get(queue);
+    return maxQueueResource == null ?
+        queueMaxResourcesDefault : maxQueueResource;
```
templedf: I'm not a fan of the ternary operator unless it really makes things clearer.  I don't see the point here.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/ConfigurableResource.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.apache.hadoop.classification.InterfaceAudience.Private;
+import org.apache.hadoop.classification.InterfaceStability.Unstable;
+import org.apache.hadoop.yarn.api.records.Resource;
+import org.apache.hadoop.yarn.server.resourcemanager.resource.ResourceType;
+
+import java.util.HashMap;
+
+/**
+ * A {@link ConfigurableResource} object represents an entity that is used to
```
templedf: No point linking to yourself.  Just make it a {@code} instead.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/ConfigurableResource.java
```diff
@@ -0,0 +1,62 @@
...
+      long memory = (long) (clusterResource.getMemorySize() * percentages[0]);
+      int vcore = (int) (clusterResource.getVirtualCores() * percentages[1]);
+      return Resource.newInstance(memory, vcore);
+    } else {
+      return resource;
+    }
+  }
+
+  public Resource getResource() {
+    return getResource(null);
```
templedf: This also seems kinda pointless and a bit brittle.  Just return resource.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FSQueue.java
templedf: Yeah, you should avoid greaterThan() (et al) if you can.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/fair/FairSchedulerConfiguration.java
templedf: Agreed.  This seems needlessly convoluted.

flyrain:
Committed

PR #266, opened 2017-08-17
HDFS-12315: Use Path instead of String to check closedFiles set

dosoft:


elek:

PR #272, opened 2017-09-12
HDFS-12424 Fix sorting of the datatable on the Datanodes page in the NN UI

shawnam:
This fix basically comes down to three things:

- type: numeric isn't a thing -- it needs to be type num per the [datatable doc](https://datatables.net/reference/option/columns.type)
- orderData was wrong in both instances; we don't want those columns to sort by some other column
- version sorting by anything other than string is going to be even more confusing than it is if we sort it by string


shawnam:
I'm going to go the patch route.

PR #282, opened 2017-10-20
HADOOP-14971 Merge S3A committers into trunk

steveloughran:
HADOOP 13786 & MAPREDUCE-6823 code as a PR for better review


hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/DirectoryStagingCommitter.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+
+/**
+ * This commits to a directory.
+ * The conflict policy is
+ * <ul>
+ *   <li>FAIL: fail the commit</li>
+ *   <li>APPEND: add extra data to the destination.</li>
+ *   <li>REPLACE: delete the destination directory.</li>
```
rdblue: I think it's worth mentioning that this delete happens if and only if all tasks completed successfully. This is a fairly safe operation.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/Paths.java
```diff
@@ -0,0 +1,306 @@
...
+   * @param childPath full path under the base path.
+   * @return the relative path
+   */
+  public static String getRelativePath(Path basePath,
+      Path childPath) {
+    //
+    // Use URI.create(Path#toString) to avoid URI character escape bugs
+    URI relative = URI.create(basePath.toString())
+        .relativize(URI.create(childPath.toString()));
+    return relative.getPath();
```
rdblue: We updated this because we found that there can be character escape bugs. Here's the new implementation:
```
    return basePath.toUri().relativize(fullPath.toUri()).getPath();
```

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/Paths.java
```diff
@@ -0,0 +1,306 @@
...
+   */
+  static Path getMultipartUploadCommitsDirectory(FileSystem fs,
+      Configuration conf, String uuid) throws IOException {
+    return path(tempDirForStaging(fs, conf),
+        UserGroupInformation.getCurrentUser().getShortUserName(),
+        uuid,
+        STAGING_UPLOADS);
+  }
+
+  // TODO: verify this is correct, it comes from dse-storage
```
rdblue: Probably don't need this comment.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/Paths.java
```diff
@@ -0,0 +1,306 @@
...
+  private static Path localTemp(Configuration conf, int taskId, int attemptId)
+      throws IOException {
+    String[] dirs = conf.getTrimmedStrings(BUFFER_DIR);
+    Random rand = new Random(Objects.hashCode(taskId, attemptId));
+    String dir = dirs[rand.nextInt(dirs.length)];
+
+    return FileSystem.getLocal(conf).makeQualified(new Path(dir));
```
rdblue: We ended up changing this to balance across multiple local directories using `LocalDirAllocator`. That requires keeping track of the directory used for each task attempt in a static hashmap to ensure you don't lose track of files. Here's what we use now:

```java
  @VisibleForTesting
  public static Cache<TaskAttemptID, Path> tempFolders = CacheBuilder
      .newBuilder().build();

  public static Path getLocalTaskAttemptTempDir(final Configuration conf,
                                                final String uuid,
                                                final TaskAttemptID attemptID)
      throws IOException {
    try {
      return tempFolders.get(attemptID, new Callable<Path>() {
        @Override
        public Path call() throws Exception {
          LocalDirAllocator allocator = new LocalDirAllocator(
              "mapreduce.cluster.local.dir");
          return FileSystem.getLocal(conf).makeQualified(
              allocator.getLocalPathForWrite(uuid, conf));
        }
      });
    } catch (ExecutionException e) {
      throw new RuntimeException(e.getCause());
    } catch (UncheckedExecutionException e) {
      if (e.getCause() instanceof RuntimeException) {
        throw (RuntimeException) e.getCause();
      }
      throw new RuntimeException(e);
    }
  }
```

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/Paths.java
```diff
@@ -0,0 +1,306 @@
...
+      partitions.add(partition != null ? partition : TABLE_ROOT);
+    }
+
+    return partitions;
+  }
+
+  /**
+   * path filter.
+   */
+  static final class HiddenPathFilter implements PathFilter {
```
rdblue: Seems like we write this class over and over. Should we move it into the public API instead of having a local one here?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
```diff
@@ -0,0 +1,908 @@
...
+      TaskAttemptContext context) throws IOException {
+    super(outputPath, context);
+    this.constructorOutputPath = checkNotNull(getOutputPath(), "output path");
+    Configuration conf = getConf();
+    this.uploadPartSize = conf.getLongBytes(
+        MULTIPART_SIZE, DEFAULT_MULTIPART_SIZE);
+
+    // Spark will use a fake app ID based on the current minute and job ID 0.
+    // To avoid collisions, use the YARN application ID for Spark.
+    this.uuid = getUploadUUID(conf, context.getJobID());
```
rdblue: In some cases, Spark will create a UUID for each write and store it in `spark.sql.sources.writeJobUUID`. I'd recommend grabbing that if possible instead of using the application ID.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
```diff
@@ -0,0 +1,908 @@
...
+  public StagingCommitter(Path outputPath,
+      TaskAttemptContext context) throws IOException {
+    super(outputPath, context);
+    this.constructorOutputPath = checkNotNull(getOutputPath(), "output path");
+    Configuration conf = getConf();
+    this.uploadPartSize = conf.getLongBytes(
+        MULTIPART_SIZE, DEFAULT_MULTIPART_SIZE);
+
+    // Spark will use a fake app ID based on the current minute and job ID 0.
+    // To avoid collisions, use the YARN application ID for Spark.
```
rdblue: Should this comment be in `getUploadUUID`?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
```diff
@@ -0,0 +1,908 @@
...
+    this.uuid = getUploadUUID(conf, context.getJobID());
+    this.uniqueFilenames = conf.getBoolean(
+        FS_S3A_COMMITTER_STAGING_UNIQUE_FILENAMES,
+        DEFAULT_STAGING_COMMITTER_UNIQUE_FILENAMES);
+    setWorkPath(buildWorkPath(context, uuid));
+    this.wrappedCommitter = createWrappedCommitter(context, conf);
+    // forces evaluation and caching of the resolution mode.
+    ConflictResolution mode = getConflictResolutionMode(getJobContext());
+    LOG.debug("Conflict resolution mode: {}", mode);
+    this.finalOutputPath = constructorOutputPath;
```
rdblue: Not being able to change the output directory in a subclass makes it so we can't use this committer as the base for our batch committers. That's not a big deal for us, since we will probably continue using our own for a while. But, it would be nice to be able to change the output location based on table metadata. Our directory committer, for example, will create a new directory for data instead of deleting the current path, then it will switch the table location to the new directory. That can't be done unless there is a way to hook this output path.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
```diff
@@ -0,0 +1,908 @@
...
+      return null;
+    }
+  }
+
+  /**
+   * The staging committers do not require "magic" commit support.
+   * @return false
+   */
+  @Override
+  protected boolean isMagicFileSystemRequired() {
```
rdblue: Is it possible to get rid of this? I never remember what "magic" refers to. I'd recommend either a better name or removing it from the API.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
```diff
@@ -0,0 +1,908 @@
...
+    } else {
+      boolean threw = true;
+      // before the uploads, report some progress
+      context.progress();
+
+      PendingSet pendingCommits = new PendingSet(commitCount);
+      try {
+        Tasks.foreach(taskOutput)
+            .stopOnFailure()
+            .throwFailureWhenFinished()
```
rdblue: We recently caught a bug with retry where I accidentally forgot to add `throwFailureWhenFinished`. I think it would be be a better API to throw unless there's a call to suppress, like `suppressFailures`. I haven't made this update in our copy either, but you might consider for maintainability it before this goes in.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
```diff
@@ -0,0 +1,908 @@
...
+      // this, so warn that there is some kind of problem in the protocol.
+      LOG.warn("{}: No files to commit", getRole());
+    } else {
+      boolean threw = true;
+      // before the uploads, report some progress
+      context.progress();
+
+      PendingSet pendingCommits = new PendingSet(commitCount);
+      try {
+        Tasks.foreach(taskOutput)
```
rdblue: We added retry to `Tasks` recently, which helps reliability. Here's our implementation, which is based on the exponential back-off code from Presto:

```java
    private <E extends Exception> void runTaskWithRetry(Task<I, E> task, I item)
        throws E {
      long start = System.currentTimeMillis();
      int attempt = 0;
      while (true) {
        attempt += 1;
        try {
          task.run(item);
          break;

        } catch (Exception e) {
          LOG.warn("Retrying task after failure: " + e.getMessage());

          long durationMs = System.currentTimeMillis() - start;
          if (attempt >= maxAttempts || durationMs > maxDurationMs) {
            throw e;
          }

          for (Class<? extends Exception> exClass : stopRetryExceptions) {
            if (exClass.isInstance(e)) {
              throw e;
            }
          }

          int delayMs = (int) Math.min(
              minSleepTimeMs * Math.pow(scaleFactor, attempt - 1),
              maxSleepTimeMs);
          int jitter = ThreadLocalRandom.current()
              .nextInt(Math.max(1, (int) (delayMs * 0.1)));

          try {
            TimeUnit.MILLISECONDS.sleep(delayMs + jitter);
          } catch (InterruptedException ie) {
            Thread.currentThread().interrupt();
            throw new RuntimeException(ie);
          }
        }
      }
    }
```

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
```diff
@@ -0,0 +1,908 @@
...
+            .throwFailureWhenFinished()
+            .executeWith(buildThreadPool(context))
+            .run(stat -> {
+              Path path = stat.getPath();
+              File localFile = new File(path.toUri().getPath());
+              String relative = Paths.getRelativePath(attemptPath, path);
+              String partition = Paths.getPartition(relative);
+              String key = getFinalKey(relative, context);
+              Path destPath = getDestS3AFS().keyToQualifiedPath(key);
+              SinglePendingCommit commit = getCommitOperations()
```
rdblue: Looks like job commit logic has been consolidated based on this generic `SinglePendingCommit` class, right? Where does the job commit logic live now?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/PartitonedStagingCommitterFactory.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.s3a.S3AFileSystem;
+import org.apache.hadoop.fs.s3a.commit.AbstractS3GuardCommitterFactory;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.output.PathOutputCommitter;
+
+/**
+ * Factory for the {@link PartitionedStagingCommitter}.
+ */
+public class PartitonedStagingCommitterFactory
```
steveloughran: yes. I have spelt this wrong. Will fix on the next update

hadoop-common-project/hadoop-common/src/main/resources/core-default.xml
```diff
@@ -1344,34 +1338,34 @@
...
-  <name>fs.s3a.retry.limit</name>
-  <value>4</value>
-  <description>
-    Number of times to retry any repeatable S3 client request on failure,
-    excluding throttling requests.
-  </description>
+  <name>fs.s3a.attempts.maximum</name>
+  <value>20</value>
+  <description>How many times we should retry commands on transient errors,
+  excluding throttling errors.</description>
```
ajfabbri: Interesting.  One of my concerns about all the retry logic being added here is that it is an invasive change and I'm feeling like there might be unintended consequences somewhere.  I've been thinking that making it more configurable would mitigate the risk.. I'd lean towards making more types/classes of retry configuration instead of fewer.   For example here, I'd like to have SDK retries configured separately.  I mentioned before also the idea of having another retry policy for riskier parts (e.g. delete).  Thoughts?

steveloughran: I worry about overconfig of things which aren't easy to test. At the same time, the bulk/recursive ops are special cases because we need to think "how best to retry?". Is it per-op, or would you want to step back and start from scratch. With the fact that delete retries aren't isolated from other ops in the bucket, that's one where we could make a case for saying "no, no retries except in higher level algorithms (like the commit one)"

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/DirectoryStagingCommitter.java
steveloughran: done, added docs to partitioned committer and fixed error in Partitioned commitTaskInternal() which stated the delete was in task commit.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/PartitionedStagingCommitter.java
```diff
@@ -0,0 +1,134 @@
...
+      // FAIL checking is done on the task side, so this does nothing
+      break;
+    case APPEND:
+      // no check is needed because the output may exist for appending
+      break;
+    case REPLACE:
+      for (Path partitionPath : partitions) {
+        LOG.info("{}: removing partition path to be replaced: " +
+            getRole(), partitionPath);
+        fs.delete(partitionPath, true);
```
steveloughran: Ryan: what do you think we should for `TABLE_ROOT` part here? nonrecursive list of root entries & delete?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/Paths.java
steveloughran: where was fullPath created?

steveloughran: fixed

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/Paths.java
steveloughran: OK, I'll do that. 
1. Any tests?
2. Task committer will invalidate the cache value for its task ID at end of taskCommit

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/Paths.java
steveloughran: its ubiquitous, isn't it? And in java 8, we can reimplement as a closure...

steveloughran: OK, moved up to S3AUtils; combined with other filters in s3a (Listing) and the one in S3ATestUtils.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
steveloughran: done, added full text on what goes on, and make sure that conf.getTrimmed() is used consistently there

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
steveloughran: Ok, I'll make it non-final :)

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
steveloughran: I was trying really hard to think up a good name here. I will change it in this committer but use it elsewhere as "supportsPendingWritesToSpecialPaths" is just too nonintuitive

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
steveloughran: Update: next revision will 
1. eliminate `finalOutputPath`as a field
1. call `setOutputPath(finalOutputPath)` in constructor
1. then call `getOutputPath()` and use it in the constructor to set up the destFS, which is always fetched with `getDestFS()`
1. And all uses of `getOutputPath(JobContext)` are gone (in job setup and commit); expect the committer to be set up with the correct output path.

This should mean that whatever sets output path will define that output path,; it should be overrideable, provided that the constructor-invoked methods don't cause problems before a subclass is inited. Will that work?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
steveloughran: done

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
steveloughran: I am going to steer clear of it for now, as there is retry logic deeper in the commit options and FS itself which tries to disguishable between retryable and unrecoverable failures (ConnectionRefused vs UnknownHost, ...); it's happening further down. But I could life this code (and whatever setsup the stopRetryException list if you think it should all go together

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
steveloughran: All moved to CommitOperations and WriteOperationsHelper: the latter bridges into the AWS S3 client calls kept all hidden behind retry & instrumentation; the former doing the load & apply of the pending files, which are now serialized in JSON in two forms, "single commit" and multiple commits. the staging committer create a single multiple commit file per task (which contains the single commits); the magic committer creates one single commit file per written file, saved somewhere under  __magic/; these files are consolidated in task commit to single files, just to reduce load times on job commit

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/InconsistentAmazonS3Client.java
```diff
@@ -171,14 +209,15 @@ public static InconsistentAmazonS3Client castFrom(AmazonS3 c) throws
...
+  public DeleteObjectsResult deleteObjects(
+      DeleteObjectsRequest deleteObjectsRequest)
       throws AmazonClientException, AmazonServiceException {
     for (DeleteObjectsRequest.KeyVersion keyVersion :
         deleteObjectsRequest.getKeys()) {
       registerDeleteObject(keyVersion.getKey(), deleteObjectsRequest
-          .getBucketName());
+        .getBucketName());
     }
+    maybeFail();
```
ajfabbri: I'd move the failure point above the for loop.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/InconsistentAmazonS3Client.java
```diff
@@ -426,8 +490,10 @@ private void registerDeleteObject(String key, String bucket) {
     if (shouldDelay(key)) {
       // Record summary so we can add it back for some time post-deletion
       S3ObjectSummary summary = null;
-      ObjectListing list = listObjects(bucket, key);
-      for (S3ObjectSummary result : list.getObjectSummaries()) {
+      ListObjectsV2Request request =
+          new ListObjectsV2Request().withBucketName(bucket)
+              .withPrefix(key);
```
ajfabbri: Any reason for the change here? Do we need to skip all the tests that depend on this class when v1 API is configured?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/InconsistentAmazonS3Client.java
steveloughran: done

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/InconsistentAmazonS3Client.java
steveloughran: No real reason. I do recall how an aggressive throttle rate had real problems with listing because a single op could fail too many times for the retry handlers to cope. But I don't see that this has fixed it because its not calling the innerList() op which is needed to bypass that inconsistency. 
I'll revert to v2 and innerListV1, so that there's no failures in this loop, and, assuming all stores support v1 alongside v2, work everywher

steveloughran: moved back to v1. IDE also suggested fancy java 8 findfirst() streaming op, which I also switched to. Once you can read it, it's better

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
rdblue: Up to you. This is more important to us since we use the S3 client directly. If you're using a mechanism in S3A, then you can probably add retries elsewhere. I would recommend having them, though. We see throttling from S3, which shouldn't take down a job.

The `stopRetryException` list is per execution and is seeded with a `NotRetryableException` that can be thrown in the body of run. That way you can identify certain exceptions (like 404 errors in S3) that won't be retried.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -1028,60 +1183,98 @@ public S3AStorageStatistics getStorageStatistics() {
...
+    validateListArguments(request);
+    return invoker.retryUntranslated(
+        request.toString(),
+        true,
+        () -> {
+          if (useListV1) {
+            Preconditions.checkArgument(request.isV1());
+            return S3ListResult.v1(s3.listObjects(request.getV1()));
+          } else {
+            Preconditions.checkArgument(!request.isV1());
```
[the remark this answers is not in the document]
steveloughran: done

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -1130,17 +1359,23 @@ private void blockRootDelete(String key) throws InvalidRequestException {
...
    */
+  @Retries.RetryRaw
   private void deleteObjects(DeleteObjectsRequest deleteRequest)
-      throws MultiObjectDeleteException, AmazonClientException {
+      throws MultiObjectDeleteException, AmazonClientException, IOException {
     incrementWriteOperations();
-    incrementStatistic(OBJECT_DELETE_REQUESTS, 1);
     try {
-      s3.deleteObjects(deleteRequest);
+      invoker.retryUntranslated("delete",
```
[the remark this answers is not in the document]
steveloughran: OK. How about: no retry/untranslated in innerDelete, -> retryTranslated in API delete, with anything the committer doing wrapping it in its own retries and/or ignoring of failures (which of course, is  a valid option in some cases)

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -1269,6 +1500,8 @@ PutObjectResult putObjectDirect(PutObjectRequest putObjectRequest)
     try {
       PutObjectResult result = s3.putObject(putObjectRequest);
       incrementPutCompletedStatistics(true, len);
+      // update metadata
+      finishedWrite(putObjectRequest.getKey(), len);
```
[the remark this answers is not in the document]
steveloughran: oh we did, it just got moved around. It's what updates s3guard. Critical it's there. What has probably just changed is where it gets invoked from. Before it was part of innerCopyFromLocalFile(), but that's had the inner put pulled out into executePut(), which is how the magic committers can PUT its summary data up directly into the __magic dirs

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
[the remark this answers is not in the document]
ajfabbri: not quite following your comment here.  One question on my mind is the behavior of recursive delete.  Should it fail if the set of objects changes underneath it during execution?  For example, I'm executing a recursive delete, and an object disappears due to concurrent modification or eventual consistency.  Should the delete fail, or succeed because the contract of delete(path) was satisfied (stuff was deleted).  Based on my reading of your RetryPolicy, you don't retry on object not found (good), but we still fail.  We may want to actually throw to top-level delete() and restart the recursive delete.  A bit orthogonal but I want to understand the big picture to feel confident this code is right, and where it is headed in the future.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/s3guard/DynamoDBMetadataStore.java
```diff
@@ -306,12 +338,15 @@ public void initialize(Configuration config) throws IOException {
...
-  private void setMaxRetries(Configuration config) {
+  private void initDataAccessRetries(Configuration config) {
     int maxRetries = config.getInt(S3GUARD_DDB_MAX_RETRIES,
         S3GUARD_DDB_MAX_RETRIES_DEFAULT);
     dataAccessRetryPolicy = RetryPolicies
         .exponentialBackoffRetry(maxRetries, MIN_RETRY_SLEEP_MSEC,
             TimeUnit.MILLISECONDS);
+    dataAccess = new Invoker(dataAccessRetryPolicy,
+        this::retryEvent
+    );
```
ajfabbri: Doesn't look like this is ever used.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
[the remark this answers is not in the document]
ajfabbri: yep confirmed it looks like we are still calling this exactly once. thanks.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Invoker.java
```diff
@@ -0,0 +1,446 @@
...
+    Exception caught;
+    RetryPolicy.RetryAction retryAction;
+    boolean shouldRetry;
+    do {
+      try {
+        // execute the operation, returning if successful
+        return operation.execute();
+      } catch (IOException | SdkBaseException e) {
+        caught = e;
+      }
```
ajfabbri: Commented on this section before.. may have had it wrong then, but to be clear: you want to propagate things like NullPointerException thrown from operation.execute()?  I suppose unexpected exceptions are always non-retryable but I'm surprised we don't want to make it an IOE here.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Listing.java
```diff
@@ -540,10 +541,11 @@ public int getBatchSize() {
      * initial set of results/fail if there was a problem talking to the bucket.
      * @param listPath path of the listing
      * @param request initial request to make
-     * */
+     * @throws IOException if listObjects raises one.
+     */
     ObjectListingIterator(
         Path listPath,
-        S3ListRequest request) {
+        S3ListRequest request) throws IOException {
```
ajfabbri: yeah good catch.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -2286,6 +2598,7 @@ private SSECustomerKey generateSSECustomerKey() {
    * @param length  total length of file written
    */
   @InterfaceAudience.Private
+  @Retries.RetryRaw("Exceptions are swallowed")
```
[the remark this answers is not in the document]
steveloughran: yes, it is a form of translation, isn't it: ex -> void

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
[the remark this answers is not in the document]
steveloughran: Fundamentally, the core contract of a filesystem directory delete: it changes from the state "is" to the state "isnt", in an O(1) atomic op, is not satisified, we just pretend it does..

There's already ambiguity in our pretence. If an object disappears during a delete, that is not detected. If a new object is added if it is in a large dir and a later list() call finds it, it is deleted. Otherwise, it remains. Which means that we don't even detect updates to a path, not unless after a delete() you do another list to see if there has been a change (or do some leasing in dynamo, but let's not go there).

AFAIK, delete() succeeds if any child is taken away; the outcome on an add is undefined. rename is a bit more ambiguous w.r.t either.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java
```diff
@@ -1028,60 +1183,98 @@ public S3AStorageStatistics getStorageStatistics() {
...
+  protected S3ListResult listObjects(S3ListRequest request) throws IOException {
     incrementReadOperations();
+    incrementStatistic(OBJECT_LIST_REQUESTS);
+    validateListArguments(request);
+    return invoker.retryUntranslated(
+        request.toString(),
+        true,
+        () -> {
+          if (useListV1) {
+            Preconditions.checkArgument(request.isV1());
```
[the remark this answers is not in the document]
steveloughran: done

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/Invoker.java
steveloughran: I'm taking the view that anything which isn't some network IOE or from the AWS SDK is not something to go near. Invalid args, NPE, assertions Interruption, OOM etc, all stuff there's no attempt to retry. Bear in mind the logic for the S3ARetryPolicy ignores all of those anyway, so even if caught, they'd not be retried. (the S3Guard dataAccessPolicy is exponential backoff, as before, but as its no longer catching NPEs &c it is having it scope of what-to-retry loosened slightly. 

Oh, and by only handling IOEs we can keep the signature of the invoke calls always {{throws IOE}}.

If you want a non IOE, non SDK exception to be retried on your lambda-exp: catch and translate before you invoke()

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/commit/staging/StagingCommitter.java
steveloughran: I'll leave our retrying in the s3a code for now; there's also whatever goes on in the AWS SDK, which we aren't fully understanding of. Ideally I'd like to hook up the S3a instrumentation & stats to get numbers of what things are being retried in the transfer manager. Work for another day

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ABlockOutputStream.java
```diff
@@ -379,7 +379,7 @@ public void close() throws IOException {
             bytes)) {
           multiPartUpload.complete(partETags);
         } else {
-          LOG.info("upload completion delayed until job commit");
+          LOG.info("File {} will be visible when the job is committed");
```
ajfabbri: missing parameter for LOG.info()

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/commit/AbstractITCommitProtocol.java
```diff
@@ -1257,6 +1293,74 @@ public void testAMWorkflow() throws Throwable {
...
+    } finally {
+      // uncommitted files to this path need to be deleted in tests which fail
+      abortMultipartUploadsUnderPath(job2Dest);
+    }
+
+  }
+
   protected void validateTaskAttemptPathDuringWrite(Path p) throws IOException {
 
   }
```
ajfabbri: Good catch on this case, and thanks for adding a test to cover it.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ABlockOutputStream.java
steveloughran: fixed

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/commit/AbstractITCommitProtocol.java
steveloughran: oh, just debugging things

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/committers.md
```diff
@@ -550,6 +551,102 @@ Conflict management is left to the execution engine itself.
...
+  <name>fs.s3a.committer.staging.conflict-mode</name>
+  <value>fail</value>
+  <description>
+    Staging committer conflict resolution policy.
+    Supported: "fail", "append", "replace".
+  </description>
+</property>
+
+<property>
+  <name>s.s3a.committer.staging.abort.pending.uploads</name>
```
ajfabbri: typo here

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
```diff
@@ -871,6 +871,166 @@ options are covered in [Testing](./testing.md).
...
+1. There is also throttling taking place inside the AWS SDK; this is managed
+by the value `fs.s3a.attempts.maximum`.
+1. Throttling events are tracked in the S3A filesystem metrics and statistics.
+1. Amazon KMS may thottle a customer based on the total rate of uses of 
+KMS *across all user accounts and applications*.
+
+Throttling of S3 requests is all too common; it is caused by too many clients
+trying to access the same shard of S3 Storage. This generatlly
+happen if there are too many reads, those being the most common in Hadoop
+applications. This problem is exacerbated by Hive's partitioning
```
ajfabbri: /generatlly happen/generally happens/

ajfabbri:
Looks good so far.. Commented on a couple of minor typos.

ajfabbri:
These last two commits look fine as well.  I'm +1 as of commit d5dcf98

steveloughran:
:)

PR #287, opened 2017-11-01
HDFS-10323. transient deleteOnExit failure in ViewFileSystem due to close() ordering

wenxinhe:


elek:

PR #289, opened 2017-11-07
[YARN-6483] Add nodes transitioning to DECOMMISSIONING state to the list of updated nodes returned by the Resource Manager as a response to the Application Master heartbeat

juanrh:
This is an alternative approach to https://issues.apache.org/jira/browse/YARN-3224 for notifying all affected application masters when a node transitions into the DECOMMISSIONING state. This change modifies the AllocateResponse that the YARN Resource Manager uses to respond to heartbeat request from application masters, to add any node that has transitioned to DECOMMISSIONING state since the last heartbeat to the list of NodeReport objects that is part of the AllocateResponse object. We also add a new field to each NodeReport to add the decommission timeout for DECOMMISSIONING nodes, thus covering the same functionality of the original proposal in YARN-3224.

### DIFF
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-api/src/main/java/org/apache/hadoop/yarn/api/records/NodeReport.java
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-api/src/main/java/org/apache/hadoop/yarn/api/records/NodeReport.java
@@ -53,7 +53,8 @@ public abstract class NodeReport {
       String httpAddress, String rackName, Resource used, Resource capability,
       int numContainers, String healthReport, long lastHealthReportTime) {
     return newInstance(nodeId, nodeState, httpAddress, rackName, used,
-        capability, numContainers, healthReport, lastHealthReportTime, null, null, null);
+        capability, numContainers, healthReport, lastHealthReportTime,
+        null, null, null);
   }
 
   @Private
@@ -61,7 +62,8 @@ public abstract class NodeReport {
   public static NodeReport newInstance(NodeId nodeId, NodeState nodeState,
       String httpAddress, String rackName, Resource used, Resource capability,
       int numContainers, String healthReport, long lastHealthReportTime,
-      Set<String> nodeLabels, Integer decommissioningTimeout, NodeUpdateType nodeUpdateType) {
+      Set<String> nodeLabels, Integer decommissioningTimeout,
+      NodeUpdateType nodeUpdateType) {
     NodeReport nodeReport = Records.newRecord(NodeReport.class);
     nodeReport.setNodeId(nodeId);
     nodeReport.setNodeState(nodeState);
@@ -188,8 +190,8 @@ public abstract class NodeReport {
   public abstract void setLastHealthReportTime(long lastHealthReport);
   
   /**
-   * Get labels of this node
-   * @return labels of this node
+   * Get labels of this node.
+   * @return labels of this node.
    */
   @Public
   @Stable
@@ -200,8 +202,8 @@ public abstract class NodeReport {
   public abstract void setNodeLabels(Set<String> nodeLabels);
 
   /**
-   * Get containers aggregated resource utilization in a node
-   * @return containers resource utilization
+   * Get containers aggregated resource utilization in a node.
+   * @return containers resource utilization.
    */
   @Public
   @Stable
@@ -219,8 +221,8 @@ public abstract class NodeReport {
   }
 
   /**
-   * Get node resource utilization
-   * @return node resource utilization
+   * Get node resource utilization.
+   * @return node resource utilization.
    */
   @Public
   @Stable
@@ -231,7 +233,8 @@ public abstract class NodeReport {
   public abstract void setNodeUtilization(ResourceUtilization nodeUtilization);
 
   /**
-   * Optional decommissioning timeout in seconds (null indicates absent timeout).
+   * Optional decommissioning timeout in seconds (null indicates absent
+   * timeout).
    * @return the decommissioning timeout in second.
    */
   public Integer getDecommissioningTimeout() {
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-api/src/main/java/org/apache/hadoop/yarn/api/records/NodeReport.java
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-api/src/main/java/org/apache/hadoop/yarn/api/records/NodeReport.java
@@ -53,7 +53,7 @@ public abstract class NodeReport {
       String httpAddress, String rackName, Resource used, Resource capability,
       int numContainers, String healthReport, long lastHealthReportTime) {
     return newInstance(nodeId, nodeState, httpAddress, rackName, used,
-        capability, numContainers, healthReport, lastHealthReportTime, null, null);
+        capability, numContainers, healthReport, lastHealthReportTime, null, null, null);
   }
 
   @Private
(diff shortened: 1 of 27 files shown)


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-api/src/main/java/org/apache/hadoop/yarn/api/records/NodeReport.java
```diff
@@ -53,15 +53,15 @@ public static NodeReport newInstance(NodeId nodeId, NodeState nodeState,
...
+        capability, numContainers, healthReport, lastHealthReportTime, null, null);
   }
 
   @Private
   @Unstable
   public static NodeReport newInstance(NodeId nodeId, NodeState nodeState,
       String httpAddress, String rackName, Resource used, Resource capability,
       int numContainers, String healthReport, long lastHealthReportTime,
-      Set<String> nodeLabels) {
+      Set<String> nodeLabels, Integer decommissioningTimeout) {
```
xslogic: Can we use int ? Am not confortable using Integer - given the sometimes arbitrary boxing/unboxing rules.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/rmnode/RMNodeImpl.java
```diff
@@ -1160,6 +1160,11 @@ public void transition(RMNodeImpl rmNode, RMNodeEvent event) {
       // Update NM metrics during graceful decommissioning.
       rmNode.updateMetricsForGracefulDecommission(initState, finalState);
       rmNode.decommissioningTimeout = timeout;
+      // Notify NodesListManager to notify all RMApp so that each Application Master
+      // could take any required actions.
+      rmNode.context.getDispatcher().getEventHandler().handle(
+          new NodesListManagerEvent(
+              NodesListManagerEventType.NODE_USABLE, rmNode));
```
xslogic: Hmmm.. don't think we should be sending NODE_USABLE event here.. since technically, it is not usable.
Maybe consider creating a new NODE_DECOMMISSIONING event ?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-api/src/main/java/org/apache/hadoop/yarn/api/records/NodeReport.java
juanrh: This `Integer` comes from [`RMNode.getDecommissioningTimeout()`](https://github.com/apache/hadoop/blob/trunk/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/rmnode/RMNode.java#L189) that was already returning an Integer before this patch, because only nodes in DECOMMISSIONING state have an associated decommission timeout, so `null` is used to express absent timeout. In this patch `RMNode.getDecommissioningTimeout()` is used in `DefaultAMSProcessor.handleNodeUpdates` to get the argument `decommissioningTimeout` for `BuilderUtils.newNodeReport`. If we use a `int` for `decommissioningTimeout` in `NodeReport.newInstance` I think we should also use an `int` for the same argument in `BuilderUtils.newNodeReport` for uniformity, which implies a conversion from `null` to `-1` in `DefaultAMSProcessor.handleNodeUpdates`.

So I think we should either keep using `Integer decommissioningTimeout` everywhere, enconding absent timeout with `null`, or use `int decommissioningTimeout` everywhere, enconding absent timeout with a negative timeout, which is coherent with `message NodeReportProto` using an unsigned int for `decommissioning_timeout`. What do you think about these 2 alternatives?

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/rmnode/RMNodeImpl.java
juanrh: I wasn't very sure about using `NODE_USABLE`, but while I was making the changes to follow your suggestion, I have found that in the current code [`TestRMNodeTransitions.testResourceUpdateOnDecommissioningNode`](https://github.com/apache/hadoop/blob/trunk/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/test/java/org/apache/hadoop/yarn/server/resourcemanager/TestRMNodeTransitions.java#L994) is asserting that `NodesListManagerEventType.NODE_USABLE` is expected for a node that transitions to `DECOMMISSIONING`. Also, NodesListManagerEventType is transformed into the corresponding `RMAppNodeUpdateType` in `NodesListManager.handle` to build a `RMAppNodeUpdateEvent` that is processed in `RMAppImpl.processNodeUpdate` which just uses the `RMAppNodeUpdateType` for logging.

So it looks like it is ok to use `NodesListManagerEventType.NODE_USABLE` for nodes in decommissioning state. Do you still think it's worth adding some additional value for `NodesListManagerEventType` and `RMAppNodeUpdateType`?

xslogic: I feel we should make intentions explicit - having a separate event type might make the code cleaner and easier to follow - rather than overloading. It could be that the assumption in the testcase is wrong (will have to double check though), in which case - it is perfectly alright to modify the testcasse with the new event.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-api/src/main/java/org/apache/hadoop/yarn/api/records/NodeReport.java
xslogic: Yeah - sorry, just noticed that. maybe stick with Integer (although I dont like it much). Maybe we can raise another JIRA to fix it properly via your second suggestion.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/rmnode/RMNodeImpl.java
juanrh: Makes sense, I have added a new commit for adding a new value `NODE_DECOMMISSIONING` for `NodesListManagerEventType`

xslogic: Thanks - looking at the patch.. can you also attach a consolidated patch on the JIRA ? So as to kick Jenkins.

juanrh: I have just attached the patch. Thanks a lot for taking a look!

juanrh:
Pushed in https://github.com/apache/hadoop/commit/b46ca7e73b8bac3fdbff0b13afe009308078acf2

PR #291, opened 2017-11-13
HADOOP-15033. Use java.util.zip.CRC32C for Java 9 and above

dchuyko:
I added isJava9OrAbove() flag in Shell which is statically initialized using version string parsing.
If it is true, DataChecksum creates zip implementation instead of PureJavaCrc32C. It is created using constructor method handle that's looked up also if it's 9 or above.
Signature check should ignore java.lang.invoke to let method handles code pass build.
Performance test now outputs extra column for 9 and above with results of zip CRC32C.

Results on Cavium ThunderX server (AArch64) show ~4x improvement and are listed below. The picture is about the same on x86.

|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
|    32 |  1 |     157.6 |     160.1 |   1.6% |     110.3 | -31.1% |       98.9 | -10.3% |
|    32 |  2 |     171.2 |     184.2 |   7.6% |     107.0 | -41.9% |      112.9 |   5.5% |
|    32 |  4 |     182.1 |     176.0 |  -3.3% |     104.0 | -40.9% |      113.3 |   8.9% |
|    32 |  8 |     179.3 |     173.0 |  -3.6% |     102.1 | -41.0% |      111.5 |   9.3% |
|    32 | 16 |      95.0 |      91.3 |  -3.9% |      53.4 | -41.5% |       57.5 |   7.7% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
|    64 |  1 |     326.5 |     324.6 |  -0.6% |     137.4 | -57.7% |      144.7 |   5.3% |
|    64 |  2 |     298.8 |     293.8 |  -1.7% |     133.5 | -54.6% |      139.5 |   4.5% |
|    64 |  4 |     279.3 |     273.3 |  -2.2% |     128.4 | -53.0% |      136.5 |   6.3% |
|    64 |  8 |     269.8 |     267.2 |  -0.9% |     125.9 | -52.9% |      133.6 |   6.1% |
|    64 | 16 |     140.1 |     139.3 |  -0.6% |      65.6 | -52.9% |       69.3 |   5.6% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
|   128 |  1 |     666.6 |     668.6 |   0.3% |     185.0 | -72.3% |      208.7 |  12.8% |
|   128 |  2 |     656.3 |     616.9 |  -6.0% |     184.5 | -70.1% |      207.7 |  12.6% |
|   128 |  4 |     636.7 |     599.7 |  -5.8% |     181.8 | -69.7% |      203.6 |  12.0% |
|   128 |  8 |     628.3 |     592.3 |  -5.7% |     180.6 | -69.5% |      202.2 |  12.0% |
|   128 | 16 |     378.1 |     345.9 |  -8.5% |      93.1 | -73.1% |      108.4 |  16.4% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
|   256 |  1 |     906.4 |     776.1 | -14.4% |     207.9 | -73.2% |      239.2 |  15.0% |
|   256 |  2 |     882.9 |     833.0 |  -5.6% |     206.6 | -75.2% |      238.6 |  15.5% |
|   256 |  4 |     837.9 |     801.0 |  -4.4% |     204.2 | -74.5% |      235.9 |  15.5% |
|   256 |  8 |     806.1 |     776.7 |  -3.7% |     201.7 | -74.0% |      225.6 |  11.8% |
|   256 | 16 |     493.6 |     456.0 |  -7.6% |     107.8 | -76.4% |      121.3 |  12.5% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
|   512 |  1 |    1016.3 |     850.4 | -16.3% |     214.6 | -74.8% |      248.7 |  15.9% |
|   512 |  2 |     969.7 |     894.4 |  -7.8% |     212.1 | -76.3% |      241.5 |  13.8% |
|   512 |  4 |     894.0 |     829.4 |  -7.2% |     207.4 | -75.0% |      232.1 |  11.9% |
|   512 |  8 |     886.8 |     831.8 |  -6.2% |     205.2 | -75.3% |      238.6 |  16.3% |
|   512 | 16 |     518.3 |     525.4 |   1.4% |     111.1 | -78.9% |      124.8 |  12.4% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
|  1024 |  1 |    1120.0 |    1130.5 |   0.9% |     219.1 | -80.6% |      258.0 |  17.8% |
|  1024 |  2 |    1054.9 |     983.9 |  -6.7% |     217.5 | -77.9% |      254.8 |  17.1% |
|  1024 |  4 |     956.8 |     766.0 | -19.9% |     213.0 | -72.2% |      241.1 |  13.2% |
|  1024 |  8 |     835.5 |     864.9 |   3.5% |     209.6 | -75.8% |      247.2 |  18.0% |
|  1024 | 16 |     545.0 |     540.5 |  -0.8% |     110.1 | -79.6% |      131.8 |  19.7% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
|  2048 |  1 |    1189.2 |    1190.5 |   0.1% |     223.3 | -81.2% |      263.3 |  17.9% |
|  2048 |  2 |    1116.4 |    1027.5 |  -8.0% |     220.6 | -78.5% |      260.0 |  17.9% |
|  2048 |  4 |    1009.9 |     931.0 |  -7.8% |     215.3 | -76.9% |      253.3 |  17.6% |
|  2048 |  8 |     950.4 |     881.0 |  -7.3% |     213.2 | -75.8% |      251.5 |  17.9% |
|  2048 | 16 |     570.7 |     521.7 |  -8.6% |     111.7 | -78.6% |      132.1 |  18.3% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
|  4096 |  1 |    1264.7 |     991.2 | -21.6% |     225.7 | -77.2% |      267.6 |  18.6% |
|  4096 |  2 |    1161.3 |    1023.4 | -11.9% |     222.2 | -78.3% |      263.2 |  18.5% |
|  4096 |  4 |    1037.7 |     931.6 | -10.2% |     217.4 | -76.7% |      256.2 |  17.9% |
|  4096 |  8 |    1010.8 |     925.2 |  -8.5% |     214.7 | -76.8% |      255.4 |  18.9% |
|  4096 | 16 |     626.8 |     575.2 |  -8.2% |     115.4 | -79.9% |      129.8 |  12.5% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
|  8192 |  1 |    1264.7 |    1264.0 |  -0.1% |     225.2 | -82.2% |      269.3 |  19.6% |
|  8192 |  2 |    1182.1 |    1077.8 |  -8.8% |     222.1 | -79.4% |      264.6 |  19.1% |
|  8192 |  4 |    1056.2 |     968.4 |  -8.3% |     217.3 | -77.6% |      257.6 |  18.6% |
|  8192 |  8 |    1015.0 |     890.3 | -12.3% |     216.0 | -75.7% |      255.1 |  18.1% |
|  8192 | 16 |     616.7 |     571.2 |  -7.4% |     113.1 | -80.2% |      135.5 |  19.8% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
| 16384 |  1 |    1226.4 |    1116.2 |  -9.0% |     221.6 | -80.1% |      262.9 |  18.6% |
| 16384 |  2 |    1152.5 |    1080.3 |  -6.3% |     219.0 | -79.7% |      259.1 |  18.3% |
| 16384 |  4 |     895.7 |     976.4 |   9.0% |     214.5 | -78.0% |      253.1 |  18.0% |
| 16384 |  8 |     985.8 |     961.2 |  -2.5% |     211.0 | -78.0% |      252.2 |  19.5% |
| 16384 | 16 |     567.6 |     565.2 |  -0.4% |     110.0 | -80.5% |      133.6 |  21.5% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
| 32768 |  1 |    1107.3 |     913.8 | -17.5% |     217.2 | -76.2% |      257.1 |  18.3% |
| 32768 |  2 |    1026.4 |     998.0 |  -2.8% |     214.2 | -78.5% |      252.2 |  17.7% |
| 32768 |  4 |     942.1 |     908.1 |  -3.6% |     209.9 | -76.9% |      246.8 |  17.6% |
| 32768 |  8 |     896.2 |     798.5 | -10.9% |     206.2 | -74.2% |      245.3 |  19.0% |
| 32768 | 16 |     549.0 |     504.9 |  -8.0% |     109.6 | -78.3% |      127.7 |  16.5% |
|  bpc  | #T ||      Zip ||     ZipC | % diff || PureJava | % diff || PureJavaC | % diff |
| 65536 |  1 |    1101.6 |    1078.4 |  -2.1% |     217.2 | -79.9% |      256.3 |  18.0% |
| 65536 |  2 |    1026.1 |     846.7 | -17.5% |     213.7 | -74.8% |      253.0 |  18.4% |
| 65536 |  4 |     936.5 |     841.9 | -10.1% |     210.1 | -75.0% |      246.7 |  17.4% |
| 65536 |  8 |     899.7 |     886.5 |  -1.5% |     207.9 | -76.5% |      242.9 |  16.8% |
| 65536 | 16 |     497.8 |     511.7 |   2.8% |     107.1 | -79.1% |      124.2 |  16.0% |


hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
```diff
@@ -78,6 +100,16 @@ public static Checksum newCrc32() {
     return new CRC32();
   }
 
+  public static Checksum newCrc32C() {
+    try {
+      return Shell.isJava9OrAbove() ? (Checksum) NEW_CRC32C_MH.invoke()
+              : new PureJavaCrc32C();
+    } catch (Throwable e) {
+      // Should not reach here.
+      throw new RuntimeException(e);
```
steveloughran: RuntimeException shouldn't be wrapped; leave the other Throwables alone. So: `catch(RuntimeException e) { throw e;}`, then catch and wrap Exception only,

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
```diff
@@ -70,6 +74,24 @@ public static Type valueOf(int id) {
     }
   }
 
+  private static final MethodHandle NEW_CRC32C_MH;
+  static {
```
steveloughran: I'd like this to all be on demand, so that the introspection is only done if needed, which allows the exception to be thrown at the right point.

Also, could we just back off from failing here at all, just log, track that the operation failed (so no attempt is repeated), and then fall back to the older mechanisms. I don't want things to overreact here

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/Shell.java
```diff
@@ -89,6 +89,22 @@ public static boolean isJava7OrAbove() {
...
+      String version = System.getProperty("java.version");
+      major = Integer.parseInt(version.split("\\.")[0].split("-")[0]);
+    } catch (NumberFormatException ignored) {
+    }
+    IS_JAVA9_OR_ABOVE = major >= 9;
+  }
+
+  public static boolean isJava9OrAbove() {
+    return IS_JAVA9_OR_ABOVE;
+  }
```
steveloughran: Given we've already got isJava7OrAbove() tagged as deprecated, we should stop repeating the same mistake.

How about some: `boolean IsJavaVersionAtLeast(int)` which can do checks for versions, store this parsed major in a private static (set the default value to 8, obviously) field, then have the new predicate return true if the param is >= the cached value

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
dchuyko: Got it, this will be more clean.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
dchuyko: Agree, logging would go better. I was trying to avoid silent failure here.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/Shell.java
dchuyko: Yes, I had exactly the same in mind. Just haven't realized we may now assume 8 as lowest version. Will it still be true if someone decides to backport the fix to previous Hadoop versions?

steveloughran: Hadoop trunk is java 8+ only.

* nobody will backport your patch to branch-2; its using Java 8 classes.
* someone could backport an `isJavaVersionAtLeast()` method to branch-2, in which case, they get to extend it.

dchuyko:
Ok, I changed the patch according to the review.

1. There's now isJavaSpecAtLeast() in Shell. This is a part I like.

2. In DataChecksum there are now: racy state "useJava9Crc32C" which is set to the above by default and falls to false on error. New Java9Crc32CFactory now holds static method handles stuff to let it be initialized on first usage. And there is now also LOG to possibly log errors that never happen. To me it looks probably like too much meat for code that won't be called ever. And in case of NOP logger no one will see the error if it happens still. From the other hand all that may look more safe.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
```diff
@@ -43,6 +49,9 @@
   public static final int CHECKSUM_CRC32C  = 2;
   public static final int CHECKSUM_DEFAULT = 3; 
   public static final int CHECKSUM_MIXED   = 4;
+
+  public static final Logger LOG = LoggerFactory.getLogger(DataChecksum.class);
```
steveloughran: make private

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
```diff
@@ -78,6 +87,18 @@ public static Checksum newCrc32() {
     return new CRC32();
   }
 
+  public static Checksum newCrc32C() {
```
steveloughran: again, private. Comment should say "use a volatile to avoid a lock here; re-entrancy unlikely  except in failure mode (and inexpensive)

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/Shell.java
```diff
@@ -86,7 +86,15 @@
    */
   @Deprecated
   public static boolean isJava7OrAbove() {
-    return true;
+    return isJavaSpecAtLeast(7);
```
steveloughran: no, this can stay true. We don't support old stuff any more

What would be good is to have a test in {{TestShell}} which calls the new API operation & sees that it is consistent with its expectations (i.e on trunk it must be 8+)

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/Shell.java
```diff
@@ -86,7 +86,15 @@
...
+  }
+
+  // "1.8"->8, "9"->9, "10"->10
+  private static final int JAVA_SPEC_VER = Math.max(8, Integer.parseInt(
+      System.getProperty("java.specification.version").split("\\.")[0]));
+
+  public static boolean isJavaSpecAtLeast(int version) {
```
steveloughran: 1. Can we call this `isJavaVersionAtLeast` ? 
1. + javadoc explaining what it does
1. Needs a test in `TestShel` whichcalls the new API operation & sees that it is consistent with its expectations (i.e on trunk it must be 8+)

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
dchuyko: Made private. It was copied from the wrong place (1/3 cases in hadoop-common btw.)

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
dchuyko: Made it package visible to keep it accessible from Crc32PerformanceTest. Note, newCrc32() is still public. Added javadoc to the method.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/Shell.java
dchuyko: Ok, reverted.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/Shell.java
dchuyko: Done all 3.

dchuyko:
A copy of the patch from the pull request was attached as HADOOP-15033.010.patch in Jira. The patch passed genericqa checks and shows same performance improvement a before. Please review this latest version of pull request.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
```diff
@@ -43,6 +49,9 @@
   public static final int CHECKSUM_CRC32C  = 2;
   public static final int CHECKSUM_DEFAULT = 3; 
   public static final int CHECKSUM_MIXED   = 4;
+
+  private static final Logger LOG = LoggerFactory.getLogger(DataChecksum.class);
+  private static volatile boolean useJava9Crc32C = Shell.isJavaVersionAtLeast(9);
```
aajisaka: @steveloughran @dchuyko The flag `useJava9Crc32C` can be final and I'm thinking changing it final also avoids synchronization. What do you think?

Sorry for very late.

dchuyko: @aajisaka The initial design was to have constructor method handle as static final in the main class and to not use the flag at all.
There was a concern that Java may not work as it must work: class or its constructor available by spec won't be found by standard means or there will be an error during invocation. I'd assume that in this case it would be fine to fail globally. Also some code around this stuff may be considered possible to fail.
These scenarios are described by try-catch blocks. In case something bad has happened on 9+ we switch the flag to use old internal plain java implementation and log the error.
The API should be thread safe/correct so we can imagine a race where multiple threads try the initialization, get errors and put down the flag.

aajisaka: Thank you for the detailed comment. Sorry I missed the code that put downs the flag when get error. Now I'm +1.

dchuyko:
Many thanks for the review!

aajisaka:
@dchuyko Thank you for your contribution! Would you close this pull request?

PR #293, opened 2017-11-20
YARN-7480 Renders tooltip on em-table cells

skmvasu:
Helper for rendering tooltip on em-table cells. Helpful for columns, where the field name is large


skmvasu:
@sunilgovind  Please review

PR #311, opened 2017-12-07
[HADOOP-15096] Don't create user with lastlog

addisonj:
This fixes a problem where in certain cases, the useradd command can create a very large diff that can blow up the host disk size.

The reason for this is that lastlog is a sparse file, but AUFS under docker apparently doesn't deal with those well and creates a very large file.


elek:

PR #318, opened 2017-12-23
[HADOOP-15214] Make Hadoop compatible with Guava 21

medb:
After this change Hadoop could build against Guava 21.0

Justification for migration is the same as previous migrations here:
https://issues.apache.org/jira/browse/HADOOP-11032


medb:
Was merged in https://issues.apache.org/jira/browse/HADOOP-15214

PR #319, opened 2017-12-23
HADOOP 15142. Register FTP and SFTP as FS services

mmolimar:


elek:

PR #332, opened 2018-01-24
The log about the node status is changed from debug or info to warn w…

zhengChina:
https://issues.apache.org/jira/browse/YARN-7809


elek:

PR #333, opened 2018-01-26
YARN-7823. add configuration to use ip as NM address rather than hostname

xulongzhe:


mikechengwei:
@xulongzhe hi,why not you continue committing this change ?

PR #339, opened 2018-02-13
HADOOP-13972 Supporting per-account configuration for ADL

ssonker:


ssonker:
@steveloughran @jzhuge Can you please have a look at this?

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/TestValidateConfiguration.java
```diff
@@ -152,6 +152,95 @@ public void testLoadDeprecatedKeys()
...
+  @Test
+  public void testPropagateAccountOptionsSpecified() throws IOException {
+    Configuration conf = new Configuration(false);
+    conf.set("fs.adl.account.dummy.oauth2.client.id", "dummyClientId");
+    conf.set("fs.adl.account.dummy.oauth2.credential", "dummyCredential");
+    conf.set("some.other.config", "someValue");
+
+    Configuration propagatedConf =
+        AdlFileSystem.propagateAccountOptions(conf, "dummy");
+    Assert.assertEquals("dummyClientId",
```
steveloughran: just statically import Assert, or have the test case extend Assert

hadoop-tools/hadoop-azure-datalake/src/main/java/org/apache/hadoop/fs/adl/AdlFileSystem.java
```diff
@@ -65,6 +66,8 @@
 import org.apache.hadoop.util.Progressable;
 import org.apache.hadoop.util.ReflectionUtils;
 import org.apache.hadoop.util.VersionInfo;
+import org.slf4j.Logger;
```
steveloughran: should be ordered according to the hadoop import ordering. I know we get this wrong, and once its wrong we don't like to fix (it breaks all patches and cherry picking), but it's good to try

hadoop-tools/hadoop-azure-datalake/src/main/java/org/apache/hadoop/fs/adl/AdlFileSystem.java
```diff
@@ -985,4 +994,65 @@ public void setUserGroupRepresentationAsUPN(boolean enableUPN) {
     oidOrUpn = enableUPN ? UserGroupRepresentation.UPN :
         UserGroupRepresentation.OID;
   }
+
+  /**
+   * Gets ADL account name from ADL FQDN
+   * @param accountFQDN ADL account fqdn
+   * @return ADL account name
+     */
```
steveloughran: indentation

hadoop-tools/hadoop-azure-datalake/src/main/java/org/apache/hadoop/fs/adl/AdlFileSystem.java
```diff
@@ -985,4 +994,65 @@ public void setUserGroupRepresentationAsUPN(boolean enableUPN) {
...
+      final String stripped = key.substring(accountPrefix.length());
+
+      // propagate the value, building a new origin field.
+      // to track overwrites, the generic key is overwritten even if
+      // already matches the new one.
+      String origin = "[" + StringUtils.join(
+              source.getPropertySources(key), ", ") +"]";
+      final String generic = AZURE_AD_PREFIX + stripped;
+      LOG.debug("Updating {} from {}", generic, origin);
+      dest.set(generic, value, key + " via " + origin);
```
steveloughran: nice!

hadoop-tools/hadoop-azure-datalake/src/site/markdown/index.md
```diff
@@ -32,6 +32,7 @@ This support comes via the JAR file `azure-datalake-store.jar`.
 * Tested for scale.
 * API `setOwner()`, `setAcl`, `removeAclEntries()`, `modifyAclEntries()` accepts UPN or OID
   (Object ID) as user and group names.
+* Supports per-account configuration
```
steveloughran: add "."

steveloughran:
As usual: which endpoint have you run the entire ADL test suite against? Have you also tried it with per account settings and no default values?

ssonker:
@steveloughran I've incorporated the review comments, please have a look.

ssonker:
@steveloughran I have a personal ADL account against which I ran the test suite. I also tried it with per account settings with no default values, and all the test cases passed.

steveloughran:
OK. For absolute strictness, and this is the same rules we apply everywhere: which azure endpoint (e.g ireland, US, etc). (this declaration forces max honesty)

BTW, have a look at https://github.com/steveloughran/cloudstore/releases/tag/tag_2018-02-13-release-2 and see what it does when given an ADL fs URI with this patch: it should list the origin of the patched properties https://github.com/steveloughran/cloudstore/releases/tag/tag_2018-02-13-release-2

ssonker:
@steveloughran The account I'm using is adl://tmphadooptest.azuredatalakestore.net which is located in Central US.
I'll update with the cloudstore dump!

ssonker:
@steveloughran I have to make some changes in cloudstore (specifically in ```ADLDiagnosticsInfo.java```) so that it can get the patched/propagated properties from ```AdlFileSystem```. Now when I run cloudstore, I get:
```
Hadoop information
==================

  Hadoop 3.2.0-SNAPSHOT
  Compiled by ssonker on 2018-02-14T04:08Z
  Compiled with protoc 2.5.0
  From source with checksum b89064fa998ad944e87adfda36171356

Diagnostics for filesystem adl://tmphadooptest.azuredatalakestore.net/
======================================================================

Azure Datalake connector
ASF Filesystem Connector to Microsoft Azure Datalake
https://hadoop.apache.org/docs/current/hadoop-azure-datalake/index.html

System Properties
=================
....
....

Selected and Sanitized Configuration Options
============================================

fs.adl.oauth2.client.id = "6**********************************1" [fs.adl.account.tmphadooptest.oauth2.client.id via [core-site.xml]]
fs.adl.oauth2.credential = "9******************************************=" [fs.adl.account.tmphadooptest.oauth2.credential via [core-site.xml]]
fs.adl.oauth2.access.token.provider.type = "ClientCredential" [core-default.xml]
fs.adl.oauth2.refresh.token = (unset)
fs.adl.oauth2.devicecode.clientapp.id = (unset)
fs.adl.oauth2.msi.port = (unset)
fs.adl.oauth2.refresh.url = "https://login.microsoftonline.com/fa****e1/oauth2/token" [fs.adl.account.tmphadooptest.oauth2.refresh.url via [core-site.xml]]
adl.feature.client.cache.readahead = (unset)
adl.feature.client.cache.drop.behind.writes = (unset)
adl.debug.override.localuserasfileowner = (unset)

Classes
=======

class org.apache.hadoop.fs.adl.AdlFileSystem was found in file:/Users/ssonker/hadoop/hadoop-dist/target/hadoop-3.2.0-SNAPSHOT/share/hadoop/tools/lib/hadoop-azure-datalake-3.2.0-SNAPSHOT.jar
class com.microsoft.azure.datalake.store.ADLStoreClient was found in file:/Users/ssonker/hadoop/hadoop-dist/target/hadoop-3.2.0-SNAPSHOT/share/hadoop/tools/lib/azure-data-lake-store-sdk-2.2.5.jar

Endpoint: https://login.microsoftonline.com/fa****e1/oauth2/token:
==============================================================================================

Canonical hostname 52.172.144.29
  IP address 52.172.144.29
Connecting to https://login.microsoftonline.com/fa****e1/oauth2/token
Response: 200 : OK
HTTP response 200 from https://login.microsoftonline.com/fa****e1/oauth2/token: OK
Using proxy: false 
....
....
```

steveloughran:
OK. Go the Hadoop JIRA, hit "submit patch" and see what it says. I Can see it'll be complaining about the javadocs, but let's see what else.

hadoop-tools/hadoop-azure-datalake/src/main/java/org/apache/hadoop/fs/adl/AdlFileSystem.java
```diff
@@ -68,12 +69,16 @@
 
 import static org.apache.hadoop.fs.adl.AdlConfKeys.*;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
steveloughran: statics at the bottom; these should go into the same block as com.*, ideally in order.

hadoop-tools/hadoop-azure-datalake/src/main/java/org/apache/hadoop/fs/adl/AdlFileSystem.java
```diff
@@ -985,4 +995,65 @@ public void setUserGroupRepresentationAsUPN(boolean enableUPN) {
...
+   *
+   * @param source Source Configuration object
+   * @param accountName account name. Must not be empty
+   * @return a (potentially) patched clone of the original
+   * @throws IOException
+   */
+  public static Configuration propagateAccountOptions(
+      Configuration source, String accountName) throws IOException {
+    if (StringUtils.isEmpty(accountName)) {
+      throw new IOException("Account name cannot be empty");
```
steveloughran: We're generally happy with {{Precondition.checkArgument()}} here, which raises IllegalArgumentException. Best just to use that, unless you have tests which look for the specific exception

steveloughran:
apart from the comments inline, I've got nothing else to add...lets see what yetus says. Good to see from the cloudstore test that the option provenance is being set, as it will help debugging in a world where things get more complex

ssonker:
@steveloughran Where can I find the build status of the patch. I was looking at https://builds.apache.org/view/H-L/view/Hadoop/job/PreCommit-HADOOP-Build/ but couldn't locate this patch. Thanks.

steveloughran:
the results should be added as a JIRA comment once completed

ssonker:
@steveloughran Yeah, I read that in [_How to contribute_ wiki](https://wiki.apache.org/hadoop/HowToContribute#Contributing_your_work), but I wanted to have a look at jenkins as well. BTW, how long does it take after submitting the patch?

steveloughran:
2-3 hours, normally. It can get a bit confused about PRs, which is why I've stopped doing it, even though github review is nice. 

if nothing happens, create a new JIRA, same title + "Yetus patch checker", mark as a duplicate of the original, and do a `git diff trunk...HEAD > HADOOP-13972-001.patch` call to build that first patch, attach it to the JIRA and submit. here's [an example](https://issues.apache.org/jira/browse/HADOOP-15003)

ssonker:
@steveloughran Please see [HADOOP-15238](https://issues.apache.org/jira/browse/HADOOP-15238) for the Yetus patch checker. Yetus successfully verified the patch.

steveloughran:
@ssonker can you close this PR now it's been merged in? thanks

ssonker:
This PR is merged as [this commit](https://github.com/apache/hadoop/commit/481d79fedc48942654dab08e23e71e80c8eb2aca). Therefore, closing it.

PR #346, opened 2018-02-25
YARN-7969 Fix 3 confusing log messages that do not match with the method name

lzh3636:


elek:

PR #355, opened 2018-03-19
HDFS-13313. Fix NullPointerException in FSEditLogOp.toString().

glglwty:
In some subclasses of FSEditLogOp, toString() uses dot(.) operator on
some fields. Those fields could be set to null by resetSubFields() and a
NullPointerException would be thrown in toString(). This patch fixes
this bug by calling StringBuilder.append(field). Thus "null" would be
printed in the field=null case, and field.toString() code is properly
reused to avoid repetition.

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSEditLogOp.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSEditLogOp.java
@@ -3826,11 +3826,7 @@ public abstract class FSEditLogOp {
     public String toString() {
       StringBuilder builder = new StringBuilder();
       builder.append("AddCacheDirectiveInfo [");
-      builder.append("id=" + directive.getId() + ",");
-      builder.append("path=" + directive.getPath().toUri().getPath() + ",");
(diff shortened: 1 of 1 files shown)


hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSEditLogOp.java
```diff
@@ -3826,11 +3826,7 @@ void fromXml(Stanza st) throws InvalidXmlException {
     public String toString() {
       StringBuilder builder = new StringBuilder();
       builder.append("AddCacheDirectiveInfo [");
-      builder.append("id=" + directive.getId() + ",");
-      builder.append("path=" + directive.getPath().toUri().getPath() + ",");
-      builder.append("replication=" + directive.getReplication() + ",");
-      builder.append("pool=" + directive.getPool() + ",");
-      builder.append("expiration=" + directive.getExpiration().getMillis());
+      builder.append(directive);
```
lqjack: What if directive is null?

glglwty: String "null" will be appended to the StringBuilder

bak12:
J

elek:

PR #368, opened 2018-04-27
HADOOP-15380

lqjack:
move crc file to trash if exist


elek:

PR #370, opened 2018-04-28
HADOOP-15427

lqjack:
change hadoop_error to hadoop_debug


lqjack:
The suggestion from https://issues.apache.org/jira/browse/HADOOP-15427

elek:

PR #371, opened 2018-04-29
HDFS-13515

lqjack:
provide the info when connect


hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/net/NetUtils.java
```diff
@@ -530,6 +530,7 @@ public static void connect(Socket socket,
       } else {
         SocketIOWithTimeout.connect(ch, endpoint, timeout);
       }
+      LOG.info("endpoint : {} , timeout : {} connect successful",endpoint,timeout);
```
functicons: LOG.info("endpoint : {} , timeout : {}, connected successfully", endpoint, timeout);

elek:

ayushtkn:
raise PR to trunk, not to a release branch

PR #373, opened 2018-04-30
HADOOP-15428

lqjack:
fix error prompt


elek:

PR #396, opened 2018-06-23
classpath description is not full for hadoop cmd

sandeepksaini:
Changed line number 43 where classpath description is not full and changed it to the one provided in hadoopv3:  "prints the class path needed to get the Hadoop jar and the required libraries"


elek:

PR #400, opened 2018-07-05
HADOOP-15585. Fix passing options via $HADOOP_OPTS

bschell:
Prior to this change, if HADOOP_OPTS contains any arguments that include a space, the command is not parsed correctly. The only alternative appears to be to use 'eval'. Switching to use 'eval' *instead of* 'exec' also works, but it results in an intermediate bash process being left alive throughout the entire lifetime of the Java process being started. Using 'exec' prefixed by 'eval' as has been done in this commit gets the best of both worlds, in that options with spaces are parsed correctly, and you don't end up with an intermediate bash process as the parent of the Java process.


elek:

PR #407, opened 2018-08-03
MAPREDUCE-7127. Add aggregated webservice endpoints to fetch all tasks & their taskAttempts

bschell:
There is a usecase to poll Hadoop for various running Tasks and display info on each individual task attempt to the user. To improve performance add additional endpoints to HS and AM webservices that will fetch aggregated data on hadoop tasks including their task attempts for a current job.


elek:

PR #413, opened 2018-09-07
Adding EC2ContainerCredentialsProviderWrapper to the credential providers

nrworld:
This change adds EC2ContainerCredentialsProviderWrapper to the credential providers list.

It will enable hadoop-aws library to be used on AWS fargate.


elek:

steveloughran:
Can you create an ASF JIRA for this under HADOOP-15620; we don't directly look at the PR list, so things like this get lost. 

now, regarding the patch, I don't think it is directly needed....you can just declare that as the credential provider to be `com.amazonaws.auth.EC2ContainerCredentialsProviderWrapper` and it will picked up automatically. We just do the default set so that most people don't need to worry about it (and we've just added the session settings to the list).

How about you add something to the S3A docs telling people what to do?

PR #427, opened 2018-10-15
YARN-6636. Add basic fairscheduler nodelabel support.

bschell:
Supports unfair label scheduling and non-exclusive node labels.


elek:

PR #433, opened 2018-10-23
HADOOP-15870 S3AInputStream.remainingInFile should use nextReadPos

lqjack:
Fix the nextPos

### DIFF
--- a/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractSeekTest.java
+++ b/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractSeekTest.java
@@ -99,14 +99,18 @@ public abstract class AbstractContractSeekTest extends AbstractFSContractTestBas
     describe("seek and read a 0 byte file");
     instream = getFileSystem().open(zeroByteFile);
     assertEquals(0, instream.getPos());
+    assertEquals(0,instream.available());
     //expect initial read to fai;
     int result = instream.read();
     assertMinusOne("initial byte read", result);
+    assertEquals(0,instream.available());
     byte[] buffer = new byte[1];
     //expect that seek to 0 works
     instream.seek(0);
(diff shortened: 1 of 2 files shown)


steveloughran:
1. I think the same problem surfaces in `remainingInCurrentRequest()` too, which is where some inspection is going to be needed to carefully understand what's up.
1. For this method, I think maybe we could cut it completely, and fix up `available()` to return a default value when `wrappedStream==null` (0 or maybe 1 except when at EOF), and when the stream isn't null, delegate to it. That pushes down the problem of estimating the number of non-blocking bytes into the http connector.

Tests.

In `AbstractContractSeekTest` there's some existing tests which could take some more asserts on that available call

* `testSeekZeroByteFile`: `available() == 0`, always
* `testSeekReadClosedFile` : call available() on an empty file, it should raise some IllegalStateException or IOE, but not an NPE

For some of the other tests, I think you insert a couple of checks to say " available > 0" after a seek + read() Call. This also clarifies that "what should be available after a seek but before a read()?" As for cloudstores with lazy seek, available == 0, though if that breaks gzip then maybe they should return 1, so the assert should be available >1 with an option in the contract to say "actually returns 0". 

It's a tricky one to test on because really, the sole asserts should be
1. fails if the stream is closed (unless its a 0?)
1. returns >= if the stream is open
1. is always < remaining file length.

Probably assert #3 is the one to check for: available() is valid when it is >=0 and <= (filelength -getPos()), with care taken about that second clause to make sure there's no off-by-one errors in the assert/code

lqjack:
@steveloughran  Thanks very much , I will update.

lqjack:
@steveloughran I have updated the code , please help reveiw and comment, thanks .

lqjack:
@steveloughran  Due to the upgrade to the latest version of the Mac OS, I need take some time to resolve the environment before I can run the test successful .

steveloughran:
no worries

elek:

steveloughran:
this is merged in, closing

PR #439, opened 2018-11-15
YARN-8833 fix compute shares may  lock the scheduling process

yoelee:
When compute fair share, there may be a chance triggering the problem of Integer overflow, and entering an infinite loop, which blocks the scheduling process.


elek:

PR #449, opened 2018-12-18
HDFS-14158. Fix the Checkpointer not to ignore the configured "dfs.namenode.checkpoint.period" > 5 minutes

tiwalter:
URL: https://issues.apache.org/jira/browse/HDFS-14158


elek:

aajisaka:
This issue is not a problem. Just closing.

hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/Checkpointer.java
```diff
@@ -160,7 +159,7 @@ public void run() {
         break;
       }
       try {
-        Thread.sleep(periodMSec);
+        Thread.sleep(LongMath.gcd(periodMSec, checkpointPeriodMSec));
```
aajisaka: If periodMsec is set to 60 and checkPointPeriodMsec is set to 503, the sleep time becomes 1 (= gcd(60, 503)) sec and edit log check happens every seconds. I suggest sleeping periodMsec instead of gcd.

tiwalter: Hi @aajisaka, 
i came to the same conclusion, but if you choose periodMsec instead, the next checkpoint would be processed after 540 seconds (it's not the configured checkpoint period). 

What if you create a new variable 'lastEditLogCheckTime' and add an if-statement before checking the edit log `if(now >= lastEditLogCheckTime + periodMSec)`?

Then we are safe and don't miss any of the configured periods and we don't check to often.

What's your opinion?

aajisaka: Now periodMsec is min of checkpointCheckPeriod and checkpointPeriod, so the sleep time is smaller than or equal to the configured checkpoint period. Isn't it?

tiwalter: Yes, exactly.

tiwalter:
Hi @aajisaka,
the last commit contains the suggested if statement before the edit log size check. Now we are safe that we don't check to often and don't miss any configured **checkpointPeriod** as well as **checkpointCheckPeriod**.

PR #467, opened 2019-01-16
[HDFS-14208]fix bug that there is some a large number missingblocks after failove to active

ZanderXu:
when the cluster is very large, Standby startup takes about 1 hour.
Standby starts and exits the safe mode, and failover to active, a large number of missingblocks often appear. And it takes about 6 hours before missingblocks gradually disappear.

[HDFS-14208](https://issues.apache.org/jira/browse/HDFS-14208)

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManager.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManager.java
@@ -2532,8 +2532,9 @@ public class BlockManager implements BlockStatsMXBean {
         return !node.hasStaleStorages();
       }
       if (context != null) {
-        if (!blockReportLeaseManager.checkLease(node, startTime,
-              context.getLeaseId())) {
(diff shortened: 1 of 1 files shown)


elek:

brahmareddybattula:
Standby was stopped for some time before it starts..? can we've one testcase for this..?

hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManager.java
```diff
@@ -2532,8 +2532,9 @@ public boolean processReport(final DatanodeID nodeID,
         return !node.hasStaleStorages();
       }
       if (context != null) {
-        if (!blockReportLeaseManager.checkLease(node, startTime,
-              context.getLeaseId())) {
+        if (!namesystem.isInStartupSafeMode()
```
jojochuang: I think this is a dup of HDFS-12914. In HDFS-12914, this part of code is moved to another place to fix the bug.

This fix is essentially saying if namenode is in safe mode, don't check for lease.

PR #484, opened 2019-02-14
HDDS-1103. Fix rat/findbug/checkstyle errors in ozone/hdds projects

elek:
see: https://issues.apache.org/jira/browse/HDDS-1103


elek:
acceptance tests are also fixed: I moved the auditlogparser to a separated directory (it calculates the number of the audit events, should be executed together with other tests) and added a sleep to be sure that the om is started.

Removed the "-it" flags from docker-compose executions to make it possible to run from docker in docker.

xiaoyuyao:
Thanks @elek  for fixing these. +1 pending Jenkins.

elek:
Remaining test failures are not related AFAIK.

elek:
Only the licence header is conflicted with HDDS-1110. I rebased.

elek:
Thanks the review and the merge.

PR #486, opened 2019-02-15
HDDS-1092. Use Java 11 JRE to run Ozone in containers

elek:
see: https://issues.apache.org/jira/browse/HDDS-1092


elek:
Note: I will wait for a normal jenkins execution first. If works well, I can commit the trunk patch first. 
If trunk patch is in, I can bump the java version to java 11 in the base image (hadoop-docker-runner branch)

elek:
Only the ozonefs is failed which will be fixed in #488.

PR #488, opened 2019-02-15
HDDS-1114. Fix findbugs/checkstyle/accepteance errors in Ozone

elek:
Unfortunately as the previous two big commits (error handling HDDS-1068, checkstyle HDDS-1103) are committed in the same time a few new errors are introduced during the rebase.

This patch will fix the remaining 5 issues (+ a type in the acceptance test executor) 

See: https://issues.apache.org/jira/browse/HDDS-1114


elek:
It turned out that the error message always should be filled for OMException. The status code itself is not enough, as external clients (eg. ozonefs) will check only the getMessage of the exception.

PR #489, opened 2019-02-15
HDDS-1115. Provide ozone specific top-level pom.xml

elek:
Ozone build process doesn't require the pom.xml in the top level hadoop directory as we use hadoop 3.2 artifacts as parents of hadoop-ozone and hadoop-hdds. The ./pom.xml is used only to include the hadoop-ozone/hadoop-hdds projects in the maven reactor.

From command line, it's easy to build only the ozone artifacts:

{code}
mvn clean install -Phdds  -am -pl :hadoop-ozone-dist  -Danimal.sniffer.skip=true  -Denforcer.skip=true
{code}

Where: '-pl' defines the build of the hadoop-ozone-dist project
and '-am' defines to build all of the dependencies from the source tree (hadoop-ozone-common, hadoop-hdds-common, etc.)

But this filtering is available only from the command line.

With providing a lightweight pom.ozone.xml we can achieve the same:

 * We can open only hdds/ozone projects in the IDE/intellij. It makes the development faster as IDE doesn't need to reindex all the sources all the time + it's easy to execute checkstyle/findbugs plugins of the intellij to the whole project.
 * Longer term we should create an ozone specific source artifact (currently the source artifact for hadoop and ozone releases are the same) which also requires a simplified pom.

In this patch I also added the .mvn directory to the .gitignore file.

With 
{code}
mkdir -p .mvn && echo "-f ozone.pom.xml" > .mvn/maven.config" you can persist the usage of the ozone.pom.xml for all the subsequent builds (in the same dir)

How to test?

Just do a 'mvn -f ozonze.pom.xml clean install -DskipTests'

See: https://issues.apache.org/jira/browse/HDDS-1115

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,5 @@ output.xml
 report.html
 
 hadoop-hdds/docs/public
+
+.mvn
--- a/pom.ozone.xml
+++ b/pom.ozone.xml
@@ -1,0 +1,310 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


anuengineer:
Thank you for your contribution. I am going to commit this.  Just a nit as an FYI. The pom is named pom.ozone.xml and not as 

> Just do a 'mvn -f ozonze.pom.xml clean install -DskipTests' 

as you commented. I am committing it as-is; since for me, the **pom.ozone.xml** sounds as sweet as **ozone.pom.xml**.

PR #491, opened 2019-02-15
HDDS-1116. Add java profiler servlet to the Ozone web servers

elek:
Thanks to [~gopalv] we learned that [~prasanth_j] implemented a helper servlet in Hive to initialize new [async profiler|https://github.com/jvm-profiling-tools/async-profiler] sessions and provide the svg based flame graph over HTTP. (see HIVE-20202)

It seems to very useful as with this approach the profiling could be very easy.

This patch imports the servlet from the Hive code base to the Ozone code base with minor modification (to make it work with our servlet containers)

 * The two servlets are unified to one
 * Streaming the svg to the browser based on IOUtils.copy 
 * Output message is improved

By default the profile servlet is turned off, but you can enable it with 'hdds.profiler.endpoint.enabled=true' ozone-site.xml settings. In that case you can access the /prof endpoint from scm,om,s3g. 

You should upload the async profiler first (https://github.com/jvm-profiling-tools/async-profiler) and set the ASYNC_PROFILER_HOME environment variable to find it. 

See: https://issues.apache.org/jira/browse/HDDS-1116


hadoop-ozone/dist/src/main/compose/ozone/docker-compose.yaml
```diff
@@ -18,6 +18,7 @@ version: "3"
 services:
    datanode:
       image: apache/hadoop-runner
+      privileged: true #required by the profiler
```
prasanthj: This can be avoided with initContainer running in privileged mode that updates the following
```
            sudo bash -c 'echo 1 > /proc/sys/kernel/perf_event_paranoid'
            sudo bash -c 'echo 0 > /proc/sys/kernel/kptr_restrict'
```
with this initContainer will apply the required changes and will complete. main container can still run in non-previleged mode.

prasanthj: Sorry, I missed the part that it is running using docker-compose and not k8s.

elek: Thanks the comment. I think we can assume that the kernel parameters are adjusted. I will test it without the privileged flag and remove those lines..

PR #493, opened 2019-02-15
HADOOP-16114 Ensure NetUtils#canonicalizeHost returns same canonicalized host name for a given host

Praveen2112:
Currently `NetUtils#canonicalizedHost` returns different canonicalized host name for the same host, this patch resolves it.


steveloughran:
committed to trunk via patch; closing

PR #496, opened 2019-02-18
HDDS-1129. Fix findbug/checkstyle errors in hadoop-hdds projects.

elek:
HDDS-1114 fixed all the findbug/checkstyle problems but in the mean time new patches are committed with newer error.

Here I would like to cleanup the projects again.

(Except the static field in RatisPipelineProvider which will be ignored in this patch and tracked in HDDS-1128)

See: https://issues.apache.org/jira/browse/HDDS-1129


jnp:
+1 for the patch. Please commit if jenkins is satisfactory.

elek:
checkstyle and finfbugs results are full green. Unit test failures are independent and tracked by https://issues.apache.org/jira/browse/HDDS-1127.

elek:
Pushed to the trunk. Thanks the review @jnp

PR #507, opened 2019-02-21
HDDS-1154. Add tracing to the client side of StorageContainerLocationProtocol and OzoneManagerProtocol

elek:
The best way to display the network performance in the tracing server is to start a new span (=save tracing information) on both of the client and and server side.

This issue is about improving the client side tracing of OzoneManager and StorageLocationManager.

The easiest way to turn on tracing for a class which implements an interface is to create a java proxy with TracingUtil.createProxy. With this utility we can execute the required tracing methods (create span/close span) around the original methods without adding any boilerplate code.

Thanks to the current design hadoop rpc calls are create by the *ClientSideTranslatorPB classes which implements the original protocol interface (eg. StorageContainerLocationProtocolClientSideTranslatorPB implements StorageContainerLocationProtocol) which means that it can easily instrumented by TracingUtil.createProxy.

The only thing what we need is to use the interface everywhere (StorageContainerLocationProtocol) instead of the implementation (ClientSideTranslator) as a client.

The only required method which is not published in the ClientSideTranslator is the close method. With adding the close method to the interface (extends Closable) we are able to use the interface everywhere which can be instrumented to send the tracing information.

See: https://issues.apache.org/jira/browse/HDDS-1154

### DIFF
--- a/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/ContainerOperationClient.java
+++ b/hadoop-hdds/client/src/main/java/org/apache/hadoop/hdds/scm/client/ContainerOperationClient.java
@@ -23,8 +23,7 @@ import org.apache.hadoop.hdds.scm.XceiverClientSpi;
 import org.apache.hadoop.hdds.scm.container.common.helpers.ContainerWithPipeline;
 import org.apache.hadoop.hdds.scm.container.ContainerInfo;
 import org.apache.hadoop.hdds.scm.pipeline.Pipeline;
-import org.apache.hadoop.hdds.scm.protocolPB
-    .StorageContainerLocationProtocolClientSideTranslatorPB;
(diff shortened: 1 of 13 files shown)


elek:
How to test it (after a clean build):

```
cd hadoop-ozone/dist/target/ozone-0.4.0-SNAPSHOT/compose/ozonetrace
docker-compose up -d
docker-compose scale datanode=3
#wait a few seconds
docker-compose exec scm ozone freon rk --numOfKeys=10 --numOfBuckets=10 --numOfVolumes=10 --factor=THREE --replicationType=RATIS
```
1. Check the jaeger ui from the browser: http://localhost:16686
2. Choose 'freon' in the 'service' drop down
3. Click to the 'find traces' button
4. Click to a 'creon: createKey' record from the results

After this patch you should see the client/server side of the related calls:

eg. 
 * freon: StorageContainerLocationProtocol.getContainerIdWithPipeline
  * StorageContainerManager.getContainerWithPipeline

PR #589, opened 2019-03-11
HADOOP-16109. Parquet reading S3AFileSystem causes EOF

steveloughran:
Nobody gets seek right. No matter how many times they think they have.

Reproducible test from: Dave Christianson
Fixed seek() logic: Steve Loughran

Change-Id: I39b87f3d5daa98f65de2c0a44e348821a4930573


steveloughran:
tested S3 ireland

PR #590, opened 2019-03-11
HADOOP-16109. Parquet reading S3AFileSystem causes EOF

steveloughran:
Nobody gets seek right. No matter how many times they think they have.

Reproducible test from: Dave Christianson
Fixed seek() logic: Steve Loughran

Change-Id: I268bfcb39ed2626b7281c793d8123f62d8d30caa

### DIFF
--- a/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractSeekTest.java
+++ b/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractSeekTest.java
@@ -272,7 +272,7 @@ public abstract class AbstractContractSeekTest extends AbstractFSContractTestBas
     describe("Seek round a large file and verify the bytes are what is expected");
     Path testSeekFile = path("bigseekfile.txt");
     byte[] block = dataset(100 * 1024, 0, 255);
-    createFile(getFileSystem(), testSeekFile, false, block);
+    createFile(getFileSystem(), testSeekFile, true, block);
     instream = getFileSystem().open(testSeekFile);
     assertEquals(0, instream.getPos());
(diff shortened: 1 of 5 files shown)


steveloughran:
branch-2 variant; hadoop-aws tests all happy exccept for: all the S3n tests which were upset at the lack of a test field and the DDB lifecycle one failing when the container is pay-per-request. Unrelated.

```
 testDynamoDBInitDestroyCycle(org.apache.hadoop.fs.s3a.s3guard.ITestS3GuardToolDynamoDB)  Time elapsed: 18.179 s  <<< ERROR!
com.amazonaws.services.dynamodbv2.model.AmazonDynamoDBException: One or more parameter values were invalid: Neither ReadCapacityUnits nor WriteCapacityUnits can be specified when BillingMode is PAY_PER_REQUEST (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ValidationException; Request ID: J4B2L7CSQAGEF9LD028G4T8ACVVV4KQNSO5AEMVJF66Q9ASUAAJG)
```

Main change from the branch-3.0 patch is the move from lambda to java-7 lang callable & marking more things as final

steveloughran:
Interesting: Yetus isn't seeing these. Any suggestions Mr @aw-was-here ? Some special keyword I can use to kick-off a build?

aw-was-here:
> Yetus isn't seeing these. 

Nope, Jenkins isn't seeing this branch because it doesn't have a Jenkinsfile:

https://builds.apache.org/view/H-L/view/Hadoop/job/hadoop-multibranch/view/change-requests/

Rebase, rebase, rebase.

steveloughran:
this was only a cherry pick, so I fixed by rerunning tests locally till I was happy.

PR #690, opened 2019-04-04
HADOOP-14747: S3AInputStream to implement CanUnbuffer

sahilTakiar:
[HADOOP-14747](https://issues.apache.org/jira/browse/HADOOP-14747): S3AInputStream to implement CanUnbuffer

Change Summary:
* Added a contract specification for `CanUnbuffer` to `fsdatainputstream.md`
    * Unlike most other interfaces the logic for `FSDataInputStream#unbuffer` is a bit different since it delegates to `StreamCapabilitiesPolicy#unbuffer`
    * The one odd thing I noticed while writing up the contract specification is that the current implementation of `unbuffer` allows callers to invoke `unbuffer` on a closed file without issue; I'm not sure if that was by design or not so for now the specification allows for it and dictates that calling `unbuffer` after `close` is a no-op, but let me know if we think the behavior should be changed
* `AbstractContractUnbufferTest` contains the contract tests and I added implementations for HDFS and S3; I don't see contract tests for `CryptoInputStream` so I left it alone
* I added S3 specific tests: both unit tests (using mocks) and itests
* The actual implementation of `unbuffer` in `S3AInputStream` just calls `closeStream`

Testing:
* Ran all unit tests and S3 itests against US East (N. Virginia) using `mvn verify`
    * All tests passed, `ITestS3AContractDistCp` failed once but worked when I retried it
* I didn't run the scale tests, or several of the other tests that require additional configuration; let me know if I should run them

### DIFF
--- a/hadoop-common-project/hadoop-common/src/site/markdown/filesystem/fsdatainputstream.md
+++ b/hadoop-common-project/hadoop-common/src/site/markdown/filesystem/fsdatainputstream.md
@@ -275,6 +275,43 @@ class, which can react to a checksum error in a read by attempting to source
 the data elsewhere. If a new source can be found it attempts to reread and
 recheck that portion of the file.
 
+### `CanUnbuffer.unbuffer()`
+
+This operation instructs the source to release any system resources they are
+currently holding on to, such as buffers, sockets, file descriptors, etc. Any
+subsequent IO operation will likely have to reacquire these resources.
+Unbuffering is useful in situation where streams need to remain open, but no IO
+operation is expected from the stream in the immediate future (examples include
+file handle cacheing).
+
+#### Preconditions
+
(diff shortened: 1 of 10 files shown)


steveloughran:
if people are calling unbuffer on a closed file, then they will only complain if we stop it. How applications actually use HDFS is the defacto FS API. Fun to try and work out what their expectations are though —you normally find out from stack traces.

sahilTakiar:
Ok. In that case, the approach I took in the patch should be good. The contract specifies that calling `unbuffer` after `close` should be a no-op, and I added contract tests that validate `unbuffer` succeeds on closed files.

steveloughran:
reviewed, no issues, liked the docs & new tests. +1,  committed

steveloughran:
merged. Thanks!

PR #715, opened 2019-04-10
HDDS-1370. Command Execution in Datanode fails becaue of NPE

bharatviswa504:
The reason for this can be if we take more time in executing a task, and in taskAwait we will return state as RUNNING and continue, and now when we execute task again internally endPoint state can be shutdown. Below code can return null. So, we need a null check over there.

Callable<EndpointStateMachine.EndPointStates> endpointTask
          = getEndPointTask(endpoint);

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
@@ -86,7 +86,16 @@ public class RunningDatanodeState implements DatanodeState {
     for (EndpointStateMachine endpoint : connectionManager.getValues()) {
       Callable<EndpointStateMachine.EndPointStates> endpointTask
           = getEndPointTask(endpoint);
-      ecs.submit(endpointTask);
+      if (endpointTask != null) {
+        ecs.submit(endpointTask);
+      } else {
+        // This can happen if a task is taking more time than the timeOut
+        // specified for the task in await, and when it is completed the task
+        // has set the state to Shutdown, we may see the state as shutdown
+        // here. So, we need to Shutdown DatanodeStateMachine.
+        LOG.error("State is Shutdown in RunningDatanodeState");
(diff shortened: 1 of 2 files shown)


hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/states/datanode/RunningDatanodeState.java
```diff
@@ -86,7 +86,16 @@ public void execute(ExecutorService executor) {
...
-      ecs.submit(endpointTask);
+      if (endpointTask != null) {
+        ecs.submit(endpointTask);
+      } else {
+        // This can happen if a task is taking more time than the timeOut
+        // specified for the task in await, and when it is completed the task
+        // has set the state to Shutdown, we may see the state as shutdown
+        // here. So, we need to Shutdown DatanodeStateMachine.
+        LOG.error("State is Shutdown in RunningDatanodeState");
+        context.setState(DatanodeStateMachine.DatanodeStates.SHUTDOWN);
```
arp7: Is this line necessary? I don't know this code well so I am not sure.

bharatviswa504: Added this because of the caller checks for datanode state. If endPoint state changed to shutdown, we need to stop this .

Calling code:
 while (context.getState() != DatanodeStates.SHUTDOWN) {
      try {
        LOG.debug("Executing cycle Number : {}", context.getExecutionCount());
        long heartbeatFrequency = context.getHeartbeatFrequency();
        nextHB.set(Time.monotonicNow() + heartbeatFrequency);
        context.execute(executorService, heartbeatFrequency,
            TimeUnit.MILLISECONDS);

arp7: I don't fully understand... unfortunately this stateContext logic is 10x more complex than it needs to be. 

However I think setting the state to SHUTDOWN should be safe so I will +1 the change.

bharatviswa504:
Opened HDDS-1415 for flaky test.
For others, there is already an open Jira.

PR #716, opened 2019-04-10
HADOOP-16205 Backporting ABFS driver from trunk to branch 2.0

kowon2008:


steveloughran:
For anyone reviewing this, this PR is for the review, but the actual final commit will be the full history

hadoop-common-project/hadoop-common/src/site/markdown/FileSystemShell.md
```diff
@@ -415,6 +415,7 @@ Options:
 * -S: Sort output by file size.
 * -r: Reverse the sort order.
 * -u: Use access time rather than modification time for display and sorting.  
+* -e: Display the erasure coding policy of files and directories only.
```
steveloughran: this is from https://issues.apache.org/jira/browse/HDFS-11647 ; unless the whole patch is to be backported, it should be left out

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileStatus.java
```diff
@@ -201,6 +201,15 @@ public FsPermission getPermission() {
     return permission;
   }
 
+  /**
```
steveloughran: comes in as part of HADOOP-14223. Is it needed? As this is such a core class I'm nervous about doing changes here except consistently with branch-3

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/shell/Ls.java
```diff
@@ -245,9 +263,27 @@ protected void processPath(PathData item) throws IOException {
       return;
     }
     FileStatus stat = item.stat;
+
```
steveloughran: this is cherry picking in and then commenting out something which seems unrelated.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractAppendTest.java
```diff
@@ -75,14 +83,11 @@ public void testBuilderAppendToEmptyFile() throws Throwable {
...
     try {
-      FSDataOutputStream out = getFileSystem().append(target);
-      //got here: trouble
-      out.close();
-      fail("expected a failure");
-    } catch (Exception e) {
-      //expected
-      handleExpectedException(e);
+      getFileSystem().append(target).close();
+    } catch (Exception ex) {
```
steveloughran: this removes the entire failure path of the test: The assertion that appending to a nonexistent file raises an error. Why did you make this change?

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractAppendTest.java
```diff
@@ -117,13 +122,9 @@ public void testBuilderAppendToExistingFile() throws Throwable {
...
+      handleExpectedException(ex);
```
steveloughran: again, test is no longer a test of the behavior of any filesystem

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractAppendTest.java
```diff
@@ -149,4 +150,31 @@ public void testRenameFileBeingAppended() throws Throwable {
                                                  dataset.length);
     ContractTestUtils.compareByteArrays(dataset, bytes, dataset.length);
   }
+
```
steveloughran: HADOOP-15744 needs to be applied to this branch; else HDFS tests will break.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractGetFileStatusTest.java
```diff
@@ -272,38 +275,22 @@ public void testListFilesNoDir() throws Throwable {
...
   public void testLocatedStatusNoDir() throws Throwable {
     describe("test the LocatedStatus call on a path which is not present");
-    try {
-      RemoteIterator<LocatedFileStatus> iterator
-          = getFileSystem().listLocatedStatus(path("missing"));
-      fail("Expected an exception, got an iterator: " + iterator);
-    } catch (FileNotFoundException expected) {
-      // expected
-    }
+    getFileSystem().listLocatedStatus(path("missing"));
```
steveloughran: I'd prefer retaining the intercept() code of trunk. I know java 7 hates l-expressions, but we've done the backport elsewhere: see org.apache.hadoop.fs.s3a.s3guard.AbstractS3GuardToolTestBase for some examples. IntellJ IDEA will actually do the conversion from an lambda expression to a Callable for you if you ask it nicely

steveloughran: I'd prefer retaining the intercept() code of trunk. I know java 7 hates l-expressions, but we've done the backport elsewhere: see org.apache.hadoop.fs.s3a.s3guard.AbstractS3GuardToolTestBase for some examples. IntellJ IDEA will actually do the conversion from an lambda expression to a Callable for you if you ask it nicely

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/shell/TestLs.java
```diff
@@ -84,6 +84,7 @@ public void processOptionsNone() throws IOException {
     assertFalse(ls.isOrderSize());
     assertFalse(ls.isOrderTime());
     assertFalse(ls.isUseAtime());
+    assertFalse(ls.isDisplayECPolicy());
```
steveloughran: This is the EC policy patch again. I don't see why abfs needs it at all.

hadoop-tools/hadoop-azure/src/test/java/org/apache/hadoop/fs/azurebfs/ITestAbfsClient.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.rules.ExpectedException;
+
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.azurebfs.contracts.exceptions.AbfsRestOperationException;
+import org.apache.hadoop.fs.azurebfs.services.AbfsClient;
+import org.apache.hadoop.fs.azurebfs.services.AbfsRestOperation;
+
+import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_DEFAULT_NAME_KEY;
+import static org.apache.hadoop.fs.azurebfs.constants.TestConfigurationKeys.FS_AZURE_ACCOUNT_KEY;
+/*
```
steveloughran: remove the commented out import now you've replaced its use in the test case

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/io/TestSequenceFileSerialization.java
```diff
@@ -49,6 +49,7 @@ public void tearDown() throws Exception {
     fs.close();
   }
 
+  @SuppressWarnings("deprecation")
```
steveloughran: I'd leave ths out (or make a PR for trunk on its own and TestSequenceFileSync)

steveloughran:
-1 as is

I'm not worried about the abfs changes itself: that's what the testing is for. I've been more reviewing the changes outside the hadoop-azure module as that's where the risk of side effects is hightest. 

* There's some stuff on Erasure Coding (see Ls) which seems spurious and is mostly commented out anyway.
* `AbstractContractAppendTest` needs HADOOP-15744 applied to it
* The move away from `intercept()` in the tests means that they aren't actually stressing things (e.g AbstractContractConcatTest)

I can see why you've tried to move off intercept(), with java 8 as the target language, but the method can be used with Callable<> operations so does work with java 7. In fact, we started adding it for java 7 [HADOOP-13716](https://issues.apache.org/jira/browse/HADOOP-13716).
If you move back to the intercept calls, you can use intellij to convert the l-expressions to Callables at the click of a mouse (one caveat: needs to think the language is java 8 still, so you have to do it before changing the language version to java 7)

hadoop-common-project/hadoop-common/src/site/markdown/FileSystemShell.md
kowon2008: I found there's a commit https://github.com/apache/hadoop/commit/668817a6cefa6025ddfe082ed71d7d317d811381#diff-eec5ebba9f19087d0813f80afdb11e40 to revert parts of this commit https://github.com/apache/hadoop/commit/fa8cd1bf28f5b81849ba351a2d7225fbc580350d#diff-26c5366750139b7b3a243b5b7a3d67bb. I'll cherry pick this revert.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileStatus.java
kowon2008: This was because I involved changes in hadoop-common between trunk and branch-2 due the reason above. I have redone it and it should be excluded now.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/shell/Ls.java
kowon2008: This was because I involved changes in hadoop-common between trunk and branch-2 due the reason above. I have redone it and it should be excluded now.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractAppendTest.java
kowon2008: This should be fixed now after cleaning the commit and revert part in hadoop-common.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractAppendTest.java
kowon2008: This should be fixed now after cleaning the commit and revert part in hadoop-common.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractAppendTest.java
kowon2008: Yes I have applied to it.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractGetFileStatusTest.java
kowon2008: Yes I have fixed it.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/shell/TestLs.java
kowon2008: This was because I involved changes in hadoop-common between trunk and branch-2 due the reason above. I have redone it and it should be excluded now.

hadoop-tools/hadoop-azure/src/test/java/org/apache/hadoop/fs/azurebfs/ITestAbfsClient.java
kowon2008: This should be fixed now after cleaning the commit and revert part in hadoop-common.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/io/TestSequenceFileSerialization.java
kowon2008: This should be fixed now after cleaning the commit and revert part in hadoop-common.

hadoop-build-tools/src/main/resources/checkstyle/checkstyle.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <!DOCTYPE module PUBLIC
-    "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
-    "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
+    "-//Checkstyle//DTD Checkstyle Configuration 1.2//EN"
+    "https://checkstyle.org/dtds/configuration_1_2.dtd">
```
steveloughran: This is from HADOOP-16232. I've cherry picked that patch on its own, so the changes to the checkstyle DTDs can all be left out of this PR.

steveloughran:
checked out and applied to branch-2 (at commit #c9bfeb225b9); rebuilt, retested

+1 for the patch, once the conflict with commit #313608e4596 is addressed (ie. remove the DTD changes from the checkstyles).

Now, best way to merge this in? We were going to do the full sequence as a merge patch, weren't we? that is: 

1. I locally fork branch-2
1. apply the sequence of commits to it
1. Merge that fork into branch-2 as a merge commit
1. retest, if all good: commit to ASF repo?

Given that this PR is ~this, if you can roll back to commit #e8dfdae066e, apply c3e474fa890 to that and then push it out again, that should be the entire branch history needed to commit

hadoop-build-tools/src/main/resources/checkstyle/checkstyle.xml
kowon2008: Ok. I have rebased and removed this commit.

steveloughran:
Tested w/ Azure cardiff, all well, +1. Committed to branch-2

Thank you for your work!

PR #756, opened 2019-04-19
[HDFS-14437]Fix BUG mentionted in HDFS-14437

AngersZhuuuu:
For the bug of EditLog rolling mentioned in 
https://issues.apache.org/jira/browse/HDFS-10943

I have tell the root cause of it in jira's comment.
https://issues.apache.org/jira/browse/HDFS-14437

In the code of #logSync() this #wait

```
while (mytxid > synctxid && isSyncRunning) {
  try {
    wait(1000);
  } catch (InterruptedException ie) {
  }
}
```
when #endCurrentLogSegment call  #logSync() if  #isSyncRunning == true and mytxid > synctxid,

Current thread  will call #wait, other thread will run.

if other thread can't run , #isSyncRunning will always be true.

current thread can't run out of the while loop

this will become a dead lock.

If other thread get lock to run, They can do many things in 1000ms.

Then  other thread call logSync will end the flush process.

synctxid may be bigger than mytxid, then it will just return in the code :
```
if (mytxid <= synctxid) {
            numTransactionsBatchedInSync++;
            if (metrics != null) {
              // Metrics is non-null only when used inside name node
              metrics.incrTransactionsBatchedInSync();
            }
            return;
          }
```
 When this time you close the JournalSet's OutPutStream, it will trigger the bug.

What I change is to add a control of case of close, always when wait() stop  or been notified by other thread(when other thread finish logSync()), I make mytxid to be the max transaction Id. 
Then this bug will not happen.

 

So , the lock control is not correct.

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSEditLog.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSEditLog.java
@@ -212,6 +212,12 @@ public class FSEditLog implements LogsPurgeable {
     }
   };
 
+  private static final ThreadLocal<Boolean> isEndCurrentLogSegment = new ThreadLocal<Boolean>() {
+    @Override
+    protected synchronized Boolean initialValue() {
+      return false;
+    }
+  };
   /**
    * Constructor for FSEditLog. Underlying journals are constructed, but 
    * no streams are opened until open() is called.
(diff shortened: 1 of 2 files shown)


AngersZhuuuu:
add #waitForSyncToFinish before #endCurrentLogSegment in #rollEditLog  will have the same result as my pull request.

Both work in my test case with high QPS of rollEditLog request.

2000 thread randomly call logEdit and rollEditLog. work fine

AngersZhuuuu:
@arp7 
Hi, can you help me review this ?

PR #794, opened 2019-05-02
HADOOP-16085: use object version or etags to protect against inconsistent read after replace/overwrite

ben-roling:
This is a new PR to replace #675 .  The only difference vs what is on #675 is a merge of the latest code from trunk, resolution of minor merge conflicts when doing so, and squash down to a single commit.

#675 is being closed as-is to preserve the discussion there.

I have performed a full test run on this (against us-west-2 with a bucket with versioning enabled) which passed without any failing tests:

```
mvn -T 1C verify -Dparallel-tests -DtestsThreadCount=8 -Ds3guard -Ddynamo
```
```
Tests run: 860, Failures: 0, Errors: 0, Skipped: 145
```

PR #803, opened 2019-05-08
HADOOP-16085: S3Guard to use object version or etags (interim PR)

steveloughran:
This is #794 with my edits added.


ben-roling:
The changes here looked good to me and I pulled this into #794 as mentioned here:
https://github.com/apache/hadoop/pull/794#issuecomment-490642722

steveloughran:
thanks, I'll move onto that again with my ongoing work.

PR #805, opened 2019-05-09
HDDS-1509. TestBlockOutputStreamWithFailures#test2DatanodesFailure fails intermittently

bshashikant:


hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
```diff
@@ -259,15 +261,23 @@ private void handleException(BlockOutputStreamEntry streamEntry,
...
-        "Encountered exception {}. The last committed block length is {}, "
-            + "uncommitted data length is {} retry count {}", exception,
-        totalSuccessfulFlushedData, bufferedDataLen, retryCount);
+    if (closedContainerException) {
+      LOG.debug(
+          "Encountered exception {}. The last committed block length is {}, "
+              + "uncommitted data length is {} retry count {}", exception,
+          totalSuccessfulFlushedData, bufferedDataLen, retryCount);
+    } else {
+      LOG.info(
```
mukul1987: lets change this to WARN

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
```diff
@@ -259,15 +261,23 @@ private void handleException(BlockOutputStreamEntry streamEntry,
     if (!retryFailure) {
       closedContainerException = checkIfContainerIsClosed(t);
     }
-    PipelineID pipelineId = null;
+    Pipeline pipeline = streamEntry.getPipeline();
+    PipelineID pipelineId = streamEntry.getPipeline().getId();
```
mukul1987: This can be pipeline.getId().

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
bshashikant: Addressed in the latest patch.

hadoop-ozone/client/src/main/java/org/apache/hadoop/ozone/client/io/KeyOutputStream.java
bshashikant: Addressed in the latest patch.

bshashikant:
Thanks @mukul1987 for the review. I have committed this change to trunk.

PR #814, opened 2019-05-13
HDDS-1518. Use /etc/ozone for configuration inside docker-compose 

elek:
See https://issues.apache.org/jira/browse/HDDS-1518 for more details.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN chown hadoop /opt
 ADD scripts /opt/
 ADD scripts/krb5.conf /etc/
 RUN yum install -y krb5-workstation
-RUN mkdir -p /etc/hadoop && mkdir -p /var/log/hadoop && chmod 777 /etc/hadoop && chmod 777 /var/log/hadoop
+RUN mkdir -p /etc/hadoop && mkdir -p /var/log/hadoop && chmod 1777 /etc/hadoop && chmod 1777 /var/log/hadoop
 ENV HADOOP_LOG_DIR=/var/log/hadoop
 ENV HADOOP_CONF_DIR=/etc/hadoop
 WORKDIR /opt/hadoop
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,9 @@ RUN chown hadoop /opt
 ADD scripts /opt/
 ADD scripts/krb5.conf /etc/
 RUN yum install -y krb5-workstation
-
+RUN mkdir -p /etc/hadoop && mkdir -p /var/log/hadoop && chmod 777 /etc/hadoop && chmod 777 /var/log/hadoop
+ENV HADOOP_LOG_DIR=/var/log/hadoop
+ENV HADOOP_CONF_DIR=/etc/hadoop
 WORKDIR /opt/hadoop
 
 VOLUME /data
--- a/scripts/envtoconf.py
+++ b/scripts/envtoconf.py
@@ -36,7 +36,7 @@ class Simple(object):
 
     self.known_formats = ['xml', 'properties', 'yaml', 'yml', 'env', "sh", "cfg", 'conf']
     self.output_dir = self.args.destination
-
+    self.excluded_envs = ['HADOOP_CONF_DIR']
     self.configurables = {}
 
   def destination_file_path(self, name, extension):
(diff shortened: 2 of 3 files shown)


elek:
Note: we create config files based on name convention: FILENAME_EXT_key=value will be converted to FILENAME.EXT. As CONF is a valid extension name (to generate spark/kafka/zookeeper conf, can't remember exactly which one) HADOOP_CONF_DIR should be excluded.

To test:

 * Apply the patch
 * build the conainer with ./build.sh
 * tag it: docker tag apache/hadoop-runner:latest apache/hadoop-runner:jdk11
 * Start any of the compose based environments of ozone

To double check:

```
docker diff ozone_scm_1
C /etc
C /etc/hadoop
A /etc/hadoop/log4j.properties.raw
A /etc/hadoop/log4j2.properties
A /etc/hadoop/log4j2.properties.raw
A /etc/hadoop/ozone-site.xml
A /etc/hadoop/ozone-site.xml.raw
A /etc/hadoop/hdfs-site.xml
A /etc/hadoop/hdfs-site.xml.raw
A /etc/hadoop/log4j.properties
C /opt
A /opt/transformation.pyc
A /data
C /tmp
A /tmp/hsperfdata_hadoop
A /tmp/hsperfdata_hadoop/72
A /tmp/librocksdbjni3381692124588853026.so
A /tmp/hadoop-hadoop-scm.pid
```

As you can see the etc folder is used for configuration.

Dockerfile
```diff
@@ -36,7 +36,9 @@ RUN chown hadoop /opt
 ADD scripts /opt/
 ADD scripts/krb5.conf /etc/
 RUN yum install -y krb5-workstation
-
+RUN mkdir -p /etc/hadoop && mkdir -p /var/log/hadoop && chmod 777 /etc/hadoop && chmod 777 /var/log/hadoop
+ENV HADOOP_LOG_DIR=/var/log/hadoop
+ENV HADOOP_CONF_DIR=/etc/hadoop
```
anuengineer: Did you want to write /etc/ozone here ?

arp7:
+1 LGTM. I see @anuengineer has an open question so I will hold off committing.

Dockerfile
elek: This is a hadoop-runner image I can't write ozone. But we can introduce a separated apache/ozone-runner where we can use ozone names inside. (And we need to fix hadoop-runner anyway)

eyanghwx: We must stop the apache/*-runner idea.  It is a huge time sink for development.  It has also been noted *-runner design broke/by pass Apache voting and release process for Ozone 0.4.0 release.  Let's not go there again.

elek: Thanks @wyanghwx the comment. I  think your opinion is under discussion at https://issues.apache.org/jira/browse/HDDS-1495 and personally I am convinced yet.

But this patch doesn't introduce any new method just fixes a small bug in the existing practice (which is not preferred by you)

Do you prefer to stop all the bugfixing until we got an agreement in HDDS-1495? Do you agree to commit this patch because it's required in the current situation? Do you see any technical problems with applying this patch?


ps: here is a comment: 
https://lists.apache.org/thread.html/e5c17f08f30aa924bdfd50d215b3f468269dc9198411b069d035a891@%3Cgeneral.incubator.apache.org%3E

> The main Docker Hub at hub.docker.com is a public-facing downstream
distribution channel -- similar to Maven Central, PyPI, Debian package
management, etc.

> It is appropriate to distribute official releases through downstream channels,
but inappropriate to distribute unreleased materials through them.  (That's
why having `latest` on hub.docker.com point to git `master` is problematic.)

This is what we are doing right now. We are distributing existing releases via downstream channels.

And finally, I kindly ask to continue the discussion about the current docker practices in HDDS-1495 and don't start the same conversation everywhere. I propose it just because it can be easier to follow and manage. Does it work for you?

anuengineer: +1, Sorry for the confusion. I misunderstood the patch. Let us commit this.

eyanghwx: chmod 777 is a security hole.  It should have sticky bit set (1777) to ensure that normal users can not delete file written by root or other users.

elek: Thanks the comment @eyanghwx. I added the sticky bit in the latest commit.

anuengineer: +1, I will commit this now.

elek:
Thanks @jiwq the review, and also thanks to @arp7 @eyanghwx and @anuengineer for review and commit.

PR #817, opened 2019-05-14
HDFS-14303: branch-2 scan only meta file produce block dir structure check warn fix

iamgd67:
when scan a only meta file but no block file block, will produce a block structure warn eg:

WARN DirectoryScanner:? - Block: 1101939874 has to be upgraded to block ID-based layout. Actual block file path: /data14/hadoop/data/current/BP-1461038173-10.8.48.152-1481686842620/current/finalized/subdir174/subdir68, expected block file path: /data14/hadoop/data/current/BP-1461038173-10.8.48.152-1481686842620/current/finalized/subdir174/subdir68/subdir68

the cause is added a .getParentFile() call mistakenly


iamgd67:
Testing
runed test-patch locally, all pass

iamgd67:
using jar patch instead of pull request

PR #859, opened 2019-05-28
HDDS-1533. JVM exit on TestHddsDatanodeService. Contributed by Mukul Kumar Singh.

mukul1987:
Fix TestHddsDatanodeService exiting because of JVM exit.


elek:
The unit test failures are not releated. I tested the modification with and without the patch (`mvn test -Dtest=TestHddsDatanodeService`) and I can confirm that the problem is fixed.

Will merge it soon.

elek:
Done. Thanks @mukul1987 the fix.

PR #863, opened 2019-05-28
HADOOP-16332. Remove S3A dependency on http core.

steveloughran:
Contributed by Steve Loughran.

Change-Id: I53209c993a405fefdb5e1b692d5a56d027d3b845

### DIFF
--- a/hadoop-tools/hadoop-aws/pom.xml
+++ b/hadoop-tools/hadoop-aws/pom.xml
@@ -407,11 +407,6 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>httpcore</artifactId>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-common</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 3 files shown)


steveloughran:
Tested: S3 Ireland w/ DDB and versioning

steveloughran:
@bgaborg @ben-roling

steveloughran:
FWIW that new test is slow: 
```
Tests run: 135, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 334.725 s - in org.apache.hadoop.fs.s3a.ITestS3ARemoteFileChanged
````
I know its better in parallel, but if there was a way to reduce the option space or test case setup/teardown costs that'd be very beneficial

steveloughran:
full test run completed; one test timeout. Filed [https://issues.apache.org/jira/browse/HADOOP-16333](https://issues.apache.org/jira/browse/HADOOP-16333)

ben-roling:
Btw, you may already know this, but httpcore is currently a transitive, provided scope dependency via hadoop-common.

Whether it is really needed for anything that hadoop-aws does, I don't really know.  There's a good chance it isn't I would guess.

Nonetheless, I support the idea of limiting dependencies and have no objection to this change.  I wasn't sure about it when I added it in the first place and tried to call it out on one of the reviews but I think it slipped through the cracks.

Something else -- org.apache.http.HttpStatus is also included in aws-java-sdk-bundle.  I don't know how you feel about making reference to it given that?  You could potentially delete the httpcore dependency and still use HttpStatus due to that.  It does feel odd to depend on it coming from there though so using an internal constant still feels reasonable.

One last thing - ITestS3AConfiguration also refers to org.apache.http.HttpStatus.  Do you care to remove that reference as well?

steveloughran:
the AWS bundle should only be including that http status as shaded, and so we should treat it as private. And yes, we should cut it there. I'm more worried about production dependencies than test ones though.

I did check through the IDE for httpcore use but could only see it in swift

mackrorysd:
+1 to the idea in general. There is that reference to HttpStatus.SC_MOVED_PERMANENTLY as well - maybe we should define these in Constants?

steveloughran:
I'll take the +1 as clearance to commit. 

If we are to pull these in (and there's a case, given their hard coded use in places), we should put them in `org.apache.hadoop.fs.s3a.InternalConstants`

steveloughran:
merged; closing

PR #896, opened 2019-06-03
HADOOP-16341. ShutDownHookManager: Regressed performance on Hook remo…

t3rmin4t0r:
…vals after HADOOP-15679


steveloughran:
let's see what Yetus says; tentative approval from me

steveloughran:
@t3rmin4t0r can you submit a new PR against trunk so that Yetus will review it. Then we can backport.

PR #997, opened 2019-06-21
[HDFS-14585]Backport HDFS-8901 Use ByteBuffer in DFSInputStream#read to bran…

leosunli:
…ch-2.8

Signed-off-by: sunlisheng <sunlisheng@xiaomi.com>

### DIFF
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DataChecksum.java
@@ -304,7 +304,7 @@ public class DataChecksum implements Checksum {
       }
       return;
     }
-    if (NativeCrc32.isAvailable()) {
+    if (NativeCrc32.isAvailable() && data.isDirect()) {
       NativeCrc32.verifyChunkedSums(bytesPerChecksum, type.id, checksums, data,
           fileName, basePos);
     } else {
--- a/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/DFSInputStream.java
+++ b/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/DFSInputStream.java
@@ -45,7 +45,8 @@ import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
-import com.google.common.base.Preconditions;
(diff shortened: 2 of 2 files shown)


openinx:
@toddlipcon @jojochuang  do you have any bandwidth to review this backport ?

jojochuang:
Copying&Updating my comments from the [jira](https://issues.apache.org/jira/browse/HDFS-14585?focusedCommentId=16869186&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16869186)

Hi Lisheng Sun We typically backport from the newer branches to older branches.
The branch-2.8 is the oldest active branch. Would you please rebase on branch-2?
Branch-2.8 is near EOL. There's no point backporting something from branch-3.x just for branch-2.8

asflicense warning is false alarm. There are minor checkstyle warning that can be cleaned easily. The findbugs warning looks fishy. Looking back in my mailbox, the same warning came out a few times for branch-2. So probably unrelated, but we should file a jira for that.

Suggest to update the jira summary, as this is essentially a reimplementation of HDFS-8901 for replicated blocks. Hadoop 2 does not support EC, and so there's no need to support striped blocks. How about "Reimplement HDFS-8901 for replicated block positional read in branch-2"?
It is confusing if the jira summary is not updated to reflect the fact that striped blocks are not included.

Do we need any test? Or is this purely an performance improvement and existing tests cover it all?

leosunli:
@jojochuang 
  I issue this Jira is that HBase wishes HDFS-3246 is supported in branch-2.8.x.  HDFS-14483 is that backport HDFS-3246 ByteBuffer pread interface to branch-2.8.x.  But HDFS-14483 is blocked by not supporting ByteBuffer in DFSInputStream.  HDFS-8901 support ByteBuffer in DFSInputStream. So I issue this Jira that  only backport part of the conent of HDFS-8901 about byte[] to ByteBuffer. 

And the existing tests cover it all, and I do unit test. 

 Could you suggest this issue to branch-2, branch-2.8 or HDFS-14483 direct support ByteBuffer in DFSInputStream?  I think this issue to branch-2.8 is relatively reasonable and the change is relatively small. Thank you.

openinx:
If branch-2.8 will be EOF (as @jojochuang ),  then  I think we need to backport this patch to branch-2.9 & branch-2, and our HBase need to upgrade from 2.8.5 to 2.9.3 (maybe, or some future hadoop release). 
FYI @leosunli

leosunli:
@jojochuang @openinx  I backport this patch to branch-2.9 & branch-2 Thanks.

jojochuang:
Close. Work is done.

PR #1017, opened 2019-06-26
how to get block in mr

q1007239205:
i want to know block Before  mr task to do something


steveloughran:
ask on a mailing rather than filing PRs. thanks.

PR #1072, opened 2019-07-10
HDDS-1766. ContainerStateMachine is unable to increment lastAppliedTermIndex. Contributed by  Mukul Kumar Singh.

mukul1987:
The last applied term index wasn't updated as the method is called by multiple executors and the method was not synchronized.

### DIFF
--- a/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java
+++ b/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java
@@ -27,6 +27,7 @@ import org.apache.hadoop.hdds.protocol.datanode.proto.ContainerProtos;
 
 import org.apache.hadoop.hdds.scm.container.common.helpers.StorageContainerException;
 import org.apache.hadoop.ozone.container.common.helpers.ContainerUtils;
+import org.apache.hadoop.util.Time;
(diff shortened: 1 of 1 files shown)


bshashikant:
Looks good. Pending test results.

hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/container/common/transport/server/ratis/ContainerStateMachine.java
```diff
@@ -554,12 +556,12 @@ private ByteString getCachedStateMachineData(Long logIndex, long term,
...
     } catch (Exception e) {
       metrics.incNumReadStateMachineFails();
-      LOG.error("unable to read stateMachineData:" + e);
+      LOG.error("{} unable to read stateMachineData:", gid, e);
       return completeExceptionally(e);
     }
   }
 
-  private void updateLastApplied() {
+  private synchronized void updateLastApplied() {
```
lokeshj1703: @mukul1987  Thanks for working on this! The patch looks good to me.
I think we can avoid using synchronized by using a concurrent sorted map. Each applyTransaction removes its corresponding entry from the map and we update lastApplied based on the first entry in the map?

mukul1987: HDDS-1792 will fix this by using ConcurrentHashSet.

lokeshj1703:
@mukul1987 The changes look good to me. +1.

PR #1086, opened 2019-07-12
HADOOP-16341. ShutDownHookManager: Regressed performance on Hook removals after HADOOP-15679

zeroflag:
…vals after HADOOP-15679

Continuation of #940. Taking it over from @t3rmin4t0r.

PR #1109, opened 2019-07-17
HADOOP-16341. ShutDownHookManager: Regressed performance on Hook removals after HADOOP-15679

zeroflag:
clone of #1086 but for 3.1


steveloughran:
applied to branch-3.1, thanks

PR #1138, opened 2019-07-22
HADOOP-16443 Improve help text for setfacl --set option

sodonnel:
The help text associated with the command "setfacl --set" states:

```
--set     Fully replace the ACL, discarding all existing entries. The        
             <acl_spec> must include entries for user, group, and others for    
             compatibility with permission bits.               
```
However the actual behaviour is a bit more subtle:

>    If the ACL spec contains only access entries, then the existing default entries are retained. If the ACL spec contains only default entries, then the existing access entries are retained. If the ACL spec contains both access and default entries, then both are replaced.

This PR will improve the help text to more align with the expected behaviour.


jojochuang:
LGTM. The updated text comes from the javadoc of `AclTransformation#replaceAclEntries()`

steveloughran:
+1, committed to branch 3.2+. Thanks!

PR #1156, opened 2019-07-24
HDDS-1830 OzoneManagerDoubleBuffer#stop should wait for daemon thread to die

smengcl:


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
```diff
@@ -64,7 +65,7 @@
   private final OMMetadataManager omMetadataManager;
   private final AtomicLong flushedTransactionCount = new AtomicLong(0);
   private final AtomicLong flushIterations = new AtomicLong(0);
-  private volatile boolean isRunning;
+  private AtomicBoolean isRunning = new AtomicBoolean(true);
```
arp7: Let's also make it final.

smengcl:
Oops. Thanks for pointing it out @arp7 . Just updated the commit.

hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/ratis/OzoneManagerDoubleBuffer.java
```diff
@@ -64,7 +65,7 @@
   private final OMMetadataManager omMetadataManager;
   private final AtomicLong flushedTransactionCount = new AtomicLong(0);
   private final AtomicLong flushIterations = new AtomicLong(0);
-  private volatile boolean isRunning;
+  private final AtomicBoolean isRunning = new AtomicBoolean(true);
```
bharatviswa504: IsRunning to true should be set only after deamon.start() right?

smengcl: Thanks for pointing this out. Resolved.

arp7:
I merged this before CI had a chance to complete. Sorry about that. Let me revert and retest.

arp7:
I am not sure how revert will work with the PR. Will leave it alone for now. Let's see if subsequent CI runs uncover any issue with this change.

PR #1157, opened 2019-07-24
HADOOP-16461. Regression: FileSystem cache lock parses XML within the lock

t3rmin4t0r:
Instead of parsing a new configuration, rely on existing conf object within FileSystem


steveloughran:
patch LGTM, though Yetus isn't kicking off -presumably due to the ASF lockdown on who can run tests on ASF jenkins infra. Have you set things up so your github UI is tied up to your ASF ID and you are tagged as an ASF committer, so it will trust you?

In the meantime, I'm going to have to close this PR and submit it myself to get it through.

steveloughran:
submitted PR #1161 to see if that will kick off yetus

steveloughran:
Yetus was happy on the other branch (well, apart from no tests)

+1; committed. Thanks for finding this!

PR #1165, opened 2019-07-25
HDDS-1861. Fix TableCacheImpl cleanup logic.

bharatviswa504:


bharatviswa504:
Thank you @arp7 for the review.
Test failures are not related to this patch. I will commit this to trunk.

PR #1267, opened 2019-08-10
HDFS-14665. HttpFS: LISTSTATUS response is missing HDFS-specific fields

smengcl:


jojochuang:
The change is good.
I wish we had a section in the httpfs server doc like  https://hadoop.apache.org/docs/r3.2.0/hadoop-hdfs-httpfs/index.html that documents the differences between WebHDFS and HttpFS. Like the REST APIs unsupported.

smengcl:
> The change is good.
> I wish we had a section in the httpfs server doc like https://hadoop.apache.org/docs/r3.2.0/hadoop-hdfs-httpfs/index.html that documents the differences between WebHDFS and HttpFS. Like the REST APIs unsupported.
> 
> HttpFS in most cases uses HDFS as the backend, but in theory it should support any file system, like S3A, or local file system. So it might be prudent to call out that the output can be different depending on the backend file system used.

Right. HttpFS by default (when fs.defaultFS is not configured) points to LocalFS, where I tested that it won't show those HDFS-specific fields.

The existing doc states 

> it is interoperable with the webhdfs REST HTTP API

It does need some revamp in my opinion.

jojochuang:
Ok. Let's spend some time on the doc, and raise a new jira to update the doc.
+1 for the change in the PR, pending Jenkins.

jojochuang:
The precommit for the forked PR #1289 returned fine.  +1

PR #1291, opened 2019-08-14
HDFS-14665. HttpFS: LISTSTATUS response is missing HDFS-specific fields

smengcl:
Rebased on branch-3.1


jojochuang:
i just realized that git PR doesn't trigger precommit build for branch-3.1. It's only for trunk. See HADOOP-16035 for details.

As far as I can tell this commit compiles. So +1

PR #1350, opened 2019-08-26
YARN-9783. Remove low-level zookeeper test to be able to build Hadoop against zookeeper 3.5.5

symat:
Currently the Hadoop build (with ZooKeeper 3.5.5) fails because of a YARN test case: TestSecureRegistry.testLowlevelZKSaslLogin(). This test case seems to use low-level ZooKeeper internal code, which changed in the new ZooKeeper version.

Removing the testcase will enable us to build and test Hadoop with the latest ZooKeeper version.


aajisaka:
LGTM, +1 pending Jenkins.

aajisaka:
Hi @symat, would you create a pull request for trunk instead of master?

symat:
Sure, I will do that. Also I uploaded a patch file to Jira. Not sure about the correct process in Yarn :)

symat:
The jenkins job for the patch file failed with some unrelated error. We re-triggered that. I will close this PR and wait for the re-triggered jenkins job to succeed.

PR #1355, opened 2019-08-27
HDFS-14770. Erasure Coding: reconstructed internal block is error

runitao:
JIRA: https://issues.apache.org/jira/browse/HDFS-14770


runitao:
not a problem, close it

PR #1356, opened 2019-08-27
HADOOP-16534. Exclude submarine from hadoop source build.

nandakumar131:
When we do source package of hadoop, it should not contain submarine project/code.


nandakumar131:
\cc @sunilgovind

sunilgovind:
@nandakumar131  do u know why jenkins is not running  ?

nandakumar131:
@sunilgovind Jenkins ran for this PR, but the result wasn't published. Not sure why.

PR build: https://builds.apache.org/job/hadoop-multibranch/job/PR-1356/

PR #1376, opened 2019-08-29
HDDS-2058. Remove hadoop dependencies in ozone build.

nandakumar131:


elek:
As we discussed yesterday offline: I will commit this patch (with fixing the rat issue and removing the duplicated log4j). And I will create a separated issue to provide Ozone specific LICENCE file for the source

nandakumar131:
@elek Thanks for the review and commit.

PR #1385, opened 2019-08-30
HDDS-2060. Create Ozone specific LICENSE file for the Ozone source package

elek:
With HDDS-2058 the Ozone (source) release package doesn't contains the hadoop sources any more. We need to create an adjusted LICENSE file for the Ozone source package (We already created a specific LICENSE file for the binary package which is not changed).

In the new LICENSE file we should include entries only for the sources which are part of the Ozone release.

See: https://issues.apache.org/jira/browse/HDDS-2060


anuengineer:
Rat issues?

anuengineer:
Thanks, I have committed to trunk and will cherrypick 0.4.1

PR #970, opened 2019-06-14
HADOOP-16371: Option to disable GCM for SSL connections when running on Java 8

sahilTakiar:
[HADOOP-16371: Option to disable GCM for SSL connections when running on Java 8](https://issues.apache.org/jira/browse/HADOOP-16371)

Changes:
* Patch is based on the merged patch from HADOOP-16050
* Decided a better name for `SSLSocketFactoryEx` would be `DelegatingSSLSocketFactory` because the class is not OpenSSL specific (e.g. it is capable of just delegating to the JSSE)
* Add a bunch of code comments to `DelegatingSSLSocketFactory`
* Documented `fs.s3a.ssl.channel.mode` in `performance.md` and `core-default.xml`
* If a user tries to configure OpenSSL as the mode via `fs.s3a.ssl.channel.mode` an `UnsupportedOperationException` is thrown.

Testing Done:
* Ran all S3 tests `mvn verify` and S3 scale tests `mvn verify -Dparallel-tests -Dscale -DtestsThreadCount=16` (did not have S3Guard or kms tests setup)
* Ran `TestDelegatingSSLSocketFactory` on Ubuntu and OSX with `-Pnative` and confirmed the test passes on both systems (on OSX it is skipped, on Ubuntu it actually runs)
* Ran the ABFS tests against "East US 2" and the only failure was `ITestGetNameSpaceEnabled.testNonXNSAccount` (known issue)
* Ran `mvn package -Pdist -DskipTests -Dmaven.javadoc.skip=true -DskipShade`, un-tarred `hadoop-dist/target/hadoop-3.3.0-SNAPSHOT.tar.gz` and ran `./bin/hadoop fs -ls s3a://[my-bucket-name]/` successfully and that I could upload and read a file to S3 using the CLI without the wildlfy jar on the classpath

PR #1394, opened 2019-09-02
Hadoop-16438 ADLS Gen1 OpenSSL config control

snvijaya:
Enabling config option to control the SSL channel mode. Possible channel modes being -
(a) OpenSSL
(b) Default_JSE
(c) Default - This is the default choice if the config is not present or is invalid. When set to this, connection creation is attempted in OpenSSL mode and will fallback to Default_JSE mode on any failure.


hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
```diff
@@ -19,6 +19,7 @@
 
 package org.apache.hadoop.fs.adl.live;
 
+import com.microsoft.azure.datalake.store.SSLSocketFactoryEx.SSLChannelMode;
```
steveloughran: put a line between this and the org.apache imports

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
```diff
@@ -74,4 +75,67 @@ public void testDefaultTimeout() throws IOException {
...
+  public void testSSLChannelMode() throws IOException {
+    AdlFileSystem fs = null;
+    Configuration conf = null;
+
+    conf = AdlStorageConfiguration.getConfiguration();
+    conf.set(ADL_SSL_CHANNEL_MODE, "OpenSSl");
+    try {
+      fs = (AdlFileSystem)
+          (AdlStorageConfiguration.createStorageConnector(conf));
+    } catch (URISyntaxException e) {
```
steveloughran: just add the URISyntaxException to the list of exceptions this test may throw and remove the try/catch

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
```diff
@@ -74,4 +75,67 @@ public void testDefaultTimeout() throws IOException {
...
+      throw new IllegalStateException("ADL FileSystem initialization failed. "
+          + "Please check test.fs.adl.name property.", e);
+    }
+
+    SSLChannelMode sslChannelMode = fs.getAdlClient().getSSLChannelMode();
+    Assert.assertTrue("Channel mode needs to be OpenSSL",
```
steveloughran: assertEquals for the detailed exception message

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
```diff
@@ -74,4 +75,67 @@ public void testDefaultTimeout() throws IOException {
...
+    try {
+      fs = (AdlFileSystem)
+          (AdlStorageConfiguration.createStorageConnector(conf));
+    } catch (URISyntaxException e) {
+      throw new IllegalStateException("Can not initialize ADL FileSystem. "
+          + "Please check test.fs.adl.name property.", e);
+    }
+
+    sslChannelMode = fs.getAdlClient().getSSLChannelMode();
+    Assert.assertTrue("Channel mode needs to be Default when adl.ssl"
```
steveloughran: assertEquals for the detailed exception message

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
```diff
@@ -74,4 +75,67 @@ public void testDefaultTimeout() throws IOException {
 
     // The default value may vary by SDK, so that value is not tested here.
   }
+
+  @Test
+  public void testSSLChannelMode() throws IOException {
```
steveloughran: this test goes through all the different configurations and then retries the same action through copy-pase (==maintenance cost). Either pull the common code into a method called repeatedly, or use parameterized junit tests

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
```diff
@@ -74,4 +75,67 @@ public void testDefaultTimeout() throws IOException {
...
+    AdlFileSystem fs = null;
+    Configuration conf = null;
+
+    conf = AdlStorageConfiguration.getConfiguration();
+    conf.set(ADL_SSL_CHANNEL_MODE, "OpenSSl");
```
steveloughran: is this intentionally a different case from the rest of the docs? Because I do want the checks to be case insensitive

steveloughran:
1. Are the options case insensitive? If so, can you document this?
2. In "Default", does the user get told that they can't load openssl? As the (reverted) s3a one did this and it just added yet another noisy log message everywhere.

steveloughran:
can you take a look at #970?

assuming that went in, how would your patch link up with it? At the very least, we need consistent option values, defaults, error text, so its easiest to document and use

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
snvijaya: New line added.

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
snvijaya: Have re-formatted the test code to reduce redundant code. Please check.

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
snvijaya: Test code reformatted.

hadoop-tools/hadoop-azure-datalake/src/test/java/org/apache/hadoop/fs/adl/live/TestAdlSdkConfiguration.java
snvijaya: It was intentional to test case insensitive config value. Have highlighted it in reformatted test code.

snvijaya:
> 1. Are the options case insensitive? If so, can you document this?
[Sneha]: Have added the case insensitive option to core-default.xml value description. 
> 2. In "Default", does the user get told that they can't load openssl? As the (reverted) s3a one did this and it just added yet another noisy log message everywhere.
[Sneha]: A log line is printed only if SSLContext with OpenSSL could not be created. 
https://github.com/Azure/azure-data-lake-store-java/blob/master/src/main/java/com/microsoft/azure/datalake/store/SSLSocketFactoryEx.java [line 167]

snvijaya:
> can you take a look at #970?
> 
> assuming that went in, how would your patch link up with it? At the very least, we need consistent option values, defaults, error text, so its easiest to document and use

[Sneha]: Config inputs are inline to the above S3 commit.

steveloughran:
changes in hadoop code are good.

>> In "Default", does the user get told that they can't load openssl? As the (reverted) s3a one did this and it just added yet another noisy log message everywhere.

> A log line is printed only if SSLContext with OpenSSL could not be created.

This means that by default a warning is going to be printed unless openssl is happy. I'm not convinced that is actually useful to most people, and I'm going to make sure that when s3a adds open SSL it isn't going to bother logging at info. However: its in your SDK, so your choice.

steveloughran:
+1, committed to trunk. If you want it in earlier versions just cherrypick, retest and provide a PR which merges. I'm assuming you will want to do this, so I'm leaving it open

PR #1395, opened 2019-09-03
HDDS-1571. Refactor ContainerPlacementPolicy.

timmylicheng:
#HDDS-1571
Refactor ContainerPlacementPolicy to Placement Policy for future exploit.


hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/PlacementPolicy.java
```diff
@@ -15,22 +15,22 @@
...
 
 import org.apache.hadoop.hdds.protocol.DatanodeDetails;
 
 import java.io.IOException;
 import java.util.List;
 
 /**
- * A ContainerPlacementPolicy support choosing datanodes to build replication
+ * A PlacementPolicy support choosing datanodes to build
  * pipeline with specified constraints.
```
ChenSammi: build pipeline or for container replica

ChenSammi:
The path looks overall good. Just one small document issue.

hadoop-hdds/common/src/main/java/org/apache/hadoop/hdds/scm/PlacementPolicy.java
timmylicheng: Updated

ChenSammi:
+1. Committed to branch HDDS-1564.

PR #1398, opened 2019-09-03
HDDS-2064. OzoneManagerRatisServer#newOMRatisServer throws NPE when OM HA is configured incorrectly

smengcl:
The WIP patch can prevent NPE in `TestOzoneManagerConfiguration` unit test `testWrongConfigurationNoOMNodes` and `testWrongConfigurationNoOMAddrs`.

Before the second commit, it fails `testWrongConfiguration` and `testMultipleOMServiceIds`. - The reasons of the failures are that there is another logic checking `isOMAddressSet` right after my patch in `OzoneManager`. And those unit tests are expecting a different exception.

The second commit has addressed the unit test failures. Please review.

### DIFF
--- a/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
+++ b/hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
@@ -615,20 +615,27 @@ public final class OzoneManager extends ServiceRuntimeInfoImpl
             " system with " + OZONE_OM_SERVICE_IDS_KEY + " and " +
             OZONE_OM_ADDRESS_KEY;
         throw new OzoneIllegalArgumentException(msg);
-      } else if (!isOMAddressSet && found == 0) {
-        String msg = "Incorrect configuration. Unable to find OzoneManager" +
-            " node address for service id " + serviceId + ". Please" +
-            " check and reconfigure: " + OZONE_OM_SERVICE_IDS_KEY +
-            ", " + OZONE_OM_NODES_KEY + " and " + OZONE_OM_ADDRESS_KEY;
-        throw new OzoneIllegalArgumentException(msg);
       }
     }
-
+/*
+    if (found == 0) {
+      String msg = "Incorrect configuration. Unable to perform" +
+          " self-discovery for current OzoneManager node. Please" +
+          " check and reconfigure: " + OZONE_OM_SERVICE_IDS_KEY +
+          ", " + OZONE_OM_NODES_KEY + " and " + OZONE_OM_ADDRESS_KEY;
+      throw new OzoneIllegalArgumentException(msg);
+    }
+*/
     if (!isOMAddressSet) {
       // No OM address is set. Fallback to default
       InetSocketAddress omAddress = OmUtils.getOmAddress(conf);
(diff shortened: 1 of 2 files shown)


hadoop-ozone/ozone-manager/src/main/java/org/apache/hadoop/ozone/om/OzoneManager.java
```diff
@@ -614,6 +615,12 @@ private void loadOMHAConfigs(Configuration conf) {
             " system with " + OZONE_OM_SERVICE_IDS_KEY + " and " +
             OZONE_OM_ADDRESS_KEY;
         throw new OzoneIllegalArgumentException(msg);
+      } else if (!isOMAddressSet && found == 0) {
```
bharatviswa504: We cannot add this condition here, because think of case
OZONE_OM_SERVICE_IDS_KEY = ns1,ns2
Because after one iteration of nameserviceID if we are not able to find, we should not say illegal configuration. We should try out all name services and see if it is matching with any nameservice. 

And also, we can eliminate iteration of for loop with OmUtils.emptyAsSingletonNull(omServiceIds))  and similar for nameServiceIds. As for HA this is a must required configuration which needs to be set.

bharatviswa504: And also try out test cases with multiple name services set also. And for this test case, you can use ozone.om.node.id if needed.

smengcl: @bharatviswa504 Thanks for the review. You are right. Now I understand that it makes sense to walk through all configured service ids before declaring failure.

I should probably just put `found == 0` check outside the `serviceId` loop.

elek:
Thank you very much to open this pull request.

During the weekend the Ozone source code has been moved out from [apache/hadoop](https://github.com/apache/hadoop) repository to [apache/hadoop-ozone](https://github.com/apache/hadoop-ozone) repository.

This git commits are rewritten, but the branch of this pull request is also transformed (state of Saturday morning), you can use the new, migrated branch to recreate this pull request.

Your pull request is important for us: **Can you please re-create your pull request in the new repository?**

**1**. Create a new fork of https://github.com/apache/hadoop-ozone

**2**. Clone it and have both your fork and the apache repo as remotes:

```
git clone git@github.com:smengcl/hadoop-ozone.git
cd hadoop-ozone
git remote add apache git@github.com:apache/hadoop-ozone.git
git fetch apache
```

**3**. Fetch your migrated branch and push it to your fork.

```
git checkout -b HDDS-2064 apache/HDDS-2064
git push origin HDDS-2064
```

**4**. And create the new pull request on the new repository.

https://github.com/apache/hadoop-ozone/compare/master...smengcl:HDDS-2064?expand=1

If you need more information, please check [this](https://cwiki.apache.org/confluence/display/HADOOP/Hadoop+Ozone+source+tree+split) wiki page or contact with me (my github user name + apache.org).

Thank you, and sorry for the inconvenience.

smengcl:
Due to the refactoring done in HDDS-2162. This fix has been included in that commit. I will repurpose the jira to add a unit test for the HA config. I'm closing this PR. Will open another one in hadoop-ozone repo.

PR #1388, opened 2019-08-31
HADOOP-16255. Add ChecksumFs.rename(path, path, boolean) to rename crc file as well when FileContext.rename(path, path, options) is called.

HeartSaVioR:
Please refer https://issues.apache.org/jira/browse/HADOOP-16255 for more details.

FYI, `FileContext.rename(path, path, options)` leaks crc file for source of rename when CheckFs or its descendant is used as underlying filesystem. https://issues.apache.org/jira/browse/SPARK-28025 took a workaround via removing crc file manually, and we hope to get rid of workaround eventually.

PR #1413, opened 2019-09-06
HADOOP-16255. Add ChecksumFs.rename(path, path, boolean) to rename crc file as well when FileContext.rename(path, path, options) is called. [BRANCH-2]

HeartSaVioR:
Please refer https://issues.apache.org/jira/browse/HADOOP-16255 for more details.

FYI, FileContext.rename(path, path, options) leaks crc file for source of rename when CheckFs or its descendant is used as underlying filesystem. https://issues.apache.org/jira/browse/SPARK-28025 took a workaround via removing crc file manually, and we hope to get rid of workaround eventually.

This PR is ported version of #1388 for branch-2.

### DIFF
--- a/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/TestChecksumFs.java
+++ b/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/TestChecksumFs.java
@@ -1,0 +1,135 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.fs;
+
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
cc. @steveloughran

HadoopTestBase doesn't seem to be available in branch-2 so I removed it. Except that it was clean cherry-pick. Let's see the build result.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/TestChecksumFs.java
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.hadoop.fs.permission.FsPermission;
+import org.apache.hadoop.test.GenericTestUtils;
+
+import static org.apache.hadoop.fs.CreateFlag.*;
+import static org.junit.Assert.*;
+
+/**
+ * This class tests the functionality of ChecksumFs.
+ */
+public class TestChecksumFs {
```
steveloughran: copy in the TestTimeout rule from the HadoopTestBase class; we need all new tests to timeout before mvn gets bored and kills the process.

steveloughran:
Don't see yetus picking this up; it seems to like trunk the best. Can you attach it as a .patch to the JIRA with a name like HADOOP-16255-branch-2-001.patch; yetus should know to run it against branch-2

HeartSaVioR:
I just subscribed the common-dev@ and recently there is a discussion regarding build - looks like it recognizes JIRA number, so maybe it requires JIRA number to be prefixed (though not sure Yetus recognizes branch). Let me just try to modify title and wait for a hour: if Yetus didn't run I'll add a patch file to JIRA issue.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/TestChecksumFs.java
HeartSaVioR: Ah OK. I thought it would be fine as HadoopTestBase doesn't exist. I'll copy the rule here.

HeartSaVioR:
I just attached the patch file to JIRA issue - looks like Yetus doesn't work for this case. Closing.

PR #1423, opened 2019-09-10
HDDS-2106. Avoid usage of hadoop projects as parent of hdds/ozone

elek:
Ozone uses hadoop as a dependency. The dependency defined on multiple level:

 1. the hadoop artifacts are defined in the <dependency> sections
 2. both hadoop-ozone and hadoop-hdds projects uses "hadoop-project" as the parent

As we already have a slightly different assembly process it could be more resilient to use a dedicated parent project instead of the hadoop one. With this approach it will be easier to upgrade the versions as we don't need to be careful about the pom contents only about the used dependencies.

See: https://issues.apache.org/jira/browse/HDDS-2106


elek:
Thanks @arp7 and @adoroszlai the review. There are no more warnings (missing parts are also migrated from pom.xml) and the integration test failures are not related.

Will merge it soon.

For the records: this is just the first step. As we have this brand new parent pom, later we can simplify the hadoop-ozone/pom.xml and hadoop-hdds/pom.xml as many common parts can be moved to pom.ozone.xml

PR #1428, opened 2019-09-12
HADOOP-16556. Fix some alerts raised by LGTM

malcolmtaylor:
Fix the following alerts:

- `Array index out of bounds` in KerberosName.java

- `Contradictory type checks` in GenericExceptionHandler

- `Missing format argument` in RegistrySecurity and HttpFSExceptionProvider

These alerts are shown in https://lgtm.com/projects/g/apache/hadoop/?mode=tree

### DIFF
--- a/hadoop-common-project/hadoop-auth/src/main/java/org/apache/hadoop/security/authentication/util/KerberosName.java
+++ b/hadoop-common-project/hadoop-auth/src/main/java/org/apache/hadoop/security/authentication/util/KerberosName.java
@@ -281,7 +281,7 @@ public class KerberosName {
         if (paramNum != null) {
           try {
             int num = Integer.parseInt(paramNum);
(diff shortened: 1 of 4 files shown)


hadoop-common-project/hadoop-auth/src/main/java/org/apache/hadoop/security/authentication/util/KerberosName.java
```diff
@@ -281,7 +281,7 @@ static String replaceParameters(String format,
         if (paramNum != null) {
           try {
             int num = Integer.parseInt(paramNum);
-            if (num < 0 || num > params.length) {
+            if (num < 0 || num >= params.length) {
```
steveloughran: nice catch

steveloughran:
+1, merged into trunk. Thanks!

PR #1434, opened 2019-09-12
HDDS-2120. Remove hadoop classes from ozonefs-current jar

elek:
We have two kind of ozone file system jars: current and legacy. current is designed to work only with exactly the same hadoop version which is used for compilation (3.2 as of now).

But as of now the hadoop classes are included in the current jar which is not necessary as the jar is expected to be used in an environment where  the hadoop classes (exactly the same hadoop classes) are already there. They can be excluded.

See: https://issues.apache.org/jira/browse/HDDS-2120

### DIFF
--- a/hadoop-ozone/ozonefs-lib-current/pom.xml
+++ b/hadoop-ozone/ozonefs-lib-current/pom.xml
@@ -91,6 +91,16 @@
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-ozone-filesystem</artifactId>
       <scope>compile</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-common</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-hdfs</artifactId>
+        </exclusion>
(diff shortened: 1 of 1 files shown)


arp7:
Thanks for the quick fix @elek. Should we also exclude Ratis.. since HBase potentially has a dependency on Ratis and we should use the in-build version instead.

What about third party dependencies? Downstream components could run into conflicts.. I believe that will require shading.

elek:
bq. Should we also exclude Ratis.. since HBase potentially has a dependency on Ratis and we should use the in-build version instead.

We need Ratis on the classpath. I am not sure what is the best approach for support HBase. If we exclude it, we need to put two jar files to the classpath of hive/spark: ozonefs and ratis.

Do you know which HBase version uses Ratis?

bq. What about third party dependencies? Downstream components could run into conflicts.. I believe that will require shading

Yes, they can be shaded in HDDS-2121 (at least half of them. Some of them should be kept in original form such as protobuf or the logging).

avijayanhwx:
LGTM +1

elek:
Thanks the review @arp7 and @ajayydv 

The test failures are not related. I am merging it to the trunk right now...

arp7:
It looks like this failed compilation in Jenkins... does the Jenkins job need to be updated to use the separated pom compilation command?

PR #1436, opened 2019-09-12
HDFS-14846: libhdfs tests are failing on trunk due to jni usage bugs

sahilTakiar:
[HDFS-14846](https://issues.apache.org/jira/browse/HDFS-14846): libhdfs tests are failing on trunk due to jni usage bugs

JIRA describes the issue. The fix is to remote any duplicate calls to `DeleteLocalRef`.


anuengineer:
Thank you for the contribution. I have committed this patch to the trunk branch.

PR #1458, opened 2019-09-17
HDDS-2016. Add option to enforce gdpr in Bucket Create command.

dineshchitlangia:
Verified by running in a local cluster as the previous patches for this feature covers the tests.


dineshchitlangia:
failures unrelated to patch

dineshchitlangia:
@elek thanks for review and commit.
Yes, HDDS-2140 will address the test needs.
Regarding the placement of GDPR page, I put it at top level for visibility/promotion/marketing since we didn't have a section for Compliance and I couldn't club this under Security. That said, I am open to moving it around in my next patch. @anuengineer thoughts?

PR #1472, opened 2019-09-18
HDDS-2150. Update dependency versions to avoid security vulnerabilities.

hanishakoneru:
The following dependency versions have known security vulnerabilities. We should update them to recent/ later versions.
- Apache Thrift 0.11.0 (dependency of JaegerTracing)
- Apache Zookeeper 3.4.13
- Jetty Servlet 9.3.24


pom.ozone.xml
```diff
@@ -127,6 +127,9 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     <jackson.version>1.9.13</jackson.version>
     <jackson2.version>2.9.9</jackson2.version>
 
+    <!-- jaegertracing veresion -->
+    <jaeger.version>1.0.0</jaeger.version>
```
adoroszlai: Jaeger 1.0 depends on newer OpenTracing (0.33), which is not backwards compatible.

https://github.com/opentracing/opentracing-java/pull/339
https://github.com/opentracing/opentracing-java#deprecated-members-since-031

`hadoop-hdds-common` compiles only due to explicit dependency on `opentracing-util` 0.31.0.  However, it fails at runtime with [`NoSuchMethodError`](https://github.com/elek/ozone-ci/blob/259712a9df53dd8531786e23676ebed13f527918/pr/pr-hdds-2150-pzdq9/integration/hadoop-ozone/ozonefs/org.apache.hadoop.fs.ozone.contract.ITestOzoneContractDistCp.txt#L6).

For the security fix I think it is enough to upgrade to Jaeger 0.34, which [updated Apache Thrift to 0.12](https://github.com/jaegertracing/jaeger-client-java/blob/136a849202e8d0a95e007e6faae38f1519cdba55/build.gradle#L22).  [Latest Jaeger Client release](https://github.com/jaegertracing/jaeger-client-java/releases/latest) 0.35.2 should be OK, too, as it depends on OpenTracing 0.32, which still has the deprecated methods.  In this case OpenTracing version should be changed to 0.32.0.

hanishakoneru:
Thank you @adoroszlai . I have updated the jaeger tracing version to 0.34.0.
Also removed the zookeeper dependency from ozone. Ozone does not need a direct dependency on zookeeper.

dependencyTree-
```diff
@@ -0,0 +1,4202 @@
...
+[INFO] +- org.apache.hadoop:hadoop-hdds-client:jar:0.5.0-SNAPSHOT:compile
+[INFO] |  \- io.netty:netty-all:jar:4.0.52.Final:compile
+[INFO] +- org.apache.hadoop:hadoop-hdds-server-framework:jar:0.5.0-SNAPSHOT:compile
+[INFO] +- org.apache.hadoop:hadoop-hdds-docs:jar:0.5.0-SNAPSHOT:compile
+[INFO] +- org.apache.hadoop:hadoop-hdds-container-service:test-jar:tests:0.5.0-SNAPSHOT:test
+[INFO] +- org.hamcrest:hamcrest-core:jar:1.3:test
+[INFO] +- io.dropwizard.metrics:metrics-core:jar:3.2.4:compile
+[INFO] |  \- org.slf4j:slf4j-api:jar:1.7.25:compile
+[INFO] +- org.assertj:assertj-core:jar:3.8.0:test
+[INFO] +- org.
```
adoroszlai: I think this file was added accidentally.

PR #1512, opened 2019-09-24
HADOOP-16598. Backport "HADOOP-16558 [COMMON+HDFS] use protobuf-maven…

Apache9:
…-plugin to generate protobuf classes" to all active branches


Apache9:
Oh, seem we do not have a jenkins file on branch-3.2... Let me upload a patch to the jira...

PR #1618, opened 2019-10-08
MAPREDUCE-7240.Exception 'Invalid event: TA_TOO_MANY_FETCH_FAILURE at…

chimney-lee:
… SUCCESS_FINISHING_CONTAINER' cause job error

### DIFF
--- a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
+++ b/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
@@ -1976,7 +1976,7 @@ public abstract class TaskAttemptImpl implements
     @SuppressWarnings("unchecked")
     @Override
     public void transition(TaskAttemptImpl taskAttempt, TaskAttemptEvent event) {
-       if (taskAttempt.getInternalState() ==
+      if (taskAttempt.getInternalState() ==
           TaskAttemptStateInternal.SUCCESS_FINISHING_CONTAINER){
       sendContainerCleanup(taskAttempt, event);
       }
--- a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
+++ b/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
@@ -1976,6 +1976,10 @@ public abstract class TaskAttemptImpl implements
     @SuppressWarnings("unchecked")
     @Override
     public void transition(TaskAttemptImpl taskAttempt, TaskAttemptEvent event) {
+       if (taskAttempt.getInternalState() ==
+          TaskAttemptStateInternal.SUCCESS_FINISHING_CONTAINER){
+      sendContainerCleanup(taskAttempt, event);
+      }
       TaskAttemptTooManyFetchFailureEvent fetchFailureEvent =
           (TaskAttemptTooManyFetchFailureEvent) event;
       // too many fetch failure can only happen for map tasks
--- a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
+++ b/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
@@ -340,12 +340,6 @@ public abstract class TaskAttemptImpl implements
          TaskAttemptStateInternal.SUCCEEDED,
          TaskAttemptEventType.TA_CONTAINER_COMPLETED,
          new ExitFinishingOnContainerCompletedTransition())
-     // Transitions from SUCCESS_FINISHING_CONTAINER state to FAILED stat
-     // When the event TA_TOO_MANY_FETCH_FAILURE received
-      .addTransition(TaskAttemptStateInternal.SUCCESS_FINISHING_CONTAINER,
-         TaskAttemptStateInternal.FAILED,
-         TaskAttemptEventType.TA_TOO_MANY_FETCH_FAILURE,
-         new TooManyFetchFailureTransition())
      // Given TA notifies task T_ATTEMPT_SUCCEEDED when it transitions to
      // SUCCESS_FINISHING_CONTAINER, it is possible to receive the event
      // TA_CONTAINER_CLEANED in the following scenario.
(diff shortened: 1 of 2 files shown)


hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
```diff
@@ -340,6 +340,12 @@
          TaskAttemptStateInternal.SUCCEEDED,
          TaskAttemptEventType.TA_CONTAINER_COMPLETED,
          new ExitFinishingOnContainerCompletedTransition())
+     // Transitions from SUCCESS_FINISHING_CONTAINER state to FAILED stat
+     // When the event TA_TOO_MANY_FETCH_FAILURE received
+      .addTransition(TaskAttemptStateInternal.SUCCESS_FINISHING_CONTAINER,
+         TaskAttemptStateInternal.FAILED,
+         TaskAttemptEventType.TA_TOO_MANY_FETCH_FAILURE,
+         new TooManyFetchFailureTransition())
```
aajisaka: Would you move the lines between
`// Transitions from SUCCESS_FINISHING_CONTAINER state` and  `// Transitions from FAIL_FINISHING_CONTAINER state`?

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/test/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TestTaskAttempt.java
```diff
@@ -1422,6 +1422,29 @@ public void testTimeoutWhileFailFinishing() throws Exception {
...
+        .getJobId(), 1, TaskType.REDUCE);
+    TaskAttemptId reduceTAId =
+        MRBuilderUtils.newTaskAttemptId(reducetaskId, 0);
+
+    taImpl.handle(new TaskAttemptEvent(taImpl.getID(),
+        TaskAttemptEventType.TA_DONE));
+
+    assertEquals("Task attempt's internal state is not " +
+        "SUCCESS_FINISHING_CONTAINER", taImpl.getInternalState(),
+        TaskAttemptStateInternal.SUCCESS_FINISHING_CONTAINER);
```
aajisaka: The order of the arguments is not correct. It should be:
```
    assertEquals("Task attempt's internal state is not " +
        "SUCCESS_FINISHING_CONTAINER",
        TaskAttemptStateInternal.SUCCESS_FINISHING_CONTAINER,
        taImpl.getInternalState());
```

aajisaka:
In addition, would you create a pull request for trunk instead of branch-2.8?

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/test/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TestTaskAttempt.java
```diff
@@ -1422,6 +1422,30 @@ public void testTimeoutWhileFailFinishing() throws Exception {
...
+
+    assertEquals("Task attempt's internal state is not " +
+        "SUCCESS_FINISHING_CONTAINER", 
+        TaskAttemptStateInternal.SUCCESS_FINISHING_CONTAINER,
+        taImpl.getInternalState());
+
+    taImpl.handle(new TaskAttemptTooManyFetchFailureEvent(taImpl.getID(),
+        reduceTAId, "Host"));
+    assertEquals("Task attempt is not in FAILED state", taImpl.getState(),
+        TaskAttemptState.FAILED);
```
aajisaka: Would you fix this line as well?

jlowe:
I think this introduces a new bug.  The code is teleporting straight from `SUCCESS_FINISHING_CONTAINER` to `FAILED` via the existing `TooManyFetchFailureTransition` class.  That existing transition does not expect the container to still be running, so it doesn't handle cleaning up the container.  This scenario needs to ensure the container gets cleaned up as part of the transition, which means there needs to be additional logic beyond just calling that existing transition.

chimney-lee:
> I think this introduces a new bug. The code is teleporting straight from `SUCCESS_FINISHING_CONTAINER` to `FAILED` via the existing `TooManyFetchFailureTransition` class. That existing transition does not expect the container to still be running, so it doesn't handle cleaning up the container. This scenario needs to ensure the container gets cleaned up as part of the transition, which means there needs to be additional logic beyond just calling that existing transition.

thanks for review !  aggree
My plan :
add a new transition called `TooManyFetchFailureTransitionOnSuccessFinishing`   with additional line  
 `sendContainerCleanup(taskAttempt, event);`   cmpare with existing TooManyFetchFailureTransition class.  
Will this be ok ? @aajisaka @jlowe

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
```diff
@@ -377,6 +377,12 @@
              TaskAttemptEventType.TA_COMMIT_PENDING,
              TaskAttemptEventType.TA_FAILMSG,
              TaskAttemptEventType.TA_FAILMSG_BY_CLIENT))
+     // Transitions from SUCCESS_FINISHING_CONTAINER state to FAILED stat
+     // When the event TA_TOO_MANY_FETCH_FAILURE received
```
jlowe: Nit: IMO this comment isn't worth it since it exactly parrots the code.  Also I'd prefer to see this transition appear before the ignored transitions in this state, as the convention has been to put ignored transitions at the end of the list of event handlers for each state.

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
```diff
@@ -1970,6 +1976,10 @@ private void logAttemptFinishedEvent(TaskAttemptStateInternal state) {
     @SuppressWarnings("unchecked")
     @Override
     public void transition(TaskAttemptImpl taskAttempt, TaskAttemptEvent event) {
+      if (taskAttempt.getInternalState() ==
+          TaskAttemptStateInternal.SUCCESS_FINISHING_CONTAINER){
```
jlowe: Whitespace nit.
```suggestion
          TaskAttemptStateInternal.SUCCESS_FINISHING_CONTAINER) {
```

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/main/java/org/apache/hadoop/mapreduce/v2/app/job/impl/TaskAttemptImpl.java
```diff
@@ -1970,6 +1976,10 @@ private void logAttemptFinishedEvent(TaskAttemptStateInternal state) {
...
+      sendContainerCleanup(taskAttempt, event);
```
jlowe: Indentation looks off here.
```suggestion
        sendContainerCleanup(taskAttempt, event);
```

PR #1728, opened 2019-11-26
HADOOP-16728. Avoid removing all the channels from connectionInfo when borrowing from the pool

killerwhile:
Please have a look at this bug. I added a test case to demonstrated the issue. Thanks


lionelfleury:
Any news on this one?

killerwhile:
The problem seems to be addressed in https://issues.apache.org/jira/browse/HADOOP-15358, which has been released in Hadoop 3.3.0.

PR #1754, opened 2019-12-11
HADOOP-16751: DurationInfo text parsing/formatting should be moved out of hotpath

rbalamohan:
DurationInfo is in hotpath in S3A and it would be good to lazy init formatted text. https://issues.apache.org/jira/browse/HADOOP-16751 has the profiler output.

### DIFF
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DurationInfo.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/DurationInfo.java
@@ -18,7 +18,6 @@
 
 package org.apache.hadoop.util;
 
-import com.google.common.base.Suppliers;
 import org.slf4j.Logger;
 
 import org.apache.hadoop.classification.InterfaceAudience.Public;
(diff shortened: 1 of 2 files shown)


hadoop-yetus:
:broken_heart: **-1 overall**






| +0 :ok: |  reexec  |   0m  0s |  Docker mode activated.  |
| -1 :x: |  docker  |   0m 10s |  Docker failed to build yetus/hadoop:104ccca9169.  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| GITHUB PR | https://github.com/apache/hadoop/pull/1754 |
| JIRA Issue | HADOOP-16751 |
| Console output | https://builds.apache.org/job/hadoop-multibranch/job/PR-1754/1/console |
| versions | git=2.17.1 |
| Powered by | Apache Yetus 0.11.1 https://yetus.apache.org |


steveloughran:
well spotted perf issue.

The patch is doing lazy eval, but it now does it in: start, finish and toString calls, so at least 2x current overhed.

better: just have a `String text` field which is set to the formatted result if the output is to be logged which is 

```
(logAtInfo && log.isInfoEnabled() || log.isDebugEnabled()
```

set that in the constructor and reference as needed,

rbalamohan:
Thanks for the review Steve. 

It would be evaluated only once on need basis (textStr takes care of that). 

Having `String text` based on `(logAtInfo && log.isInfoEnabled() || log.isDebugEnabled()`, it would fail in some cases. E.g invoking `toString` on `new DurationInfo(log, false, "test")` would end up with "null" for text.

steveloughran:
I see. and reviewing it, yes, that `getFormattedText()` does update the cached value.

+1

steveloughran:
merged to trunk; thanks

PR #1781, opened 2019-12-24
YARN-10058.Capacity Scheduler dispatcher hang when async thread crash

tuyu:
## NOTICE


TaoYang526:
@tuyu Thanks for the contribution. It is truly harmful when the scheduling thread exits in silent without triggering any state transition. I think it's reasonable to let RM transition to standby when HA is enabled or just shut down (so that the scheduling thread issue can be aware of instead of just missing and the system doesn't work). 
+1 for the patch, would you mind to add UT for this?

shameersss1:
@TaoYang526  - I think the patch is against 3.2.1 which is pretty old. Need to raise patch against lastest trunk

TaoYang526:
@shameersss1 You are right, I have checked this patch is applicable for the latest trunk, didn't notice the merge target is branch-3.2.1.
@tuyu can you raise a patch against trunk?

shameersss1:
@TaoYang526  - I see Tuyu being inactive for past 2.5 years, Not sure if it is taken forward. Since you have spotted this first, Would you mind taking it forward ?

TaoYang526:
@shameersss1  I'll take this forward next week if no response from @tuyu.

PR #1798, opened 2020-01-07
YARN-8374. Upgrade objenesis to 2.6

aajisaka:
JIRA: https://issues.apache.org/jira/browse/YARN-8374


aajisaka:
Thank you, @steveloughran

PR #1812, opened 2020-01-21
HADOOP-13223: remove need for winutils.exe

wintsonpryme:
Please remove or minimize need for native libraries "winutils" to run Spark locally on Windows


steveloughran:
looks like the PR is against the wrong branch -make sure it's generated against trunk

steveloughran:
@wintsonpryme somehow this PR is including everyone else's work.

Please create a PR against trunk, with only your changes. Closing this one

steveloughran:
Reminder: We do want a patch for this, but your code has to be on hadoop trunk, and the PR created against it. Thanks,.

PR #1907, opened 2020-03-23
HADOOP-16854. Fix to prevent OutOfMemoryException and Make the threadpool and bytebuffer pool common across all AbfsOutputStream instances

bilaharith:
HADOOP-16854 Fix to prevent OutOfMemoryException and Make the threadpool and bytebuffer pool common across all AbfsOutputStream instances

### DIFF
--- a/hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AbfsConfiguration.java
+++ b/hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AbfsConfiguration.java
@@ -731,7 +731,7 @@ public class AbfsConfiguration{
   }
 
   @VisibleForTesting
-  void setWriteBufferSize(int bufferSize) {
+  public void setWriteBufferSize(int bufferSize) {
     this.writeBufferSize = bufferSize;
   }
 
--- a/hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AbfsConfiguration.java
+++ b/hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AbfsConfiguration.java
@@ -120,10 +120,20 @@ public class AbfsConfiguration{
       DefaultValue = AZURE_BLOCK_LOCATION_HOST_DEFAULT)
   private String azureBlockLocationHost;
 
-  @IntegerConfigurationValidatorAnnotation(ConfigurationKey = AZURE_CONCURRENT_CONNECTION_VALUE_OUT,
+  @BooleanConfigurationValidatorAnnotation(ConfigurationKey = AZURE_SHOULD_USE_OLD_ABFSOUTPUTSTREAM,
+      DefaultValue = DEFAULT_AZURE_SHOULD_USE_OLD_ABFSOUTPUTSTREAM)
+  private boolean shouldUseOlderAbfsOutputStream;
+
+  @IntegerConfigurationValidatorAnnotation(ConfigurationKey = AZURE_WRITE_CONCURRENCY_FACTOR,
       MinValue = 1,
-      DefaultValue = MAX_CONCURRENT_WRITE_THREADS)
-  private int maxConcurrentWriteThreads;
+  DefaultValue = DEFAULT_WRITE_CONCURRENCY_FACTOR)
+  private int writeConcurrencyFactor;
+
+  @IntegerConfigurationValidatorAnnotation(ConfigurationKey = AZURE_MAX_WRITE_MEM_USAGE_PERCENTAGE,
+      MinValue = MIN_VALUE_MAX_AZURE_WRITE_MEM_USAGE_PERCENTAGE,
+      MaxValue = MAX_VALUE_MAX_AZURE_WRITE_MEM_USAGE_PERCENTAGE,
+      DefaultValue = DEFAULT_MAX_AZURE_WRITE_MEM_USAGE_PERCENTAGE)
+  private int maxWriteMemoryUsagePercentage;
 
   @IntegerConfigurationValidatorAnnotation(ConfigurationKey = AZURE_LIST_MAX_RESULTS,
       MinValue = 1,
@@ -437,12 +447,20 @@ public class AbfsConfiguration{
     return this.azureBlockLocationHost;
   }
 
-  public int getMaxConcurrentWriteThreads() {
-    return this.maxConcurrentWriteThreads;
-  }
-
   public int getMaxConcurrentReadThreads() {
     return this.maxConcurrentReadThreads;
+  }
+
+  public boolean shouldUseOlderAbfsOutputStream() {
+    return this.shouldUseOlderAbfsOutputStream;
+  }
+
+  public int getWriteConcurrencyFactor() {
+    return this.writeConcurrencyFactor;
+  }
+
+  public int getMaxWriteMemoryUsagePercentage() {
+    return this.maxWriteMemoryUsagePercentage;
   }
 
   public int getListMaxResults() {
@@ -730,6 +748,12 @@ public class AbfsConfiguration{
   @VisibleForTesting
   void setListMaxResults(int listMaxResults) {
     this.listMaxResults = listMaxResults;
+  }
+
+  @VisibleForTesting
+  void setShouldUseOlderAbfsOutputStream(
+      boolean shouldUseOlderAbfsOutputStream) {
+    this.shouldUseOlderAbfsOutputStream = shouldUseOlderAbfsOutputStream;
   }
 
   private String getTrimmedPasswordString(String key, String defaultValue) throws IOException {
--- a/hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AbfsConfiguration.java
+++ b/hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AbfsConfiguration.java
@@ -731,7 +731,7 @@ public class AbfsConfiguration{
   }
 
   @VisibleForTesting
-  public void setWriteBufferSize(int bufferSize) {
+  void setWriteBufferSize(int bufferSize) {
     this.writeBufferSize = bufferSize;
   }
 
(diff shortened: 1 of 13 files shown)


hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AbfsConfiguration.java
```diff
@@ -713,7 +731,7 @@ void setReadBufferSize(int bufferSize) {
   }
 
   @VisibleForTesting
-  void setWriteBufferSize(int bufferSize) {
+  public void setWriteBufferSize(int bufferSize) {
```
snvijaya: why public ?

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AzureBlobFileSystemStore.java
```diff
@@ -416,33 +417,39 @@ public OutputStream createFile(final Path path, final boolean overwrite, final F
...
     try (AbfsPerfInfo perfInfo = startTracking("createFile", "createPath")) {
       boolean isNamespaceEnabled = getIsNamespaceEnabled();
-      LOG.debug("createFile filesystem: {} path: {} overwrite: {} permission: {} umask: {} isNamespaceEnabled: {}",
-              client.getFileSystem(),
-              path,
-              overwrite,
-              permission.toString(),
-              umask.toString(),
-              isNamespaceEnabled);
+      LOG.debug(
```
snvijaya: Do not introduce any formatting changes unless needed by actual change.

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AzureBlobFileSystemStore.java
```diff
@@ -416,33 +417,39 @@ public OutputStream createFile(final Path path, final boolean overwrite, final F
...
+            client,
+            AbfsHttpConstants.FORWARD_SLASH + getRelativePath(path),
+            0,
+            abfsConfiguration.getWriteBufferSize(),
+            abfsConfiguration.isFlushEnabled(),
+            abfsConfiguration.isOutputStreamFlushDisabled(),
+            abfsConfiguration.isAppendWithFlushEnabled(),
+            appendBlob);
+      }
+      return new AbfsOutputStream(client,
```
snvijaya: New line needed after a block end

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/constants/ConfigurationKeys.java
```diff
@@ -41,8 +41,11 @@
   public static final String AZURE_WRITE_BUFFER_SIZE = "fs.azure.write.request.size";
   public static final String AZURE_READ_BUFFER_SIZE = "fs.azure.read.request.size";
   public static final String AZURE_BLOCK_SIZE_PROPERTY_NAME = "fs.azure.block.size";
-  public static final String AZURE_BLOCK_LOCATION_HOST_PROPERTY_NAME = "fs.azure.block.location.impersonatedhost";
+  public static final String AZURE_BLOCK_LOCATION_HOST_PROPERTY_NAME = "fs.azure.block.location.impersonatedhost";;
```
snvijaya: Remove delta

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/constants/FileSystemConfigurations.java
```diff
@@ -49,7 +49,11 @@
   public static final int DEFAULT_AZURE_LIST_MAX_RESULTS = 500;
 
   public static final int MAX_CONCURRENT_READ_THREADS = 12;
-  public static final int MAX_CONCURRENT_WRITE_THREADS = 8;
+  public static final int DEFAULT_WRITE_CONCURRENCY_FACTOR = 4;
```
snvijaya: concurrency reduced to 4 ? why.

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/services/AbfsByteBufferPool.java
```diff
@@ -0,0 +1,140 @@
...
+    this.numBuffersInUse = 0;
+    freeBuffers = new ArrayBlockingQueue<>(maxFreeBuffers);
+
+    double maxMemoryAllowedForPool =
+        Runtime.getRuntime().maxMemory() * maxWriteMemUsagePercentage / 100;
+    double bufferCountByMemory = maxMemoryAllowedForPool / bufferSize;
+    double bufferCountByMaxFreeBuffers =
+        maxFreeBuffers + Runtime.getRuntime().availableProcessors();
+
+    maxBuffersInUse = (int) Math
```
snvijaya: if max is determined at constructor, how can the this bufferpool be utilized better memory available increases post this point.

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/services/AbfsByteBufferPool.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  /**
+   * @return byte[] from the pool if available otherwise new byte[] is returned.
+   * Waits if pool is empty and already maximum number of buffers are in use.
+   */
+  public byte[] get() {
+    byte[] byteArray = null;
+    synchronized (this) {
```
snvijaya: instead of locking for counter, why not use AtomicInteger ?

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/services/AbfsOutputStream.java
```diff
@@ -62,33 +62,32 @@
   private long lastFlushOffset;
   private long lastTotalAppendOffset = 0;
 
-  private final int bufferSize;
+  private static int bufferSize;
```
snvijaya: Why should this be static and not final ?

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/services/AbfsOutputStream.java
```diff
@@ -100,19 +99,62 @@ public AbfsOutputStream(
...
+    init(abfsConfiguration);
+    buffer = new byte[bufferSize];
+  }
+
+  private void init(final AbfsConfiguration conf) {
+    if (isCommonPoolsInitialised()) {
+      return;
+    }
+
+    initWriteBufferPool(conf);
```
snvijaya: As discussed, all implementations not tied to the AbfsOutputStream should reside outside this class.

hadoop-tools/hadoop-azure/src/test/java/org/apache/hadoop/fs/azurebfs/ITestAbfsReadWriteAndSeek.java
```diff
@@ -21,6 +21,8 @@
 import java.util.Arrays;
 import java.util.Random;
 
+import org.apache.hadoop.fs.azurebfs.services.AbfsOutputStream;
```
snvijaya: Did you try using profiling tools before and after the change ?

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AbfsConfiguration.java
bilaharith: Changed

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AzureBlobFileSystemStore.java
bilaharith: Done

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/AzureBlobFileSystemStore.java
bilaharith: Done

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/constants/ConfigurationKeys.java
bilaharith: Done

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/constants/FileSystemConfigurations.java
bilaharith: Not changed, Removed an unused constant. This value 4 is currently hardcoded in AbfsOutputStream class. Now made it a config.

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/services/AbfsByteBufferPool.java
bilaharith: There will be synchronisation issues as comparison with the value won't be atomic operations.

hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/services/AbfsOutputStream.java
bilaharith: static final variables cannot be initialised in constructor. And we cannot initialise the same either during declaration or in a static block.

PR #1910, opened 2020-03-24
HADOOP-16930. Add hadoop-aws documentation for ProfileCredentialsProvider

nchammas:
It's very handy, and people are [already using it](https://stackoverflow.com/q/50242843/877069). It would be good to document it.


nchammas:
cc @steveloughran

steveloughran:
+1, committed, thanks

PR #1967, opened 2020-04-20
YARN-9898. Workaround of Netty-all dependency aarch64 support

liusheng:
The YARM CSI client now depend on the netty-all 4.1.48.Final which is
lack of aarch64 support, We have tried to promote aarch64 support in
Netty community, see[1], but still in review progress. For the recent
Hadoop release, it is better to add a workaround for this issue by using
a manully built netty-all package uploaded in org.openlabtesting.netty
maven repo.

[1] https://github.com/netty/netty/pull/9804

PR #1983, opened 2020-04-28
YARN-9898. Workaround of Netty-all dependency aarch64 support

liusheng:
The YARM CSI client now depend on the netty-all 4.1.48.Final which is
lack of aarch64 support, We have tried to promote aarch64 support in
Netty community, see[1], but still in review progress. For the recent
Hadoop release, it is better to add a workaround for this issue by using
a manully built netty-all package uploaded in org.openlabtesting.netty
maven repo.

[1] https://github.com/netty/netty/pull/9804

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs-client/pom.xml
+++ b/hadoop-hdfs-project/hadoop-hdfs-client/pom.xml
@@ -64,7 +64,7 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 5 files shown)


liusheng:
Not sure why the PR https://github.com/apache/hadoop/pull/1967 cannot trigger CI, so here re open an new PR.

johnou:
@liusheng better that you update Netty to 4.1.50 and not use unofficial binaries..

liusheng:
@johnou yes, We have done the 4.1.50 upgrade in Hadoop, thank you :)

PR #2027, opened 2020-05-19
HADOOP-17049. javax.activation-api and jakarta.activation-api define overlapping classes

aajisaka:
JIRA: https://issues.apache.org/jira/browse/HADOOP-17049

Remove javax.activation-api from dependency.


jojochuang:
jakarta.activation-api is the new name for javax.activation-api since 1.2.1, and we include jakarta.activation-api because we updated onto jackson-databind 2.10

PR #2051, opened 2020-06-03
HDFS-15385 Upgrade boost library

GauthamBanasandra:
Hi all,

I've upgraded the boost library to the latest. Please do take a look.
Also, I need some help with the CI. I saw that the whole boost library was checked in as part of the Hadoop codebase. From what I've learned, the third party code must be kept separate or must be pulled at the time of build and should definitely not be part of the codebase.
So, my changes expect that the boost libraries be installed in the system. Could someone with access to the CI machines please install the latest version of boost by running the following command -
```bash
$ sudo apt-get install libboost-all-dev
```

PR #2092, opened 2020-06-24
HDFS-15429. mkdirs should work when parent dir is an internalDir and fallback configured.

umamaheswararao:
https://issues.apache.org/jira/browse/HDFS-15429

### DIFF
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/viewfs/ViewFs.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/viewfs/ViewFs.java
@@ -1134,8 +1134,7 @@ public class ViewFs extends AbstractFileSystem {
 
     @Override
     public void mkdir(final Path dir, final FsPermission permission,
-        final boolean createParent) throws AccessControlException,
-        FileAlreadyExistsException {
+        final boolean createParent) throws IOException {
       if (theInternalDir.isRoot() && dir == null) {
         throw new FileAlreadyExistsException("/ already exits");
       }
@@ -1143,20 +1142,24 @@ public class ViewFs extends AbstractFileSystem {
       if (this.fsState.getRootFallbackLink() != null) {
         AbstractFileSystem linkedFallbackFs =
             this.fsState.getRootFallbackLink().getTargetFileSystem();
-        Path p = Path.getPathWithoutSchemeAndAuthority(
+        Path parent = Path.getPathWithoutSchemeAndAuthority(
             new Path(theInternalDir.fullPath));
-        String child = (InodeTree.SlashPath.equals(dir)) ?
+        String leafChild = (InodeTree.SlashPath.equals(dir)) ?
             InodeTree.SlashPath.toString() :
             dir.getName();
-        Path dirToCreate = new Path(p, child);
+        Path dirToCreate = new Path(parent, leafChild);
         try {
-          linkedFallbackFs.mkdir(dirToCreate, permission, createParent);
+          // We are here because, the parent dir already exist in the mount
+          // table internal tree. So, let's create parent always in fallback.
+          linkedFallbackFs.mkdir(dirToCreate, permission, true);
+          return;
         } catch (IOException e) {
           if (LOG.isDebugEnabled()) {
             StringBuilder msg = new StringBuilder("Failed to create {}")
                 .append(" at fallback fs : {}");
             LOG.debug(msg.toString(), dirToCreate, linkedFallbackFs.getUri());
           }
+          throw e;
         }
       }
 
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/viewfs/ViewFs.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/viewfs/ViewFs.java
@@ -1139,6 +1139,46 @@ public class ViewFs extends AbstractFileSystem {
       if (theInternalDir.isRoot() && dir == null) {
         throw new FileAlreadyExistsException("/ already exits");
       }
+
+      if (this.fsState.getRootFallbackLink() != null) {
+        AbstractFileSystem linkedFallbackFs =
+            this.fsState.getRootFallbackLink().getTargetFileSystem();
+        Path p = Path.getPathWithoutSchemeAndAuthority(
+            new Path(theInternalDir.fullPath));
+        boolean isExist = false;
+        try {
+          isExist = FileContext.getFileContext(linkedFallbackFs, conf).util()
+              .exists(p);
+        } catch (IOException e) {
+          if (LOG.isDebugEnabled()) {
+            StringBuilder msg = new StringBuilder("The parent ")
+                .append(" does not exist at fallback : ");
+            LOG.debug(msg.toString(), p, linkedFallbackFs.getUri());
+          }
+        }
+        String child = (InodeTree.SlashPath.equals(dir)) ?
+            InodeTree.SlashPath.toString() :
+            dir.getName();
+        Path dirToCreate = new Path(p, child);
+        if (isExist) {
+          try {
+            linkedFallbackFs.mkdir(dirToCreate, permission, createParent);
+          } catch (IOException e) {
+            if (LOG.isDebugEnabled()) {
+              StringBuilder msg = new StringBuilder("Failed to create {}")
+                  .append(" at fallback fs : {}");
+              LOG.debug(msg.toString(), p, linkedFallbackFs.getUri());
+            }
+          }
+        } else {
+          StringBuilder msg = new StringBuilder(
+              "Trying to create directory on parent {}  of mount ")
+              .append("link. This mount path does not exit at fallback")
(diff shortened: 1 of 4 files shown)


umamaheswararao:
Thanks a lot @rakeshadr for the discussion offline. I relaxed the restriction on creating parent dirs in the case of fallback does not have the directory structure same as internal mount path. That make sense because anyway user can create that dir tree in fallback when the given directory has two levels deeper from internal mount dir.
I updated the patch and added another test to cover it. Please take a look at it.

umamaheswararao:
For some reason Yetus is failing at last and not posting results.
Here is the link: https://builds.apache.org/job/hadoop-multibranch/job/PR-2092/2/console
No test failures related to this change and no additional checkstyle issues.

```
15:36:33  ============================================================================
15:36:33  ============================================================================
15:36:33                               checkstyle: patch
15:36:33  ============================================================================
15:36:33  ============================================================================
15:36:33  
15:36:33  
15:36:55  cd /home/jenkins/jenkins-slave/workspace/hadoop-multibranch_PR-2092/src
15:36:55  /usr/bin/mvn --batch-mode checkstyle:checkstyle -Dcheckstyle.consoleOutput=true -Ptest-patch -DskipTests -Ptest-patch > /home/jenkins/jenkins-slave/workspace/hadoop-multibranch_PR-2092/out/buildtool-patch-checkstyle-root.txt 2>&1
15:39:31  
15:39:31  root: The patch generated 0 new + 90 unchanged - 1 fixed = 90 total (was 91)

```
https://builds.apache.org/job/hadoop-multibranch/job/PR-2092/2/testReport/

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/viewfs/ViewFs.java
```diff
@@ -1139,6 +1139,27 @@ public void mkdir(final Path dir, final FsPermission permission,
...
+          linkedFallbackFs.mkdir(dirToCreate, permission, createParent);
+        } catch (IOException e) {
+          if (LOG.isDebugEnabled()) {
+            StringBuilder msg = new StringBuilder("Failed to create {}")
+                .append(" at fallback fs : {}");
+            LOG.debug(msg.toString(), dirToCreate, linkedFallbackFs.getUri());
+          }
+        }
+      }
+
```
ayushtkn, line 1162: Thanx @umamaheswararao for the work here,
* Are the changes here in viewFs.java covered anywhere in test? If not will be good to cover this change as well
* Regarding the IOE, we are logging the IOE and then throwing `readOnlyMountTable("mkdir", dir);` instead of the actual exception. Is it intended? It actually suppress the actual reason, There is a difference in behavior between the implementation here and at ViewFileSystem, in case of exception there we handle it and don't throw `readOnlyMountTable("mkdir", dir);` rather a response, though mkdirs and mkdir behave differently and returning false there still might work, but here a person can not fix the issue based on the end exception he receives.

* I think the `createParent` needs to be tackled, when the parent is there in the mount table, in that case we have to explicitly change it to true. I tried tweaking one of your test give a check :
`
 @ Test
  public void testMkdirsOfDeepTreeWithFallbackLinkAndMountPathMatchingDirExist()
      throws Exception {
    Configuration conf = new Configuration();
    conf.setBoolean(Constants.CONFIG_VIEWFS_MOUNT_LINKS_AS_SYMLINKS, false);
    ConfigUtil.addLink(conf, "/user1/hive",
        new Path(targetTestRoot.toString() + "/").toUri());
    Path fallbackTarget = new Path(targetTestRoot, "fallbackDir");
    fsTarget.mkdirs(fallbackTarget);
    ConfigUtil.addLinkFallback(conf, fallbackTarget.toUri());
    AbstractFileSystem vfs = AbstractFileSystem.get(viewFsDefaultClusterUri,
        conf);
      //user1 does not exist in fallback
      Path multipleLevelToInternalDir = new Path("/user1/test");
      Path test = Path.mergePaths(fallbackTarget, multipleLevelToInternalDir);
      assertFalse(fsTarget.exists(test));
      // Creating /user1/test
    // Parent /user1 Exists.
    assertNotNull(vfs.getFileStatus(new Path("/user1")));

    // Creating /user1/test should be success, with createParent false, as
    // parent /user1 exists.
      vfs.mkdir(multipleLevelToInternalDir,
          FsPermission.getDirDefault(),false); // This throws Exception...
      assertTrue(fsTarget.exists(test));

  }
`

umamaheswararao: 1) Yes, I agree somehow we kept Viewfs.java on lowlight, probably due t it's low usage :-) I have added tests.
2) a) This is a good point. I just checked that ViewFs#mkdir is already throwing IOE. I think we can just throw IOE out. So, that users would know what happened.
    b) Another missed was, even in positive case, I was not returning before. Added return after mkdir success.
    c) We always pass createParent as true here. We don't need to actually worry parent exist or not, because parent is exist in mount Internal dir, that's why it got resolved to InternalDirViewFS#mkdir. So, passing createParent as true always in fallback case.
3) Added tests to cover the cases with ViewFs#mkdir with createParentDir true or false, with one level and multipleLevels (means to check recursive creation. Both seems to be working.)
Let's discuss if this behavior make sense to you.

hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/fs/viewfs/TestViewFsLinkFallback.java
```diff
@@ -0,0 +1,298 @@
...
+    assertFalse(fsTarget.exists(test));
+    vfs.mkdir(multipleLevelToInternalDir, null, true);
+    assertTrue(fsTarget.exists(test));
+  }
+
+  /**
+   * Tests that mkdir with createParent false should still create parent in
+   * fallback when the same mount dir exist.
+   */
+  @ Test
```
ayushtkn, line 231: Remove space after @

hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/fs/viewfs/TestViewFileSystemLinkFallback.java
```diff
@@ -606,4 +602,166 @@ public void testLSOnLinkParentWhereMountLinkMatchesWithAFileUnderFallback()
...
+      assertTrue(vfs.mkdirs(multipleLevelToInternalDir));
+      assertTrue(fsTarget.exists(test));
+    }
+  }
+
+  /**
+   * Tests that mkdirs should return false when there is a problem with
+   * fallbackfs.
+   */
+  @Test public void testMkdirsShouldReturnFalseWhenFallbackFSNotAvailable()
```
ayushtkn, line 739: Break line after @Test 

* If inteliiJ -
File -> Settings -> CodeStyle -> Java -> Wrapping and Braces -> Method annotations -> Change to Wrap Always

umamaheswararao:
For some reason its unable to post test results in Github.
No related test failures: https://builds.apache.org/job/hadoop-multibranch/job/PR-2092/3/testReport/

No check style errors:
20:51:12  ============================================================================
20:51:12  ============================================================================
20:51:12                               checkstyle: patch
20:51:12  ============================================================================
20:51:12  ============================================================================
20:51:12  
20:51:12  
20:51:28  cd /home/jenkins/jenkins-slave/workspace/hadoop-multibranch_PR-2092/src
20:51:28  /usr/bin/mvn --batch-mode checkstyle:checkstyle -Dcheckstyle.consoleOutput=true -Ptest-patch -DskipTests -Ptest-patch > /home/jenkins/jenkins-slave/workspace/hadoop-multibranch_PR-2092/out/buildtool-patch-checkstyle-root.txt 2>&1
20:54:04  
20:54:04  root: The patch generated 2 new + 90 unchanged - 1 fixed = 92 total (was 91)

Hey @ayushtkn , do you have some idea whats going on here.
Its actually running all tests and checks but somehow it's failing to post the results.

Could not update commit status, please check if your scan credentials belong to a member of the organization or a collaborator of the repository and repo:status scope is selected

```
GitHub has been notified of this commit’s build result

Timeout has been exceeded
java.lang.InterruptedException
	at java.lang.Object.wait(Native Method)
	at hudson.remoting.Request.call(Request.java:177)
	at hudson.remoting.Channel.call(Channel.java:956)
```

ayushtkn:
Its failing due to timeout. By default the build fails at 5 hours.
HADOOP-17090, is there to track this issue.
If you want to get the Jenkins result, you can temporarily increase the timeout and revert before pushing.
ex.
https://github.com/apache/hadoop/pull/2051/commits/e362cb37aa39f0d0c16dc19b54623e06d2c6b6cc

umamaheswararao:
ok I did not know that. Thanks for letting me know. Since I have test result already (link copied above), I will move ahead to commit. If no objections.

umamaheswararao:
root: The patch generated 1 new + 90 unchanged - 1 fixed = 91 total (was 91)
The above commit fixed the remaining 1. So, it should be 0 new check style. 

Will push it shortly.

PR #2161, opened 2020-07-21
HADOOP-17142. Fix outdated properties of journal node when perform rollback

Deegue:
When rollback HDFS cluster, properties in JNStorage won't be refreshed after the storage dir changed. It leads to exceptions when starting NameNode.

We add refresh method which will refresh properties of JournalNode after rollback has changed the previous dir to current.


Deegue:
Sorry, move it to HDFS, close it.

PR #2182, opened 2020-07-31
HDFS-15505. Fix NullPointerException when call getAdditionalDatanode method with null extendedBlock parameter

hangc0276:
please see  https://issues.apache.org/jira/browse/HDFS-15505

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/protocolPB/ClientNamenodeProtocolTranslatorPB.java
+++ b/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/protocolPB/ClientNamenodeProtocolTranslatorPB.java
@@ -545,16 +545,22 @@ public class ClientNamenodeProtocolTranslatorPB implements
         .newBuilder()
         .setSrc(src)
         .setFileId(fileId)
-        .setBlk(PBHelperClient.convert(blk))
-        .addAllExistings(PBHelperClient.convert(existings))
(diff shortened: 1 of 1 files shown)


hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/protocolPB/ClientNamenodeProtocolTranslatorPB.java
```diff
@@ -545,16 +545,22 @@ public LocatedBlock getAdditionalDatanode(String src, long fileId,
         .newBuilder()
         .setSrc(src)
         .setFileId(fileId)
-        .setBlk(PBHelperClient.convert(blk))
```
goiri, line 548: Can we add a test?

hangc0276: thanks for your feedback, I will add a test soon.

Hexiaoqiao: Thanks @hangc0276, any progress here?

hangc0276: I am adding the test, and will push the code soon.

PR #2332, opened 2020-09-23
HDFS-15594. Lazy calculate live datanodes in safe mode tip

NickyYe:
https://issues.apache.org/jira/browse/HDFS-15594

## NOTICE

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManagerSafeMode.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManagerSafeMode.java
@@ -295,9 +295,11 @@ class BlockManagerSafeMode {
 
   String getSafeModeTip() {
     String msg = "";
+    boolean isBlockThresholdMet = false;
 
     synchronized (this) {
-      if (blockSafe < blockThreshold) {
+      isBlockThresholdMet = (blockSafe >= blockThreshold);
+      if (!isBlockThresholdMet) {
         msg += String.format(
             "The reported blocks %d needs additional %d"
                 + " blocks to reach the threshold %.4f of total blocks %d.%n",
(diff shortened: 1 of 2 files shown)


Hexiaoqiao (approved):
Thanks @NickyYe for your works, +1 from my side.
cc @goiri would you mind to have another check?

goiri (approved):

hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManagerSafeMode.java
```diff
@@ -309,16 +311,21 @@ String getSafeModeTip() {
...
+        } else {
+          msg += String.format("The number of live datanodes %d has reached "
+                  + "the minimum number %d. ",
+              numLive, datanodeThreshold);
+        }
       } else {
-        msg += String.format("The number of live datanodes %d has reached "
-                + "the minimum number %d. ",
-            numLive, datanodeThreshold);
+        msg += "The number of live datanodes is not calculated " +
```
goiri, line 327: As we are at it, does it make sense to use StringBuilder?

NickyYe: fixed.

goiri:
Thanks @NickyYe for the StringBuilder, that will reduce some memory too.
I'm a little surprised that Yetus is not kicking in.
@ayushtkn are you familiar on how to kick it? And now that I have your attention... any issues you can see with changing the message.

ayushtkn (approved):
I have triggered the build -
https://ci-hadoop.apache.org/job/hadoop-multibranch/view/change-requests/job/PR-2332/4/

The improvement seems fair enough, I don't think the message change would create any problem, Changing messages at UI is allowed.
It would change the exception text, I guess when there is a `SafeModeException` through `checkNameNodeSafeMode` in one case, but I don't think that would bother anyone as such. Should be all good.

goiri:
There were a lot of failures because TestFileChecksum:
https://ci-hadoop.apache.org/job/hadoop-multibranch/view/change-requests/job/PR-2332/4/testReport/
But they are not related.

PR #2339, opened 2020-09-25
HDFS-15598: ViewHDFS#canonicalizeUri should not be restricted to DFS only API.

umamaheswararao:
https://issues.apache.org/jira/browse/HDFS-15598


jojochuang (approved):
+1 The asf license error is unrelated. Suspect due to HADOOP-17283. Let me file a jira for that.

umamaheswararao:
Thank you @jojochuang for the review! I have just committed it to trunk.

PR #2341, opened 2020-09-26
HADOOP-17287. Support new Instance by non default constructor by ReflectionUtils

maobaolong:
https://issues.apache.org/jira/browse/HADOOP-17287

This PR provide possibility to new instance by non default constructor through ReflectionUtils

### DIFF
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
@@ -131,7 +131,7 @@ public class ReflectionUtils {
    *
    * @param theClass class of which an object is created
    * @param conf Configuration
-   * @param argsType the types of the arguments
+   * @param argTypes the types of the arguments
    * @param values the values of the arguments
    * @return a new object
    */
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
@@ -137,12 +137,12 @@ public class ReflectionUtils {
    */
   @SuppressWarnings("unchecked")
   public static <T> T newInstance(Class<T> theClass, Configuration conf,
-      Class<?>[] argsType, Object ... values) {
+      Class<?>[] argTypes, Object ... values) {
     T result;
     try {
       Constructor<T> meth = (Constructor<T>) CONSTRUCTOR_CACHE.get(theClass);
       if (meth == null) {
-        meth = theClass.getDeclaredConstructor(argsType);
+        meth = theClass.getDeclaredConstructor(argTypes);
         meth.setAccessible(true);
         CONSTRUCTOR_CACHE.put(theClass, meth);
       }
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
@@ -139,6 +139,12 @@ public class ReflectionUtils {
   public static <T> T newInstance(Class<T> theClass, Configuration conf,
       Class<?>[] argTypes, Object ... values) {
     T result;
+    if (argTypes.length != values.length) {
+      throw new IllegalArgumentException(argTypes.length
+          + " parameters are required but "
+          + values.length
+          + " arguments are provided");
+    }
     try {
       Constructor<T> meth = (Constructor<T>) CONSTRUCTOR_CACHE.get(theClass);
       if (meth == null) {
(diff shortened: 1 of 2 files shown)


hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
```diff
@@ -124,15 +124,29 @@ private static void setJobConf(Object theObject, Configuration conf) {
...
+   *
+   * @param theClass class of which an object is created
+   * @param conf Configuration
+   * @param argsType the types of the arguments
+   * @param values the values of the arguments
+   * @return a new object
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> T newInstance(Class<T> theClass, Configuration conf,
+      Class<?>[] argsType, Object ... values) {
```
sunchao, line 140: nit: `argsType` -> `argTypes`

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
```diff
@@ -124,15 +124,29 @@ private static void setJobConf(Object theObject, Configuration conf) {
...
+  /** Create an object for the given class and initialize it from conf
+   *
+   * @param theClass class of which an object is created
+   * @param conf Configuration
+   * @param argsType the types of the arguments
+   * @param values the values of the arguments
+   * @return a new object
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> T newInstance(Class<T> theClass, Configuration conf,
```
sunchao, line 139: Could you add a test in `TestReflectionUtils`?

sunchao (commented):
Thanks @maobaolong looks good with two nits.

maobaolong:
@sunchao Thanks for your review, I add a new test and fix the typo issues.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
maobaolong: Done

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
maobaolong: Done

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
```diff
@@ -124,15 +124,29 @@ private static void setJobConf(Object theObject, Configuration conf) {
...
+   *
+   * @param theClass class of which an object is created
+   * @param conf Configuration
+   * @param argTypes the types of the arguments
+   * @param values the values of the arguments
+   * @return a new object
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> T newInstance(Class<T> theClass, Configuration conf,
+      Class<?>[] argTypes, Object ... values) {
```
sunchao, line 140: One more thing we can do here is perhaps checking whether `argTypes.length == values.length`, and throw `IllegalArgumentException` like: "1 parameters are required but 0 arguments are provided".

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/util/TestReflectionUtils.java
```diff
@@ -168,6 +168,13 @@ public void testLogThreadInfo() throws Exception {
         containsString("Process Thread Dump: " + title));
   }
 
+  @Test
+  public void testNewInstanceForNonDefaultConstructor() {
+    Object x = ReflectionUtils.newInstance(
+        NoDefaultCtor.class, null, new Class[] {int.class}, 1);
+    Assert.assertTrue(x instanceof NoDefaultCtor);
```
sunchao, line 175: I think `Assert` is not imported and compilation will fail. Also can we test the cases when number of parameter and argument doesn't match?

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ReflectionUtils.java
maobaolong: Good idea, done.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/util/TestReflectionUtils.java
maobaolong: Thanks for check this for me, done

maobaolong:
@sunchao Thanks for the review and point a lost import for me. Now I add an argument num check and a new test case.

And,  I test the two tests which added by me, both are passed.

Please take another look, thank you.

sunchao (approved):
+1 - thanks @maobaolong !

maobaolong:
@sunchao Thanks for your review and merge this PR.

PR #2347, opened 2020-09-29
HDFS-15600 Fix TestRouterQuota#testStorageTypeQuota

huangtianhua:
After "NVDIMM" adding, we have six storage types. This
fixes the related test.


goiri (approved):

ayushtkn (commented):
No failures in the build.
https://ci-hadoop.apache.org/view/Hadoop/job/hadoop-multibranch/view/change-requests/job/PR-2347/1/testReport/
So, in the present scenario, the fix seems correct. I am +1 on it.

Just to confirm @liuml07 should we proceed with fixing the test, are you convinced that things at HDFS-15025 are good?

liuml07:
I have not had a second look at the original patch regarding compatibility, but we can merge this patch first to make test in trunk pass.

If we need to make any further changes to the original NVDIMM feature eg reordering, we can also update this test again.

ayushtkn (approved):

goiri:
@liuml07 OK to merge?

liuml07 (approved):

liuml07:
Thanks for fixing this.

PR #2320, opened 2020-09-21
MAPREDUCE-7282: remove v2 commit algorithm for correctness reasons.

steveloughran:
* All jobs requsting v2 algorithm are WARNED and then switched to v1.
* the v2 codepaths in file output committer have been removed, even
  though some of the method names "rename or merge" are unchanged.

This patch doesn't fix those tests which will fail; I have
those changed and will submit after a Yetus run, to show what breaks/
is fixed

PR #2349, opened 2020-09-29
MAPREDUCE-7282. Move away from V2 commit algorithm

steveloughran:
* Default to v1 commit algorithm
* log at WARN on job set up when v2 is used
* use a special log for that so people who know what they are doing can
  tell it to be quiet

Supercedes #2320

### DIFF
--- a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputCommitter.java
+++ b/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputCommitter.java
@@ -348,14 +348,17 @@ public class FileOutputCommitter extends PathOutputCommitter {
    * @param context the job's context
    */
   public void setupJob(JobContext context) throws IOException {
-    // Downgrade v2 to v1 with a warning.
+    // warning about v2 which is made to a custom logger so people can
+    // turn it off if they are happy with the v2 commit protocol.
     if (algorithmVersion == 2) {
       Logger log = LoggerFactory.getLogger(
           "org.apache.hadoop.mapreduce.lib.output."
               + "FileOutputCommitter.Algorithm");
 
-      log.warn("The v2 commit algorithm is deprecated;"
-          + " please switch to the v1 algorithm");
+      log.warn("The v2 commit algorithm assumes that the content of generated output files is"
+          + " consistent across all task attempts"
+          + " -if this is not true for this job, switch to the v1 commit algorithm."
+          + " See MAPREDUCE-7282");
     }
 
     if (hasOutputPath()) {
--- a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputCommitter.java
+++ b/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputCommitter.java
@@ -68,7 +68,7 @@ public class FileOutputCommitter extends PathOutputCommitter {
       "mapreduce.fileoutputcommitter.marksuccessfuljobs";
   public static final String FILEOUTPUTCOMMITTER_ALGORITHM_VERSION =
       "mapreduce.fileoutputcommitter.algorithm.version";
-  public static final int FILEOUTPUTCOMMITTER_ALGORITHM_VERSION_DEFAULT = 2;
+  public static final int FILEOUTPUTCOMMITTER_ALGORITHM_VERSION_DEFAULT = 1;
   // Skip cleanup _temporary folders under job's output directory
   public static final String FILEOUTPUTCOMMITTER_CLEANUP_SKIPPED =
       "mapreduce.fileoutputcommitter.cleanup.skipped";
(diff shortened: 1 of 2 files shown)


hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputCommitter.java
```diff
@@ -348,6 +348,16 @@ public Path getWorkPath() throws IOException {
    * @param context the job's context
    */
   public void setupJob(JobContext context) throws IOException {
+    // Downgrade v2 to v1 with a warning.
```
jbrennan333, line 351: This comment is no longer accurate.

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputCommitter.java
```diff
@@ -348,6 +348,16 @@ public Path getWorkPath() throws IOException {
...
+    if (algorithmVersion == 2) {
+      Logger log = LoggerFactory.getLogger(
+          "org.apache.hadoop.mapreduce.lib.output."
+              + "FileOutputCommitter.Algorithm");
+
+      log.warn("The v2 commit algorithm is deprecated;"
+          + " please switch to the v1 algorithm");
```
jbrennan333, line 358: I don't think we should use the word deprecated.   That implies that this algorithm will be removed in a future release

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/resources/mapred-default.xml
```diff
@@ -1562,10 +1562,35 @@
...
-  <description>The file output committer algorithm version
-  valid algorithm version number: 1 or 2
-  default to 2, which is the original algorithm
+  <value>1</value>
+  <description>The file output committer algorithm version.
+
+  There are two algorithm versions in Hadoop, "1" and "2".
+
+  The version 2 algorithm is deprecated and no longer the default
+  as task commits were not atomic.
```
jbrennan333, line 1571: Similarly, remove "deprecated".

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/resources/mapred-default.xml
```diff
@@ -1562,10 +1562,35 @@
...
+  That is:
+  1. If a second attempt commits work, there will be no leftover files from
+  a first attempt which failed during its task commit.
+  2. If a network partition causes the first task attempt to overwrite
+  some/all of the output of a second attempt, the result will be
+  exactly the same as if it had not done so.
+
+  To avoid the warning message on job setup, set the log level of the log
+  org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter.Algorithm
+  to ERROR.
```
jbrennan333, line 1593: I think this section should be moved to the end of the Algorithm 2 section below.   You can add (see below for details) to the end of the line that says why algorithm v2 in no longer the default.

jbrennan333 (changes requested):
Thanks for the updated PR @steveloughran!

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputCommitter.java
steveloughran: what do you suggest?

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/resources/mapred-default.xml
steveloughran: ok

steveloughran:
(Yetus failure is from no new tests)

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputCommitter.java
steveloughran: switching to your text

jbrennan333 (commented):
Thanks for the update @steveloughran. The code changes look good to me, but we need to resolve whether the default should be actually be changed - see @daryn-sharp's -1 in the Jira.

jiwq (commented):
LGTM Thanks @steveloughran for the work. But I'm not sure whether we should discuss in the mail list.

steveloughran:
Once the manifest committer #2971 is in, the abfs and gcs stores will get something faster than v1 but with its task failure semantics
* all treewalking for task attempt listing will be in task commit
* no dir renames in task commit, just saving of the manifest
* job commit: parallel load of manifests, merge of list of directories to create, parallel set of mkdirs and then the parallelized renames. 

In this world, we can just leave people using v1/v2 alone, and for (spark) jobs in azure and google cloud say "use the manifest committer". 
Which means I can just close this as a wontfix.

PR #2555, opened 2020-12-17
HDFS-13579. Fix potential OutOfMemory error in DatanodeHttpServer.

aajisaka:
bossGroups.shutdownGracefully() closes its child workerGroups,
so if the workerGroups.shutdownGracefully() is called after
closing bossGroups, it launches the threads again and that may cause
OOM. To fix this issue, close workerGroups before closing bossGroups.

JIRA: https://issues.apache.org/jira/browse/HDFS-13579


aajisaka:
Rethinking this, the OOM is not fixed by this change. Closing.

PR #2710, opened 2021-02-20
HDFS-15843. Make write cross-platform

GauthamBanasandra:
* Currently write from unistd.h is used
  which isn't cross-platform.
* We need to use std::cout.write instead.

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/examples/c/connect_cancel/connect_cancel.c
+++ b/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/examples/c/connect_cancel/connect_cancel.c
@@ -29,6 +29,7 @@
 
 #include "hdfspp/hdfs_ext.h"
 #include "common/util_c.h"
+#include "x-platform/c_api.h"
 
 #define ERROR_BUFFER_SIZE 1024
 
@@ -45,7 +46,7 @@ static void sighandler_direct_stdout(const char *msg) {
   if(!msg) {
     return;
   }
-  printf("%s", msg);
+  x_platform_syscall_write_to_stdout(msg);
 }
 
 static void sig_catch(int val) {
--- a/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/examples/c/connect_cancel/connect_cancel.c
+++ b/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/examples/c/connect_cancel/connect_cancel.c
@@ -26,7 +26,6 @@
 #include <stdlib.h>
 #include <string.h>
 #include <signal.h>
-#include <unistd.h>
 
 #include "hdfspp/hdfs_ext.h"
 #include "common/util_c.h"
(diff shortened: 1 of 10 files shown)


GauthamBanasandra:
@smengcl could you please review this PR?

hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/examples/c/connect_cancel/connect_cancel.c
```diff
@@ -43,10 +42,10 @@ const char *catch_exit   = "Exiting the signal handler.\n";
 // Print to stdout without calling malloc or otherwise indirectly modify userspace state.
 // Write calls to stdout may still interleave with stuff coming from elsewhere.
 static void sighandler_direct_stdout(const char *msg) {
-  if(!msg)
+  if(!msg) {
     return;
-  ssize_t res = write(1 /*posix stdout fd*/, msg, strlen(msg));
-  (void)res;
+  }
+  printf("%s", msg);
```
jojochuang: was this line added to debug and not needed?

GauthamBanasandra: @jojochuang This line is needed. I've replaced `write` call that was writing to **stdout** with `printf`, so that it's cross platform.

smengcl:
Thanks @GauthamBanasandra for the patch.

I'm wondering if the [author](https://issues.apache.org/jira/browse/HDFS-11028) uses syscall write on purpose. According to the comment, heap could be corrupted if the `printf` call uses `malloc` or equivalent (depending on C/C++ library's implementation):

https://github.com/apache/hadoop/blob/8783461e2ec3aff2630ea3574a69beb1d5c61e84/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/examples/c/connect_cancel/connect_cancel.c#L43

https://github.com/apache/hadoop/blob/869317be0a1fdff23be5fc500dcd9ae4ecd7bc29/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/examples/cc/connect_cancel/connect_cancel.cc#L47-L48

How about wrapping the `<unistd.h>` header import line and the existing approach in `#ifndef WINDOWS`? And use `#else` to wrap the new approach. Assuming Windows is the platform you are aiming for.

hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/examples/c/connect_cancel/connect_cancel.c
GauthamBanasandra: I've removed this. Please take a look at the latest commit.

GauthamBanasandra:
Thanks for your review and suggestion @smengcl , @jojochuang. I've re-implemented my solution using the `write` system call for Windows and Linux separately, neatly tucked inside the `XPlatform` class 😁. Please do have another look at my PR.

GauthamBanasandra:
@smengcl could you please review my PR?

goiri (approved):

smengcl (approved):
Thanks @GauthamBanasandra !

PR #2716, opened 2021-02-22
HADOOP-17538. Add kms-default.xml and httpfs-default.xml to site index.

iwasakims:
https://issues.apache.org/jira/browse/HADOOP-17538

While there are links to *-default.xml in the site index, kms-default.xml and httpfs-default.xml are not included. Adding them could be useful as a quick reference.

Since HTMLs are generated by XSL for hadoop-kms and hadoop-hdfs-httpfs on `mvn site`, just adding link to the .html should work.

### DIFF
--- a/hadoop-project/src/site/site.xml
+++ b/hadoop-project/src/site/site.xml
@@ -216,6 +216,8 @@
       <item name="hdfs-rbf-default.xml" href="hadoop-project-dist/hadoop-hdfs-rbf/hdfs-rbf-default.xml"/>
       <item name="mapred-default.xml" href="hadoop-mapreduce-client/hadoop-mapreduce-client-core/mapred-default.xml"/>
       <item name="yarn-default.xml" href="hadoop-yarn/hadoop-yarn-common/yarn-default.xml"/>
+      <item name="kms-default.xml" href="hadoop-kms/kms-default.html"/>
+      <item name="httpfs-default.xml" href="hadoop-hdfs-httpfs/httpfs-default.html"/>
       <item name="Deprecated Properties" href="hadoop-project-dist/hadoop-common/DeprecatedProperties.html"/>
     </menu>
 


ayushtkn (approved):
Makes sense to add the kms & httpfs, once confirmed about the path, I am +1 for this

hadoop-project/src/site/site.xml
```diff
@@ -216,6 +216,8 @@
       <item name="hdfs-rbf-default.xml" href="hadoop-project-dist/hadoop-hdfs-rbf/hdfs-rbf-default.xml"/>
       <item name="mapred-default.xml" href="hadoop-mapreduce-client/hadoop-mapreduce-client-core/mapred-default.xml"/>
       <item name="yarn-default.xml" href="hadoop-yarn/hadoop-yarn-common/yarn-default.xml"/>
+      <item name="kms-default.xml" href="hadoop-kms/kms-default.html"/>
+      <item name="httpfs-default.xml" href="hadoop-hdfs-httpfs/httpfs-default.html"/>
```
ayushtkn, line 220: Just to be sure, where does this path comes from, so as to check if it is actually there. since some are going through hadoop-project-dist/

iwasakims: > since some are going through hadoop-project-dist/

@ayushtkn The placement of docs seems to be decided based on parent pom of the projects. hadoop-common and some hadoop-hdfs projects uses hadoop-project-dist as its parents.
https://github.com/apache/hadoop/blob/rel/release-3.2.2/hadoop-common-project/hadoop-common/pom.xml#L20-L25

The parent of hadoop-kms and hadoop-hdfs-httpfs and many other projects is hadoop-project.
https://github.com/apache/hadoop/blob/rel/release-3.2.2/hadoop-common-project/hadoop-kms/pom.xml#L22-L27
https://github.com/apache/hadoop/blob/rel/release-3.2.2/hadoop-hdfs-project/hadoop-hdfs-httpfs/pom.xml#L22-L27

You can see the distribution by following cmdline on top of source tree.
```
$ find . -name pom.xml | xargs grep -A5 '<parent>'  | less
```

This inconsistency might come from historical reason and could be addressed in follow-up JIRAs.

ayushtkn: Thanx @iwasakims for confirming. you can proceed committing

iwasakims:
Thanks, @ayushtkn. I merged this.

PR #2722, opened 2021-02-24
MAPREDUCE-7320. organize test directories for ClusterMapReduceTestCase

amahussein:
[MAPREDUCE-7320: ClusterMapReduceTestCase does not clean directories](https://issues.apache.org/jira/browse/MAPREDUCE-7320)

Running Junits that extend ClusterMapReduceTestCase generate lots of directories and folders all over the place.

This PR addresses organizing the directories generated by the unit test, cleaning them up at the beginning of the execution if necessary.

- It touches `MiniYARNCluster.java` in order to change the base dir to `target/test-dir/$TEST_CLASS_NAME`
- test classes affected

  - TestMRJobClient,
  - TestStreamingBadRecords,
  - TestClusterMapReduceTestCase,
  - TestBadRecords.
  - TestMRCJCJobClient,
  - TestJobName

I tested the TestUnits that use `MiniYARNCluster.java` such as:

```bash
Class
    MiniYARNCluster

        TestOSSMiniYarnCluster  (3 usages found)
        TestMRTimelineEventHandling  (4 usages found)
        TestJobHistoryEventHandler  (3 usages found)
        TestHadoopArchiveLogs  (3 usages found)
        TestHadoopArchiveLogsRunner  (3 usages found)
        TestDynamometerInfra  (3 usages found)
        TestDSTimelineV10
        TestDSTimelineV20
        TestDSTimelineV15
        TestUnmanagedAMLauncher  (3 usages found)
        TestApplicationMasterServiceProtocolForTimelineV2
        TestFederationRMFailoverProxyProvider  (3 usages found)
        TestHedgingRequestRMFailoverProxyProvider  (4 usages found)
        TestNoHaRMFailoverProxyProvider  (5 usages found)
        TestRMFailover  (4 usages found)
        TestAMRMClient
        TestAMRMClientPlacementConstraints
        TestAMRMProxy  (5 usages found)
        TestNMClient  (3 usages found)
        TestOpportunisticContainerAllocationE2E  (3 usages found)
        TestYarnClient  (3 usages found)
        TestYarnClientWithReservation  (12 usages found)
        TestYarnCLI  (7 usages found)
        TestContainerManagerSecurity  (2 usages found)
        TestDiskFailures  (2 usages found)
        TestMiniYarnCluster  (9 usages found)
        TestMiniYARNClusterForHA  (2 usages found)
        TestMiniYarnClusterNodeUtilization  (3 usages found)
        TestEncryptedShuffle
```

MAPREDUCE-7320. fix TestEncryptedShuffle paths
MAPREDUCE-7320. move cleaning up root directory to setup

### DIFF
--- a/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
+++ b/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
@@ -230,6 +230,31 @@ public abstract class GenericTestUtils {
   }
 
   /**
+   * Creates a directory for the data/logs of the unit test.
+   * It first delete the directory if it exists.
+   *
+   * @param testClass the unit test class.
+   * @param defaultTargetRootDir the directory where the class directory is
+   *                             created.
+   * @return the Path of the root directory.
+   */
+  public static Path setupTestRootDir(Class<?> testClass,
+      String defaultTargetRootDir) {
+    // setup the test root directory
+    String targetTestDir =
+        System.getProperty(SYSPROP_TEST_DATA_DIR, defaultTargetRootDir);
+    Path testRootDirPath =
+        new Path(targetTestDir, testClass.getSimpleName());
+    System.setProperty(GenericTestUtils.SYSPROP_TEST_DATA_DIR,
+        testRootDirPath.toString());
+    // delete the folder if it exists
+    File rootDir = clearTestRootDir();
+    // create the directory
+    rootDir.mkdirs();
+    return testRootDirPath;
+  }
+
+  /**
    * Get the (created) base directory for tests.
    * @return the absolute directory
    */
@@ -243,6 +268,18 @@ public abstract class GenericTestUtils {
     dir.mkdirs();
     assertExists(dir);
     return dir;
+  }
+
+  /**
+   * Cleans-up the root directory from the property
+   * {@link #SYSPROP_TEST_DATA_DIR}.
+   *
+   * @return the absolute file of the test root directory.
+   */
+  public static File clearTestRootDir() {
+    File testRootDir = getTestDir();
+    FileUtil.fullyDelete(testRootDir);
+    return testRootDir;
   }
 
   /**
--- a/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
+++ b/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
@@ -231,27 +231,18 @@ public abstract class GenericTestUtils {
 
   /**
    * Creates a directory for the data/logs of the unit test.
-   * It first delete the directory if it exists.
+   * It first deletes the directory if it exists.
    *
    * @param testClass the unit test class.
-   * @param defaultTargetRootDir the directory where the class directory is
-   *                             created.
    * @return the Path of the root directory.
    */
-  public static Path setupTestRootDir(Class<?> testClass,
-      String defaultTargetRootDir) {
-    // setup the test root directory
-    String targetTestDir =
-        System.getProperty(SYSPROP_TEST_DATA_DIR, defaultTargetRootDir);
-    Path testRootDirPath =
-        new Path(targetTestDir, testClass.getSimpleName());
-    System.setProperty(GenericTestUtils.SYSPROP_TEST_DATA_DIR,
-        testRootDirPath.toString());
-    // delete the folder if it exists
-    File rootDir = clearTestRootDir();
-    // create the directory
-    rootDir.mkdirs();
-    return testRootDirPath;
+  public static File setupTestRootDir(Class<?> testClass) {
+    File testRootDir = getTestDir(testClass.getSimpleName());
+    if (testRootDir.exists()) {
+      FileUtil.fullyDelete(testRootDir);
+    }
+    testRootDir.mkdirs();
+    return testRootDir;
   }
 
   /**
@@ -268,18 +259,6 @@ public abstract class GenericTestUtils {
     dir.mkdirs();
     assertExists(dir);
     return dir;
-  }
-
-  /**
-   * Cleans-up the root directory from the property
-   * {@link #SYSPROP_TEST_DATA_DIR}.
-   *
-   * @return the absolute file of the test root directory.
-   */
-  public static File clearTestRootDir() {
-    File testRootDir = getTestDir();
-    FileUtil.fullyDelete(testRootDir);
-    return testRootDir;
   }
 
   /**
--- a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/src/test/java/org/apache/hadoop/mapred/ClusterMapReduceTestCase.java
+++ b/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/src/test/java/org/apache/hadoop/mapred/ClusterMapReduceTestCase.java
@@ -25,6 +25,7 @@ import org.apache.hadoop.test.GenericTestUtils;
 import org.junit.After;
 import org.junit.Before;
 
+import java.io.File;
 import java.io.IOException;
 import java.util.Map;
 import java.util.Properties;
@@ -45,19 +46,16 @@ import java.util.Properties;
  * The DFS filesystem is formated before the testcase starts and after it ends.
  */
 public abstract class ClusterMapReduceTestCase {
-  private static final String TEST_ROOT_DEFAULT_PATH =
-      System.getProperty("test.build.data", "target/test-dir");
-  private static Path testRootDir;
+  private static File testRootDir;
+  private static File dfsFolder;
 
   private MiniDFSCluster dfsCluster = null;
   private MiniMRClientCluster mrCluster = null;
 
   protected static void setupClassBase(Class<?> testClass) throws Exception {
     // setup the test root directory
-    testRootDir = GenericTestUtils.setupTestRootDir(testClass,
-        TEST_ROOT_DEFAULT_PATH);
-    System.setProperty(GenericTestUtils.SYSPROP_TEST_DATA_DIR,
-        testRootDir.toString());
+    testRootDir = GenericTestUtils.setupTestRootDir(testClass);
+    dfsFolder = new File(testRootDir, "dfs");
   }
 
 
@@ -93,9 +91,9 @@ public abstract class ClusterMapReduceTestCase {
           conf.set((String) entry.getKey(), (String) entry.getValue());
         }
       }
-      dfsCluster = new MiniDFSCluster.Builder(conf).numDataNodes(2)
-      .format(reformatDFS).racks(null).build();
-
+      dfsCluster =
+          new MiniDFSCluster.Builder(conf, dfsFolder)
+              .numDataNodes(2).format(reformatDFS).racks(null).build();
       mrCluster = MiniMRClientClusterFactory.create(this.getClass(), 2, conf);
     }
   }
@@ -151,7 +149,7 @@ public abstract class ClusterMapReduceTestCase {
    * @return path to the root directory for the testcase.
    */
   protected Path getTestRootDir() {
-    return testRootDir;
+    return new Path(testRootDir.getPath());
   }
 
   /**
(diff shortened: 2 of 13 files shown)


hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/src/test/java/org/apache/hadoop/mapred/ClusterMapReduceTestCase.java
```diff
@@ -43,8 +45,21 @@
  * The DFS filesystem is formated before the testcase starts and after it ends.
  */
 public abstract class ClusterMapReduceTestCase {
+  private static final String TEST_ROOT_DEFAULT_PATH =
+      System.getProperty("test.build.data", "target/test-dir");
```
jbrennan333: This seems redundant, since GenericTestUtils.setupTestRootDir() does the same thing.  Can't we just use GenericTestUtils.DEFAULT_TEST_DATA_DIR?  Note that GenericTestUtils.DEFAULT_TEST_DATA_DIR has the advantage of being correct in both linux and windows, where the hard-coded strings will only work in linux.

I'm actually not sure you need to pass a default to GenericTestUtils.setupTestRootDir.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
```diff
@@ -229,6 +229,31 @@ public static int uniqueSequenceId() {
...
+   */
+  public static Path setupTestRootDir(Class<?> testClass,
+      String defaultTargetRootDir) {
+    // setup the test root directory
+    String targetTestDir =
+        System.getProperty(SYSPROP_TEST_DATA_DIR, defaultTargetRootDir);
+    Path testRootDirPath =
+        new Path(targetTestDir, testClass.getSimpleName());
+    System.setProperty(GenericTestUtils.SYSPROP_TEST_DATA_DIR,
+        testRootDirPath.toString());
```
jbrennan333: I'm a little uneasy about overwriting this property.   I think I see why you are doing it, so getTestDir will do the right thing, but I worry that it might lead to unexpected behavior because getTestDir is used everywhere.  At a minimum, I think the comment on the function should clearly state that this changes the property.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
```diff
@@ -229,6 +229,31 @@ public static int uniqueSequenceId() {
     return sequence.incrementAndGet();
   }
 
+  /**
+   * Creates a directory for the data/logs of the unit test.
+   * It first delete the directory if it exists.
+   *
+   * @param testClass the unit test class.
+   * @param defaultTargetRootDir the directory where the class directory is
```
jbrennan333: This description is a little unclear.  Maybe default relative path to use if SYSPROP_TEST_DATA_DIR is not set?

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
```diff
@@ -245,6 +270,18 @@ public static File getTestDir() {
     return dir;
   }
 
+  /**
+   * Cleans-up the root directory from the property
+   * {@link #SYSPROP_TEST_DATA_DIR}.
+   *
+   * @return the absolute file of the test root directory.
+   */
+  public static File clearTestRootDir() {
```
jbrennan333: This is only called from setupTestRootDir(), so I don't think it needs to be a separate public function.  I would just do the fullyDelete() call in-line in setup TestRootDir.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
jbrennan333: I'm actually not sure we need to pass a default here.  Everywhere we use it, it is just using DEFAULT_TEST_DATA_DIR.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
```diff
@@ -229,6 +229,31 @@ public static int uniqueSequenceId() {
...
+   * @param testClass the unit test class.
+   * @param defaultTargetRootDir the directory where the class directory is
+   *                             created.
+   * @return the Path of the root directory.
+   */
+  public static Path setupTestRootDir(Class<?> testClass,
+      String defaultTargetRootDir) {
+    // setup the test root directory
+    String targetTestDir =
+        System.getProperty(SYSPROP_TEST_DATA_DIR, defaultTargetRootDir);
```
jbrennan333: Maybe just use DEFAULT_TEST_DATA_DIR for the default value.   Might also want to add the check for prop.isEmpty() like in getTestDir().

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/src/test/java/org/apache/hadoop/mapred/ClusterMapReduceTestCase.java
```diff
@@ -43,8 +45,21 @@
...
+
   private MiniDFSCluster dfsCluster = null;
-  private MiniMRCluster mrCluster = null;
+  private MiniMRClientCluster mrCluster = null;
+
+  protected static void setupClassBase(Class<?> testClass) throws Exception {
+    // setup the test root directory
+    testRootDir = GenericTestUtils.setupTestRootDir(testClass,
+        TEST_ROOT_DEFAULT_PATH);
+    System.setProperty(GenericTestUtils.SYSPROP_TEST_DATA_DIR,
```
jbrennan333: I don't think we should set the property here if we are doing it in setupTestRootDir().

jbrennan333: One question is whether we should get the original value and then reset it in an afterClass method?  Is every unit test run in a separate jvm instance?   Wouldn't want a following test to get the wrong setting.

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/src/test/java/org/apache/hadoop/mapreduce/security/ssl/TestEncryptedShuffle.java
```diff
@@ -31,37 +30,34 @@
...
-import java.net.URL;
 
 public class TestEncryptedShuffle {
 
-  private static final String BASEDIR =
-    System.getProperty("test.build.dir", "target/test-dir") + "/" +
-    TestEncryptedShuffle.class.getSimpleName();
-  
+  private static final String TEST_ROOT_DEFAULT_PATH =
+      System.getProperty("test.build.data", "target/test-dir");
```
jbrennan333: Similar to above - just use DEFAULT_TEST_DATA_DIR instead of this.  Or remove it if we decide we don't need the default for setupTestRootDir().

jbrennan333 (changes requested):
Thanks @amahussein!  Please see in-line comments.

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/src/test/java/org/apache/hadoop/mapred/ClusterMapReduceTestCase.java
jbrennan333: Ahh. I see it now.  They are different.  But I don't think you need to call getProperty.  Just pass "target/test-dir" as the default for setupTestRootDir.

amahussein: The point that if the property "test.build.data" is set globally to the unit tests, then the `ClusterMapReduceTestCase.java` should also reads this value. Otherwise, all unit tests would not consume the global properties set to the JVM. It will be a mess in the folders

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
amahussein: There has to be a global parameter so that all the modules will use the desired folder. The property is being read everywhere by different modules and this ends up picking different directories.

hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/src/test/java/org/apache/hadoop/mapred/ClusterMapReduceTestCase.java
amahussein: It will be set for the JVM executing the unittest. Each unit test should then have its own property. Then they should not overlap.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/GenericTestUtils.java
amahussein: Also, if I do not set this property, the diff will be very large to get around it.

amahussein:
@jbrennan333 I made some changes to avoid using the global property.
I think this current implementation looks more robust. Later it can be good to apply it to other unit tests.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-tests/src/test/java/org/apache/hadoop/yarn/server/MiniYARNCluster.java
```diff
@@ -20,6 +20,7 @@
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
```
jbrennan333: (nit) this import is no longer needed.

jbrennan333 (changes requested):
Thanks for the updates @amahussein!   This looks much better.
One minor nit, and I will also wait for the precommit build to complete before approving.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-tests/src/test/java/org/apache/hadoop/yarn/server/MiniYARNCluster.java
```diff
@@ -171,8 +173,11 @@ public MiniYARNCluster(
     this.numLocalDirs = numLocalDirs;
     this.numLogDirs = numLogDirs;
     this.enableAHS = enableAHS;
-    String testSubDir = testName.replace("$", "");
-    File targetWorkDir = new File("target", testSubDir);
+    String testSubDir = testName.replace("$", "").concat("-yarn");
```
jbrennan333: I wonder if this testSubDir should just be "yarn".   The targetWorkDirRoot already has the class name in it.

jbrennan333 (commented):
One more minor comment after running some tests.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-tests/src/test/java/org/apache/hadoop/yarn/server/MiniYARNCluster.java
amahussein: I think even better it should be unique folder name "yarn_TESTXYZ" . This way, multiple test cases within the same class won't override the existing folder.

jbrennan333: I was thinking "yarn" because we create a "dfs" directory for hdfs.   We could make random subdirs under it, but I don't think it's necessary.

jbrennan333 (commented):
Thanks for the updates @amahussein!  This looks good to me.
I will wait for pre-commit build to complete before approving.

amahussein:
The build error is misleading once more. It complains for missing symbol but the code is already there.
The patch compiles successfuly. It seems this happens when I force push several commits after rebasing.
We may need to file a separate jira about that.
Thanks @jbrennan333 for the review.

amahussein:
I see a previous Jira on Yetus https://issues.apache.org/jira/browse/YETUS-485
@aajisaka Do you have any thoughts about yetus seeming to fail to pick up all the changes after a rebase?

jbrennan333 (approved):
+1 this looks good to me.  Since we have both built locally and run tests with no problems, I don't think the yetus build failure is reliable.  Hopefully we will get this fixed.

PR #2675, opened 2021-02-02
HADOOP-17511. Add audit/telemetry logging to S3A connector

steveloughran:
Superceded by #2807 
----------

Notion of AuditSpan which is created for a given operation; goal is
to pass it along everywhere.

It's thread local per FS-instance; store operations pick this up in their
constructor from the StoreContext.

The entryPoint() method in S3A FS has been enhanced to initiate the spans.
For this to work, internal code SHALL NOT call those entry points (Done)
and all public API points MUST be declared as entry points. Tricky given
the lack of layering in this branch.

The auditing is intended to be a plugin point; currently there is one
which logs at Info (yes, noisy but good for development)

Change-Id: Idddbbd3138e52b40e97f9c2383453190a02cd8ec

## NOTICE

Please create an issue in ASF JIRA before opening a pull request,
and you need to set the title of the pull request which starts with
the corresponding JIRA issue number. (e.g. HADOOP-XXXXX. Fix a typo in YYY.)
For more details, please see https://cwiki.apache.org/confluence/display/HADOOP/How+To+Contribute

PR #2575, opened 2020-12-29
HADOOP-16080. hadoop-aws does not work with hadoop-client-api

sunchao:
This is a backport of #2522 to trunk

PR #2742, opened 2021-03-04
HADOOP-16721. Improve S3A rename resilience

steveloughran:
S3A Rename to

1. switch from verifying dest parent path is present and a directory to simply verifying that it isn't a file.
2. avoids race condition where thread 1 has deleted the last subdir/file but not yet recreated a parent marker
3. don't downgrade FileNotFoundException to 'false' if source doesn't exist
4. raise FileAlreadyExistsException if dest path exists or parent is a file.

If thread/process 1 deleted the subdir `dest/subdir1` one and there
are no sibling subdirectories, then then the dir `dest` would not
exist until `maybeCreateFakeParentDirectory()` had performed a
`LIST` and, if needed, a `PUT` of a marker.
This creates a window where thread/process 2, trying to rename `staging/subdir2`
into `dest` could fail "parent does not exist".

And guess what:
1. Hive still thinks renaming directories from staging to the production dataset is a good idea.
1. It can spawn many threads to do this parallel (maybe even to compensate for the slow rename())
performance of S3.
1. On a sufficiently overloaded system (lots of threads already doing the parallel filenames)
the window of parent-dir-not-found can last long enough for things to fail.
   
Prior to S3 being consistent this wouldn't have been an issue
* rename()'s need to list everything underneath the source dir wasn't safe to use with delayed list consistency
* so S3Guard was effectively mandatory.
* and delayed LIST consistency may even have reduced the window size.

The fix: go from verifying parent dir exists to simply making sure that it isn't a file
is a weakening of the requirement "parent dir must exist" -but file:// already doesn't
require that. 

Consequences
1. You can rename under a path which doesn't exist. file:// lets you do that already; contract tests have a switch for it.
   (FS contract test has switch for this)
2. Exceptions get raised when false would be returned and hive lose details of what went wrong. This
   isn't consistent with HDFS, but is with other stores (FS contract test has switch for this)
3. You can rename a few layers under a file, but then `create()` lets you do that already,
  and nobody has noticed in production.
   
If directory marker retention was enabled there's more likelihood that an empty dir marker existed -if it did then the race condition wouldn't exist. But as there are no guarantees that the marker will be there, that's not safe to rely on.

### DIFF
--- a/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
+++ b/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
@@ -1143,7 +1143,7 @@ S3 is not a filesystem. The S3A connector mimics file and directory rename by
   Directory markers are not checked for, nor that the path has any children,
 * File-by-file copy of source objects to destination.
   Parallelized, with page listings of directory objects and issuing of DELETE requests.
-* Post-delete recreation of source parent directory marker, if needed.  
+* Post-delete recreation of source parent directory marker, if needed.
 
 This is slow (`O(data)`) and can cause timeouts on code which is required
 to send regular progress reports/heartbeats -for example, distCp.
--- a/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
+++ b/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
@@ -1147,20 +1147,6 @@ of the destination of a rename is a directory -only that it is _not_ a file.
 You can rename a directory or file deep under a file if you try -after which
 there is no guarantee of the files being found in listings. Try not to do that.
 
-For extra debugging of rename issues, it is possible to configure the connector
-to raise exceptions on failures where it would normally just return `false`,
-leaving the application to report an error of the form
-"something failed but we don't know what"
-
-```xml
-<property>
-  <name>fs.s3a.rename.raises.exceptions</name>
-  <value>true</value>
-</property> 
-```
-
-Consider setting this when troubleshooting rename issues. 
-
 ## <a name="encryption"></a> S3 Server Side Encryption
 
 ### `AWSS3IOException` `KMS.NotFoundException` "Invalid arn" when using SSE-KMS
--- a/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
+++ b/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
@@ -1128,21 +1128,35 @@ to S3, but instead write directly to the final destination.
 
 ## Rename not behaving as "expected"
 
-S3 is not a filesystem. The S3A connector mimics rename by
+S3 is not a filesystem. The S3A connector mimics file and directory rename by
 
-* HEAD then LIST of source path
-* HEAD then LIST of destination path
+* HEAD then LIST of source path. The source MUST exist, else a `FileNotFoundException`
+  is raised.
+* HEAD then LIST of the destination path.
+  This SHOULD NOT exist.
+  If it does and if the source is a directory, the destination MUST be an empty directory.
+  If the source is a file, the destination MAY be a directory, empty or not.
+  If the destination exists and relevant conditions are not met, a `FileAlreadyExistsException`
+  is raised.
+* If the destination path does not exist, a HEAD request of the parent path
+  to verify that there is no object there.
+  Directory markers are not checked for, nor that the path has any children,
 * File-by-file copy of source objects to destination.
   Parallelized, with page listings of directory objects and issuing of DELETE requests.
-* Post-delete recreation of destination parent directory marker, if needed.  
+* Post-delete recreation of source parent directory marker, if needed.  
 
-This is slow (O(data)) and unsafe if the calling code expects atomic renaming.
+This is slow (`O(data)`) and can cause timeouts on code which is required
+to send regular progress reports/heartbeats -for example, distCp.
+It is _very unsafe_ if the calling code expects atomic renaming as part
+of any commit algorithm.
 This is why the [S3A Committers](committers.md) or similar are needed to safely
 commit output.
 
 There is also the risk of race conditions arising if many processes/threads
-are working with the same directory tree.
-To reduce this risk, the S3A connector does not verify the parent directory
+are working with the same directory tree
+[HADOOP-16721](https://issues.apache.org/jira/browse/HADOOP-16721).
+
+To reduce this risk, since Hadoop 3.3.1, the S3A connector no longer verifies the parent directory
 of the destination of a rename is a directory -only that it is _not_ a file.
 You can rename a directory or file deep under a file if you try -after which
 there is no guarantee of the files being found in listings. Try not to do that.
--- a/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
+++ b/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
@@ -1126,6 +1126,40 @@ We also recommend using applications/application
 options which do  not rename files when committing work or when copying data
 to S3, but instead write directly to the final destination.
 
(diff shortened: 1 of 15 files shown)


steveloughran:
Testing with  -Dparallel-tests -DtestsThreadCount=8 -Dmarkers=keep -Dscale

If the store is set to raise exceptions, a lot of tests which expect rename(bad options) to return false now get exceptions. One of the contract tests would downgrade if the raised exception was FileAlreadyExistsException and the contract xml said that was ok. I'm reluctant to go with a bigger patch.

this PR is so that Hive and friends can get better reporting on errors, rather than have them lost. It will be optional

steveloughran:
Testing: s3 london with markers==keep and delete, s3guard on and off.

* change in error reporting of s3a FS matched with relevant changes in s3a.xml for contract tests.
* skip tests verifying that you can't rename 2+ levels under a file.
* Failures related to endpoints of common-crawl and  ITestAssumeRole.testAssumeRoleBadInnerAuth: known and fixed in #2675.

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
```diff
@@ -1126,6 +1126,26 @@ We also recommend using applications/application
...
 
+## Rename not behaving as "expected"
+
+S3 is not a filesystem. The S3A connector mimics rename by
+
+* HEAD then LIST of source path
+* HEAD then LIST of destination path
+* File-by-file copy of source objects to destination.
+  Parallelized, with page listings of directory objects and issuing of DELETE requests.
+* Post-delete recreation of destination parent directory marker, if needed.
```
iwasakims: recreation of **source** parent directory marker?

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/impl/ITestlRenameDeleteRace.java
```diff
@@ -0,0 +1,200 @@
...
+import static org.apache.hadoop.fs.s3a.impl.CallableSupplier.waitForCompletion;
+import static org.apache.hadoop.io.IOUtils.cleanupWithLogger;
+
+/**
+ * HADOOP-16721: race condition with delete and rename underneath the same destination
+ * directory.
+ * This test suite recreates the failure using semaphores to guarantee the failure
+ * condition is encountered -then verifies that the rename operation is successful.
+ */
+public class ITestlRenameDeleteRace extends AbstractS3ATestBase {
```
iwasakims: Is the class name intended to be ITestRenameDeleteRace?

steveloughran: yeah. well spotted

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/impl/ITestlRenameDeleteRace.java
```diff
@@ -0,0 +1,200 @@
...
+     * Block for entry into maybeCreateFakeParentDirectory(); will be released
+     * then.
+     */
+    private final Semaphore signalCreatingFakeParentDirectory = new Semaphore(
+        1);
+
+    /**
+     * Semaphore to acquire before the marker can be listed/created.
+     */
+    private final Semaphore blockBeforCreatingMarker = new Semaphore(1);
```
iwasakims: should be blockBefor**e**CreatingMarker.

iwasakims:
LGTM overall. Tests against Tokyo region worked. I just left comments for nits.

```
$ mvn verify -Dtest=x -Dit.test=ITestS3AContractRename,ITestlRenameDeleteRace -Ds3guard
...
[INFO] Running org.apache.hadoop.fs.s3a.impl.ITestlRenameDeleteRace
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4.044 s - in org.apache.hadoop.fs.s3a.impl.ITestlRenameDeleteRace
[INFO] Running org.apache.hadoop.fs.s3a.ITestS3GuardListConsistency
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 35.603 s - in org.apache.hadoop.fs.s3a.ITestS3GuardListConsistency
[INFO] Running org.apache.hadoop.fs.contract.s3a.ITestS3AContractRename
[WARNING] Tests run: 22, Failures: 0, Errors: 0, Skipped: 2, Time elapsed: 12.523 s - in org.apache.hadoop.fs.contract.s3a.ITestS3AContractRename
```

iwasakims:
I got error on ITestS3AContractDistCp but this seems not to be related. I can not reproduce the error by running the ITestS3AContractDistCp alone by `-Dit.test=ITestS3AContractDistCp`.

```
$  mvn verify -Dtest=x -Dit.test=ITestS3AContract*
...
[INFO] Running org.apache.hadoop.fs.contract.s3a.ITestS3AContractDistCp
[ERROR] Tests run: 8, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 47.658 s <<< FAILURE! - in org.apache.hadoop.fs.contract.s3a.ITestS3AContractDistCp
[ERROR] testNonDirectWrite(org.apache.hadoop.fs.contract.s3a.ITestS3AContractDistCp)  Time elapsed: 4.055 s  <<< FAILURE!
java.lang.AssertionError: Expected 2 renames for a non-direct write distcp expected:<2> but was:<0>
        at org.apache.hadoop.fs.contract.s3a.ITestS3AContractDistCp.testNonDirectWrite(ITestS3AContractDistCp.java:98)
```

iwasakims:
> 3. don't downgrade FileNotFoundException to 'false' if source doesn't exist
> 4. raise FileAlreadyExistsException if dest path exists or parent is a file.

Should this JIRA be marked as incompatible for applicatinos assuming existing behavior?

hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md
steveloughran: oops. fixed

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/impl/ITestlRenameDeleteRace.java
steveloughran: done. My inability to type well surfaces again

hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/impl/ITestRenameDeleteRace.java
```diff
@@ -0,0 +1,248 @@
...
+
+package org.apache.hadoop.fs.s3a.impl;
+
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+
+import com.amazonaws.AmazonClientException;
+import org.apache.hadoop.thirdparty.com.google.common.util.concurrent.ListeningExecutorService;
```
iwasakims: We need fix similar to #2758 after #2575.

steveloughran: gotcha. Can we add some style checking to block this?

steveloughran: no need, the compiler finds it. Will fix

steveloughran:
testing: s3 london, most recently with `-Dparallel-tests -DtestsThreadCount=6 -Dmarkers=keep -Ds3guard -Ddynamo  -Dscale`

iwasakims (approved):
+1. The ITestRenameDeleteRace failed without the fix of S3AFileSystem as expected. The test passed with the fix regardless of `-Ds3guard`. Thanks, @steveloughran.

steveloughran:
ah, thanks -lovely. just tweaked the imports slightly for better backporting; I'll merge once the next compile is good, then do a cp and retest for branch-3.3.

PR #2746, opened 2021-03-05
HDFS-15875. Check whether file is being truncated before truncate

ferhui:
## NOTICE

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/server/namenode/TestFileTruncate.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/server/namenode/TestFileTruncate.java
@@ -33,6 +33,7 @@ import static org.junit.Assert.fail;
 import java.io.IOException;
 import java.util.concurrent.ThreadLocalRandom;
 
+import org.apache.hadoop.hdfs.server.datanode.DataNodeFaultInjector;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.hadoop.HadoopIllegalArgumentException;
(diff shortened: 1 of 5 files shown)


ferhui:
@ayushtkn Could you please help to  review this? Thanks

hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/server/namenode/TestFileTruncate.java
```diff
@@ -218,6 +219,65 @@ public void testSnapshotTruncateThenDeleteSnapshot() throws IOException {
...
+            }
+        });
+    t.start();
+    t.join();
+    Thread.sleep(60000);
+    try {
+      fs.truncate(p, data.length - 2);
+    } catch (IOException e) {
+      //GenericTestUtils.assertExceptionContains("is being truncated.", e);
+    }
```
ayushtkn: Can use LambdaTestUtils.

```
    LambdaTestUtils.intercept(RemoteException.class,
        "/testTruncateTwiceTogether/file is being truncated",
        () -> fs.truncate(p, data.length - 2));
```

hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/server/namenode/TestFileTruncate.java
```diff
@@ -218,6 +219,65 @@ public void testSnapshotTruncateThenDeleteSnapshot() throws IOException {
...
+  public void testTruncateTwiceTogether() throws Exception {
+
+    Path dir = new Path("/testTruncateTwiceTogether");
+    fs.mkdirs(dir);
+    final Path p = new Path(dir, "file");
+    final byte[] data = new byte[100 * BLOCK_SIZE];
+    ThreadLocalRandom.current().nextBytes(data);
+    writeContents(data, data.length, p);
+
+    DataNodeFaultInjector originInjector = DataNodeFaultInjector.get();
```
ayushtkn, line 237: Is this used somewhere?

hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/server/namenode/TestFileTruncate.java
```diff
@@ -218,6 +219,65 @@ public void testSnapshotTruncateThenDeleteSnapshot() throws IOException {
...
+    DataNodeFaultInjector injector = new DataNodeFaultInjector() {
+      @Override
+      public void delay() {
+        try {
+          // Bigger than soft lease period.
+          Thread.sleep(65000);
+        } catch (InterruptedException e) {
+          e.printStackTrace();
```
ayushtkn: No need to print the trace, if possible add Exception to method signature and remove this try-catch, else ignore the exception

ayushtkn (commented):
Thanx @ferhui for the fix, The prod changes LGTM. Minor stuff in the test.
Apart, I see there are bunch of sleeps, can you give a check if the amount can be reduced by means of any config, or if there is any scope of using GenericTestUtil.waitFor in the test, just to ensure this test doesn't go flaky, and do increase the test timeout also a bit.

ferhui:
@ayushtkn Thanks for review ! Will commit the fix soon!

hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/server/namenode/TestFileTruncate.java
```diff
@@ -631,10 +698,10 @@ public void testTruncateFailure() throws IOException {
     {
       try {
         fs.truncate(p, 0);
-        fail("Truncate must fail since a trancate is already in pregress.");
+        fail("Truncate must fail since a truncate is already in progress.");
       } catch (IOException expected) {
         GenericTestUtils.assertExceptionContains(
-            "Failed to TRUNCATE_FILE", expected);
+            "/dir/testTruncateFailure is being truncated", expected);
       }
```
ayushtkn: What is the difference between this case and ours? Why doesn't this exception triggers for our case? and can we accommodate our fix to the check throwing this exception

ferhui: @ayushtkn Thanks for review! 
This case happens with the same lease holder when enter recoverLeaseInternal. And our fix throw exception before enter recoverLeaseInternal. Our UT has 2 clients with different client names.
Will upload fix to throw the same exception.

ferhui:
@ayushtkn Any further questions? Thanks!

ayushtkn (approved):
Changes LGTM

ferhui:
@ayushtkn Thanks for review ! 
Merged

PR #2764, opened 2021-03-12
HADOOP-17581. Fix reference to LOG is ambiguous after HADOOP-17482.

xiaoyuyao:
JIRA: https://issues.apache.org/jira/browse/HADOOP-17581


cxorm (approved):
Thanks @xiaoyuyao for fixing it, +1.
( and I think we could resolve HADOOP-17578 after merging this patch. )

xiaoyuyao:
Thanks @cxorm for the review. There are still CI failures but I don't think it is caused by this fix. I will merge this one shortly.

PR #3499, opened 2021-09-29
HADOOP-11867. Add a high performance vectored read API to file system.

mukund-thakur:
Rebased work on top of https://github.com/apache/hadoop/pull/1830 

Conflicts:
	hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/BufferedFSInputStream.java
	hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/ChecksumFileSystem.java
	hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FSDataInputStream.java
	hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/RawLocalFileSystem.java
	pom.xml

### Description of PR
Adding support for multiple ranged read async api in PositionedReadable. The default iterates through the ranges to read each synchronously, but the intent is that FSDataInputStream subclasses can make more efficient readers especially object stores implementation.

Added benchmarks. 
Added UT's
Added new contract tests for new API spec.

### For code changes:


mukund-thakur:
CC @steveloughran  @mehakmeet @omalley

hadoop-common-project/benchmark/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                      https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.hadoop</groupId>
+    <artifactId>hadoop-project</artifactId>
+    <version>3.4.0-SNAPSHOT</version>
+    <relativePath>../../hadoop-project</relativePath>
+  </parent>
+  <artifactId>hadoop-benchmark</artifactId>
```
steveloughran: think this should go into hadoop-tools, as it allows for more things to go in later

hadoop-common-project/benchmark/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+  <version>3.4.0-SNAPSHOT</version>
+  <packaging>jar</packaging>
+
+  <name>Apache Hadoop Common Benchmark</name>
+  <description>Apache Hadoop Common Benchmark</description>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.compiler.useIncrementalCompilation>false</maven.compiler.useIncrementalCompilation>
+    <jmh.version>1.20</jmh.version>
```
steveloughran: declaration to go into hadoop project,

hadoop-common-project/benchmark/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
```
steveloughran: declaration to go into hadoop project, with reference here

hadoop-common-project/benchmark/src/main/java/org/apache/hadoop/benchmark/AsyncBenchmark.java
```diff
@@ -0,0 +1,242 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.benchmark;
+
+import org.apache.hadoop.conf.Configuration;
```
steveloughran: usual ordering nits

hadoop-common-project/benchmark/src/main/java/org/apache/hadoop/benchmark/AsyncBenchmark.java
```diff
@@ -0,0 +1,242 @@
...
+import java.nio.file.FileSystems;
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.function.IntFunction;
+
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+public class AsyncBenchmark {
```
steveloughran: wonder what it'd take to make this a contract test -or at least deployable on the command line?

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/RawLocalFileSystem.java
```diff
@@ -267,8 +282,84 @@ public boolean hasCapability(String capability) {
     public IOStatistics getIOStatistics() {
       return ioStatistics;
     }
+
+    // QQ: Should we make this synchronized?
```
steveloughran: yes. Also needs to reject if file is closed.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/RawLocalFileSystem.java
```diff
@@ -267,8 +282,84 @@ public boolean hasCapability(String capability) {
...
+        AsynchronousFileChannel channel = getAsyncChannel();
+        ByteBuffer[] buffers = new ByteBuffer[ranges.size()];
+        AsyncHandler asyncHandler = new AsyncHandler(channel, ranges, buffers);
+        for(int i = 0; i < ranges.size(); ++i) {
+          FileRange range = ranges.get(i);
+          buffers[i] = allocate.apply(range.getLength());
+          channel.read(buffers[i], range.getOffset(), i, asyncHandler);
+        }
+      } catch (IOException ioe) {
+        LOG.info("Can't get async channel", ioe);
```
steveloughran: propose log full stack at debug

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/impl/TestAsyncReaderUtils.java
```diff
@@ -0,0 +1,301 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test behavior of {@link AsyncReaderUtils}.
+ */
+public class TestAsyncReaderUtils  {
```
steveloughran: 1. `extends HadoopTestBase`
2. should go for AssertJ

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/impl/TestAsyncReaderUtils.java
```diff
@@ -0,0 +1,301 @@
...
+    Mockito.doThrow(new IOException("foo"))
+        .when(stream).readFully(ArgumentMatchers.anyLong(),
+                                ArgumentMatchers.any(ByteBuffer.class));
+    result =
+        AsyncReaderUtils.readRangeFrom(stream, new FileRangeImpl(1000, 100),
+            ByteBuffer::allocate);
+    assertTrue(result.isCompletedExceptionally());
+  }
+
+  static void runReadRangeFromPositionedReadable(IntFunction<ByteBuffer> allocate) throws Exception {
```
steveloughran: be good to make sure this backports ~OK, even before committing

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/impl/TestAsyncReaderUtils.java
```diff
@@ -0,0 +1,301 @@
...
+        buffer[i] = b++;
+      }
+      return null;
+    }).when(stream).readFully(ArgumentMatchers.anyLong(),
+        ArgumentMatchers.any(), ArgumentMatchers.anyInt(),
+        ArgumentMatchers.anyInt());
+    CompletableFuture<ByteBuffer> result =
+        AsyncReaderUtils.readRangeFrom(stream, new FileRangeImpl(1000, 100),
+            allocate);
+    assertTrue(result.isDone());
```
steveloughran: pull out to separate asserts for reuse. Maybe we should add to MoreAsserts as some
```
public static <T> void assertCompletedSuccesfully(CompletableFuture<T> future) 
{
    + both asserts with a useful .toString value, and on an exceptional failure throw the exception
}
```

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/PositionedReadable.java
```diff
@@ -18,9 +18,13 @@
 package org.apache.hadoop.fs;
 
 import java.io.*;
```
steveloughran: might be time to expand the .*, given we are editing the imports

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/PositionedReadable.java
```diff
@@ -85,4 +89,37 @@ void readFully(long position, byte[] buffer, int offset, int length)
    * the read operation completed
    */
   void readFully(long position, byte[] buffer) throws IOException;
+
+  /**
+   * What is the smallest reasonable seek?
+   * @return the minimum number of bytes
+   */
+  default int minimumReasonableSeek() {
```
steveloughran: maybre add `forVectorReads` on these, to show what they are for

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/PositionedReadable.java
```diff
@@ -85,4 +89,37 @@ void readFully(long position, byte[] buffer, int offset, int length)
...
+   * The default iterates through the ranges to read each synchronously, but
+   * the intent is that FSDataInputStream subclasses can make more efficient
+   * readers.
+   * As a result of the call, each range will have FileRange.setData(CompletableFuture)
+   * called with a future that when complete will have a ByteBuffer with the
+   * data from the file's range.
+   * @param ranges the byte ranges to read
+   * @param allocate the function to allocate ByteBuffer
+   */
+  default void readAsync(List<? extends FileRange> ranges,
```
steveloughran: * `readVectored()`; the async is part of the implementation
* add `throws IOException` in case implementations do this.

hadoop-common-project/hadoop-common/pom.xml
```diff
@@ -651,6 +651,14 @@
           </filesets>
         </configuration>
       </plugin>
+        <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+                <source>8</source>
```
steveloughran: versions to be set in hadoop project pom

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/impl/AsyncReaderUtils.java
```diff
@@ -0,0 +1,217 @@
...
+                                                            IntFunction<ByteBuffer> allocate) {
+    CompletableFuture<ByteBuffer> result = new CompletableFuture<>();
+    try {
+      ByteBuffer buffer = allocate.apply(range.getLength());
+      if (stream instanceof ByteBufferPositionedReadable) {
+        ((ByteBufferPositionedReadable) stream).readFully(range.getOffset(),
+            buffer);
+        buffer.flip();
+      } else {
+        if (buffer.isDirect()) {
```
steveloughran: propose pulling this out into its own static method which can be tested in isolation, maybe used in implementations

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/impl/CombinedFileRange.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A file range that represents a set of underlying file ranges.
+ * This is used when we combine the user's FileRange objects
+ * together into a single read for efficiency.
+ */
+public class CombinedFileRange extends FileRangeImpl {
```
steveloughran, line 32: might be good for the toString() to provide more info than the superclass does

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/ChecksumFileSystem.java
```diff
@@ -286,8 +295,123 @@ protected int readChunk(long pos, byte[] buf, int offset, int len,
     public IOStatistics getIOStatistics() {
       return IOStatisticsSupport.retrieveIOStatistics(datas);
     }
+
+    public static long findChecksumOffset(long dataOffset,
```
steveloughran, line 299: does bytesPerSum ever come in as 0?

steveloughran:
Quick review. 
* needs filesystem spec
* and a contract test
* test to include 
  -overlapping ranges
  -beyond file length
  -negative values
  -on heap and off heap buffers

* be good to have a PoC of the s3a one, so that we can happy the API xfers to the store
* and some ORC/parquet branch which shows it can be used
* is there any API we'd want, e.g an fs command which does a single async ranged get
```
hadoop fs ranged -start 0 -end 345  -out output.txt s3a://file
```
would be good for QE tests
* and we should have a path capability. Anything which supports seek will also do this (so ftp won't)

hadoop-common-project/benchmark/src/main/java/org/apache/hadoop/benchmark/AsyncBenchmark.java
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  @Benchmark
+  public void asyncRead(FileSystemChoice fsChoice,
+                        BufferChoice bufferChoice,
+                        Blackhole blackhole) throws Exception {
+    FSDataInputStream stream = fsChoice.fs.open(DATA_PATH);
+    List<FileRange> ranges = new ArrayList<>();
+    for(int m=0; m < 100; ++m) {
+      FileRangeImpl range = new FileRangeImpl(m * 1024L * 1024, 64 * 1024);
```
bogthe: Won't this move the offset every `1024*1024` bytes and only have a length of `64*1024`? Meaning you'll read `64` and move to `1024.` The same code is used 3 times in this file `m * 1024` so if it should be `64` make sure to replace all 3 of them.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/ChecksumFileSystem.java
bogthe: Looks like it can either be `1` (if constructor init fails) or another int.

bogthe (commented):
Did an initial review and looks promising.

I'm super curious to see what an async read will bring in terms of performance improvements for a job (even if it's a quick scrappy implementation in the FS).

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/RawLocalFileSystem.java
mukund-thakur: isClosed is not implemented in RawLocalFileSystem. This needs to be added separately.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/impl/CombinedFileRange.java
mukund-thakur: yes this makes sense. But there are many tests based on toString() value match. I think it is a low priority now but will do for sure before final merge.

hadoop-common-project/benchmark/src/main/java/org/apache/hadoop/benchmark/AsyncBenchmark.java
mukund-thakur: Will look at this later point in time.

hadoop-common-project/hadoop-common/src/site/markdown/filesystem/fsdatainputstream.md
```diff
@@ -443,6 +443,37 @@ The semantics of this are exactly equivalent to
 That is, the buffer is filled entirely with the contents of the input source
 from position `position`
 
+### `default void readAsync(List<? extends FileRange> ranges, IntFunction<ByteBuffer> allocate)`
```
mukund-thakur: will update to readVectored.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/impl/TestVectoredReadUtils.java
```diff
@@ -0,0 +1,315 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.fs.impl;
+
+import org.apache.hadoop.fs.ByteBufferPositionedReadable;
```
steveloughran, line 21: Nit: imports

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/impl/TestVectoredReadUtils.java
```diff
@@ -0,0 +1,315 @@
...
+    assertEquals(13, VectoredReadUtils.roundUp(13, 1));
+  }
+
+  @Test
+  public void testMerge() {
+    FileRange base = new FileRangeImpl(2000, 1000);
+    CombinedFileRange mergeBase = new CombinedFileRange(2000, 3000, base);
+
+    // test when the gap between is too big
+    assertFalse(mergeBase.merge(5000, 6000,
```
steveloughran, line 109: Can you add error message text to the true and false assertions?

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/impl/VectoredReadUtils.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.IntFunction;
+
+import org.apache.hadoop.fs.ByteBufferPositionedReadable;
+import org.apache.hadoop.fs.FileRange;
+import org.apache.hadoop.fs.PositionedReadable;
+import org.apache.hadoop.util.Preconditions;
+
+public class VectoredReadUtils {
```
steveloughran, line 36: Nit: javadocs

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/ChecksumFileSystem.java
```diff
@@ -66,7 +74,7 @@
   public static double getApproxChkSumLength(long size) {
     return ChecksumFSOutputSummer.CHKSUM_AS_FRACTION * size;
   }
-  
+
```
steveloughran, line 77: These are all needless and making the patch a bit bigger than it should be. was yetus complaining?

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/ChecksumFileSystem.java
mukund-thakur: It is written in the cheksum file as the first int using `sums.writeInt(bytesPerSum);`
Default value is 512. 
If reading this value  causes some exception it defaults to 1. 
So I don't think it can ever come as 0.

steveloughran:
I just realized I had some pending comments on one of the updates; maybe out of date now and they were all minor. Let me re-review.

steveloughran (commented):
here's some comments I had outstanding since 2021 .. sorry. will do a bigger review now

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractVectoredReadTest.java
```diff
@@ -112,6 +117,21 @@ public void testVectoredReadAndReadFully()  throws Exception {
     }
   }
 
+
+  @Test
+  public void testVectoredReadBigFile()  throws Exception {
+    FileSystem fs = getFileSystem();
+    List<FileRange> fileRanges = new ArrayList<>();
+    fileRanges.add(new FileRangeImpl(1293, 25837));
+    try (FSDataInputStream in = fs.open(path(VECTORED_READ_FILE_1MB_NAME))) {
```
steveloughran, line 126: can you use openFile here? just to make sure that codepath is happy

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractVectoredReadTest.java
```diff
@@ -35,6 +35,9 @@
   public static final int DATASET_LEN = 1024;
   private static final byte[] DATASET = ContractTestUtils.dataset(DATASET_LEN, 'a', 32);
   private static final String VECTORED_READ_FILE_NAME = "vectored_file.txt";
+  private static final String VECTORED_READ_FILE_1MB_NAME = "vectored_file_1M.txt";
+  private static final byte[] DATASET_MB = ContractTestUtils.dataset(1024 * 1024, 'a', 256);
```
steveloughran, line 39: are we OK with this not being in a scale test?

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
```diff
@@ -955,11 +955,27 @@ private void populateBuffer(int length,
                               S3ObjectInputStream objectContent) throws IOException {
     if (buffer.isDirect()) {
       byte[] tmp = new byte[length];
-      objectContent.read(tmp, 0, length);
+      readByteArray(objectContent, tmp, 0, length);
```
steveloughran, line 958: maybe we should create a buffer of a max size, read in blocks and put incrementally. otherwise we alloc a 1GB array to load a range that big, when we don't need so much heap

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
```diff
@@ -887,18 +887,22 @@ private void updateOriginalRange(FileRange child,
...
-   * @param range
-   * @throws EOFException
+   * @param range requested range.
+   * @throws EOFException end of file exception.
    */
   private void validateRangeRequest(FileRange range) throws EOFException {
     VectoredReadUtils.validateRangeRequest(range);
     if(range.getOffset() + range.getLength() > contentLength) {
-      throw new EOFException("Requested range is beyond EOF");
+      LOG.error("Requested range [{}, {}) is beyond EOF",
```
steveloughran, line 902: why log at error?

steveloughran:
OK, I've had a look at the recent changes, and had some shared screen reviews over zoom

here's what I propose


* create a feature branch  in the asf repo where we can rebase/reorder commits 
*  get what is done in as the base commit
* then we can split work on the final changes up into new PRs, each with their own JIRA.


where those changes can be

* buffer pooling (API to these calls and better weak reference handling in the elastic one or nearby)
* s3a split handling
* huge file support (new tests in the s3a huge file suite)
 
then, once happy we can do a rebase followed by merge commit into trunk/branch-3.

key point: this patch has been going and is ready to stabilize, which can be done with incremental changes in a feature branch.

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/ChecksumFileSystem.java
mukund-thakur: I don't know how it is showing up here. In intellJ diff I don't see this.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractVectoredReadTest.java
mukund-thakur: 1 mb should be fine for now. Anyway I will write a scale test for huge file test separately and then delete it from here.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
```diff
@@ -100,6 +114,7 @@
   private S3ObjectInputStream wrappedStream;
   private final S3AReadOpContext context;
   private final InputStreamCallbacks client;
+  private final ThreadPoolExecutor unboundedThreadPool;
```
mehakmeet, line 117: A comment describing that this threadpool is used for vectored reads would be helpful

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractVectoredReadTest.java
```diff
@@ -0,0 +1,291 @@
...
+  private static final String VECTORED_READ_FILE_NAME = "vectored_file.txt";
+  private static final String VECTORED_READ_FILE_1MB_NAME = "vectored_file_1M.txt";
+  private static final byte[] DATASET_MB = ContractTestUtils.dataset(1024 * 1024, 'a', 256);
+
+
+  private final IntFunction<ByteBuffer> allocate;
+
+  private final String bufferType;
+
+  @Parameterized.Parameters
```
mehakmeet, line 46: Use "name" in Parameters, for better readability of the test on IDE. Something like 
```
@Parameterized.Parameters(name = "Buffer type : {0}")
```
Would be helpful in realizing which test is running direct or array bufferType.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
```diff
@@ -793,6 +810,208 @@ public void readFully(long position, byte[] buffer, int offset, int length)
...
+    // starvation for other api's like list, delete etc.
+    // TODO: what if combined range becomes so big that memory can't be allocated.
+    checkNotClosed();
+    for (FileRange range : ranges) {
+      validateRangeRequest(range);
+      CompletableFuture<ByteBuffer> result = new CompletableFuture<>();
+      range.setData(result);
+    }
+
+    if (isOrderedDisjoint(ranges, 1, minSeekForVectorReads())) {
```
mehakmeet, line 838: Debug Logs indicating that the range was ordered disjoint and would not be merged and the same if it's merged.

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
mukund-thakur: I made this change but then realised, the temp array will never be greater than maxReadSizeForVectorReads which is by default 1 MB.

hadoop-tools/hadoop-aws/src/test/resources/log4j.properties
```diff
@@ -52,7 +52,7 @@ log4j.logger.org.apache.hadoop.ipc.Server=WARN
 
 # for debugging low level S3a operations, uncomment these lines
 # Log all S3A classes
-#log4j.logger.org.apache.hadoop.fs.s3a=DEBUG
+log4j.logger.org.apache.hadoop.fs.s3a=DEBUG
```
mehakmeet, line 55: Maybe put in while testing, shouldn't be DEBUG by default I assume?

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/test/MoreAsserts.java
```diff
@@ -57,10 +62,28 @@
...
+            .describedAs("This future is supposed to be " +
+                    "completed successfully")
+            .isTrue();
+    Assertions.assertThat(future.isCompletedExceptionally())
+            .describedAs("This future is supposed to be " +
+                    "completed successfully")
+            .isFalse();
+  }
+
+  public static <T> void assertFutureFailedExceptionaly(CompletableFuture<T> future) {
```
mehakmeet, line 83: typo: "Exceptionally"

hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java
```diff
@@ -46,8 +51,17 @@
 import java.io.EOFException;
 import java.io.IOException;
 import java.net.SocketTimeoutException;
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
```
mehakmeet, line 57: nit: Not used, could be removed.

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/localfs/TestLocalFSConractVectoredRead.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.hadoop.fs.contract.localfs;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.contract.AbstractContractVectoredReadTest;
+import org.apache.hadoop.fs.contract.AbstractFSContract;
+
+public class TestLocalFSConractVectoredRead extends AbstractContractVectoredReadTest {
```
mehakmeet, line 7: typo: "Contract"

mehakmeet (commented):
Did an initial Review, looks really good.
I think we should add an integration test around the ordered disjoint list of ranges as well to showcase them not merging and reading or having a few ranges where some are merged and some aren't by validating them.
Also, are both minSeek and maxReadSize going to be configurable? I assume they are hardcoded just so you can test them out..

hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractVectoredReadTest.java
mukund-thakur: Nice thanks

mukund-thakur:
> Did an initial Review, looks really good. I think we should add an integration test around the ordered disjoint list of ranges as well to showcase them not merging and reading or having a few ranges where some are merged and some aren't by validating them. 

Yeah already in my TODO. Thanks.

> Also, are both minSeek and maxReadSize going to be configurable? I assume they are hardcoded just so you can test them out..
Yes I will be making them configurable in consecutive PR's

Also Please refer to the new PR now. https://github.com/apache/hadoop/pull/3904

steveloughran:
now this is in the feature branch, this can be closed

PR #3561, opened 2021-10-18
YARN-10980:fix CVE-2020-8908

lujiefsi:
see  https://www.cvedetails.com/cve/CVE-2020-8908/

 

A temp directory creation vulnerability exists in all versions of Guava, allowing an attacker with access to the machine to potentially access data in a temporary directory created by the Guava API com.google.common.io.Files.createTempDir(). By default, on unix-like systems, the created directory is world-readable (readable by an attacker with access to the system). The method in question has been marked @Deprecated in versions 30.0 and later and should not be used. For Android developers, we recommend choosing a temporary directory API provided by Android, such as context.getCacheDir(). For other Java developers, we recommend migrating to the Java 7 API java.nio.file.Files.createTempDirectory() which explicitly configures permissions of 700, or configuring the Java runtime's java.io.tmpdir system property to point to a location whose permissions are appropriately configured.


lujiefsi:
due to forget to fetech upstream, i open this issue. this bug is fixed by https://issues.apache.org/jira/browse/HADOOP-17653.

PR #3954, opened 2022-02-02
Rebase Access Point feature onto branch-3.3

bogthe:
Rebased Access Point feature (the 2 new commits `648a7ff329d7458d97e8b04461ccdf56eab3456a` and `e234d31af24a375f3942e5ac147ed54e6587dc0f`) onto `branch-3.3` right before the AWS SDK upgrade. Was hoping for a quick `cherry-pick` but there's changes that come into effect with removal S3 guard which conflict so placed it before that. 

Apologies if this isn't the right way to do it! Let me know if there's anything else I can do.

Also created a PR that picks this AcessPoint feature straight into 3.3.2, which might be of help https://github.com/apache/hadoop/pull/3955 

@sunchao @steveloughran


sunchao:
This seems too disruptive. @bogthe is there anyway we can resolve the conflicts due to the removal of s3guard?

bogthe:
@sunchao not really. Whatever you resolve you can't then pick onto 3.3.2 since the changes you apply onto `3.3` need to be modified because of the S3 Guard removal.

The only non-distruptive way is to pick straight into 3.3.2 like this PR is trying to do https://github.com/apache/hadoop/pull/3955

sunchao:
I see. I'm slightly in favor of putting this in branch-3.3 for now, and then release it in Hadoop 3.3.3. What do you think?

My apologies that I didn't realize the matter is a bit complex due to the difference between branch-3.3 and branch-3.3.2. Since at this point we're wrapping up the release of 3.3.2, I'd prefer to avoid extra complexities which could delay the release further.

steveloughran:
the 3.3.2 fix isn't that big, it's that because the s3guard removal is in branch-3.3, there's a bit more toe stepping going on there

bogthe:
> the 3.3.2 fix isn't that big, it's that because the s3guard removal is in branch-3.3, there's a bit more toe stepping going on there

@sunchao would like to close this PR and use Steve's new one (basically does the thing this one intended to 😄 )

steveloughran:
closing this.

sunchao:
Oh cool, thanks @steveloughran !

steveloughran:
merged the other one in

PR #4284, opened 2022-05-08
YARN-11135. Remove unused Imports in hadoop project

hotcodemacha:
### Description of PR
Remove unused Imports in hadoop project
* JIRA: YARN-11135


hotcodemacha:
@aajisaka - Can you please have a look? Thanks.

steveloughran:
-1. wontfix. 

if this were to go in

* every pending PR would no longer apply without rebasing/reworking
* every subsequent pr would be impossible to backport without reversing the work

we have a fraction of this problem even with the movement of unshaded guava to shaded guava to hadoop's own Precondition classes; moving things from trunk to 3.3. to internal PRs always fail to merge at those two or three imports. 

we couldn't cope with this at scale. 

this is also why we don't go overboard in reordering imports in the files touched within a normal PR: IDEs may think they are being helpful by cleaning up imports, but aren't. Removing unused imports is fine, but big cleanups, tht causes too many problems.

Sorry. if you had to deal with backporting you'd understand why.

hotcodemacha:
@steveloughran - What you are saying makes sense. But let's do some clean up the code. I am changing the scope and only removing the unused imports. What do you think about it?

hotcodemacha:
Closing this PR as discussed offline with @aajisaka. Will create separate JIRA and PR for each Hadoop project and reduce the scope to only remove unused imports.

steveloughran:
ok. for tools/ do one for the different submodules, that way backporting is a lot easier.

hotcodemacha:
Thanks @steveloughran. I have started with Hadoop Map/Reduce. Here is the PR of that. https://github.com/apache/hadoop/pull/4299. Can you please review? Thanks.

PR #3749, opened 2021-12-04
HADOOP-18033. Upgrade fasterxml Jackson to 2.13.0

virajjasani:
### Description of PR
Upgrade fasterxml Jackson to 2.13.0.

### For code changes:

- [X] Does the title or this PR starts with the corresponding JIRA issue id (e.g. 'HADOOP-17799. Your PR title ...')?
- [X] If applicable, have you updated the `LICENSE`, `LICENSE-binary`, `NOTICE-binary` files?

PR #4460, opened 2022-06-18
HADOOP-18033. [WIP] Remove jsr311-api dependency

virajjasani:
### Description of PR
More discussion on HADOOP-18033

### DIFF
--- a/LICENSE-binary
+++ b/LICENSE-binary
@@ -492,7 +492,6 @@ javax.servlet:javax.servlet-api:3.1.0
 javax.servlet.jsp:jsp-api:2.1
 javax.websocket:javax.websocket-api:1.0
 javax.ws.rs:javax.ws.rs-api:2.1.1
-javax.ws.rs:jsr311-api:1.1.1
 javax.xml.bind:jaxb-api:2.2.11
 
 
--- a/hadoop-client-modules/hadoop-client-minicluster/pom.xml
+++ b/hadoop-client-modules/hadoop-client-minicluster/pom.xml
@@ -396,6 +396,14 @@
       </exclusions>
     </dependency>
 
+    <!-- Adding javax.ws.rs-api so we can make it optional to remove from our transitive tree -->
+    <dependency>
+      <groupId>javax.ws.rs</groupId>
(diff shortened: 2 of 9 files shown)


ayushtkn:
Build kind of died only. Guess the test failures screwed the build in some way, We already know something will break, just wanted to know what all break and how big is the impact....

virajjasani:
Exactly, I wish at least we could know all the failures (even though we know we are stuck with this dependency)

pjfanning:
Looks like jersey-core:1.19 depends on jsr311-api and jsr311-api and rs-api cannot be on the classpath together because they clash.
So it looks like the new dependency on rs-api needs to be removed. This may not require jackson to be downgraded too.

ayushtkn:
@pjfanning Cool, if I understood correct, you mean: javax.ws.rs-api can be removed without Jackson downgrade.
Go ahead you can create a PR with that change, if things work fine then most probably we can save us from downgrading Jackson. :-)

pjfanning:
I have https://github.com/apache/hadoop/pull/4547 - let's see if that passes CI build

virajjasani:
> So it looks like the new dependency on rs-api needs to be removed. This may not require jackson to be downgraded too.

javax.ws.rs-api was introduced only to support Jackson 2.13 upgrade. Without rs-api, (on Jackson 2.13) majority of Yarn and ATS tests fail as jackson 2.13 does have dependency on JAX-RS 2 based rs-api (PR #3749)

pjfanning:
@virajjasani I think https://github.com/FasterXML/jackson-jaxrs-providers/issues/134 which appeared in jackson-jaxrs for v2.13.0 to be the reason rs-api was added to hadoop - so in #4547, I am looking to downgrade to jackson 2.12.7

virajjasani:
> @virajjasani I think [FasterXML/jackson-jaxrs-providers#134](https://github.com/FasterXML/jackson-jaxrs-providers/issues/134) which appeared in jackson-jaxrs for v2.13.0 to be the reason rs-api was added to hadoop - so in #4547, I am looking to downgrade to jackson 2.12.7

Exactly, without downgrading Jackson, it's not possible to remove javax.ws.rs-api.

PR #5195, opened 2022-12-07
HDFS-16856: Refactor RouterAdmin to use the AdminHelper class.

omalley:
### Description of PR

This simplifies the structure of RouterAdmin to make the various subcommands, much more consistent and have less repetition throughout the code. The user visible change is that some of the error and help messages have changed and in general are more consistent.

The unit test TestRouterAdminCLI was run. I'll also test the tool against our routers.

### DIFF
--- a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/StringUtils.java
+++ b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/StringUtils.java
@@ -1135,7 +1135,25 @@ public class StringUtils {
     }
     return val;
   }
-  
+
+  /**
+   * From a list of command-line arguments, remove both an option and the
+   * next argument.
+   * @param name the name of the option
+   * @param args the list of arguments
+   * @param defaultValue the desired default value
+   * @return the value of the argument or the default value if the option wasn't
+   *         present.
+   * @throws IllegalArgumentException if the option doesn't have an argument
+   */
+  public static String popOptionWithArgument(String name,
+                                             List<String> args,
+                                             String defaultValue)
+      throws IllegalArgumentException {
+    String result = popOptionWithArgument(name, args);
+    return result == null ? defaultValue : result;
+  }
+
   /**
    * From a list of command-line arguments, remove an option.
    *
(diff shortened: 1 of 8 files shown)


hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/StringUtils.java
```diff
@@ -1182,6 +1200,19 @@ public static String popFirstNonOption(List<String> args) {
...
   }
+  /**
+   * From a list of command-line arguments, ensure that all of the arguments
+   * have been used except a possible "--".
+   *
+   * @param args  List of arguments.
+   * @throws IllegalArgumentException if some arguments were not used
+   */
+  public static void ensureAllUsed(List<String> args) throws IllegalArgumentException {
+    if (!args.isEmpty() && !(args.size() == 1 && "--".equals(args.get(0)))) {
```
goiri, line 1211: I don't think we need some of the parenthesis.

hadoop-hdfs-project/hadoop-hdfs-rbf/src/main/java/org/apache/hadoop/hdfs/tools/federation/RouterAdmin.java
```diff
@@ -439,24 +266,23 @@ public int run(String[] argv) throws Exception {
       try {
         String[] content;
         content = e.getLocalizedMessage().split("\n");
-        System.err.println(cmd.substring(1) + ": " + content[0]);
+        System.err.println(cmd+ ": " + content[0]);
```
goiri, line 269: Space.

hadoop-hdfs-project/hadoop-hdfs-rbf/src/main/java/org/apache/hadoop/hdfs/tools/federation/RouterAdmin.java
```diff
@@ -467,310 +293,265 @@ public int run(String[] argv) throws Exception {
...
+   * @param name the name of the argument
+   * @param parameters the list of all of the parameters
+   * @return null if the argument wasn't present, true or false based on the
+   * string
+   */
+  static Boolean parseOptionalBoolean(String name, List<String> parameters) {
+    String val = StringUtils.popOptionWithArgument(name, parameters);
+    if (val == null) {
+      return null;
+    } else {
```
goiri, line 380: As you return before, skip the else.

hadoop-hdfs-project/hadoop-hdfs-rbf/src/main/java/org/apache/hadoop/hdfs/tools/federation/RouterAdmin.java
```diff
@@ -467,310 +293,265 @@ public int run(String[] argv) throws Exception {
...
+      switch (val.toLowerCase()) {
```
goiri, line 381: Isn't there a parseBool or something?

hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/StringUtils.java
omalley: I find writing code that depends overly on knowing the precedence for non-math operators leads to trouble, so I'd prefer to leave them in.

PR #5356, opened 2023-02-06
HADOOP-18620 Avoid using grizzly-http-* APIs

virajjasani:
As discussed on the parent Jira HADOOP-15984, we do not have any grizzly-http-servlet version available that uses Jersey 2 dependencies. 

version 2.4.4 contains Jersey 1 artifacts: https://repo1.maven.org/maven2/org/glassfish/grizzly/grizzly-http-servlet/2.4.4/grizzly-http-servlet-2.4.4.pom

The next higher version available is 3.0.0-M1 and it contains Jersey 3 artifacts: https://repo1.maven.org/maven2/org/glassfish/grizzly/grizzly-http-servlet/3.0.0-M1/grizzly-http-servlet-3.0.0-M1.pom

Moreover, we do not use grizzly-http-* modules extensively. We use them only for few tests such that we don't have to implement all the methods of HttpServletResponse for our custom test classes.

We should get rid of grizzly-http-servlet, grizzly-http and grizzly-http-server artifacts of org.glassfish.grizzly and rather implement HttpServletResponse directly to avoid having to depend on grizzly upgrades as part of overall Jersey upgrade.

PR #5374, opened 2023-02-09
HADOOP-18620 Avoid using grizzly-http-* APIs (#5356)

virajjasani:
branch-3.3 backport PR of https://github.com/apache/hadoop/commit/4fcceff535552342c225066c4150f3997535a06a


jojochuang (approved):
+1 the removal is in test scope only, so we can backport to 3.3.x without backward compatibility concerns.

PR #5977, opened 2023-08-23
HADOOP-18860. Upgrade mockito version to 4.11.0

anmolanmol1234:
HADOOP-18860: Upgrade mockito version to 4.11.0

PR #6020, opened 2023-09-04
Oss/hadoop java17 runtime feature branch prerequisite

Aakansha99:
### Description of PR
The latest Hadoop version 3.3.3 which supports Java 8 compile and [Java 8, Java 11](https://cwiki.apache.org/confluence/display/HADOOP/Hadoop+Java+Versions) runtime i.e the code is compiled with Java 8 and can be run on both Java 8 and Java 11. The intention of this project is to extend this capability to support Java 17 for the runtime as well for Hadoop version 3.3.3 and above.

Made the following fix in this PR:

1. Upgraded guice version from 4.0 to 5.1.0 to make it compatible for JRE17

2. The current Mocktio (2.x) and PowerMocktio (1.7.x) versions are not compatible with Java 17 and Hence needs a 
appropriate version replacement.
As per the OSS discussion as mentioned mocktio and PowerMocktio, The following version bump up is required
Mockito core ---> Mockito inline 3.6.28
PowerMockito ---> 2.0.0

3. The JavaScript Engine has been removed from Java. Deprecated in Java 11, removed in Java 15.
So, need to use different script engine, as GraalVM

* Tested the patch in local development environments of AWS EMR by building the code, running unit tests, and conducting integration tests. 
* We have also simulated distributed cluster environments by running individual Hadoop services locally. 
* Custom testing frameworks and performance tests ensure compatibility 
* Maintained code reviews, quality assurance and documentation precede integration into the EMR service, which maintained reliability and performance after integration as well

### For code changes:

- [No] Does the title or this PR starts with the corresponding JIRA issue id (e.g. 'HADOOP-17799. Your PR title ...')?
- [Yes] Object storage: have the integration tests been executed and the endpoint declared according to the connector-specific documentation?
- [Yes] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [Yes] If applicable, have you updated the `LICENSE`, `LICENSE-binary`, `NOTICE-binary` files?

### DIFF
--- a/hadoop-common-project/hadoop-common/pom.xml
+++ b/hadoop-common-project/hadoop-common/pom.xml
@@ -241,7 +241,8 @@
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
+      <artifactId>mockito-inline</artifactId>
+      <version>${mockito-inline.version}</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/hadoop-hdfs-project/hadoop-hdfs/pom.xml
+++ b/hadoop-hdfs-project/hadoop-hdfs/pom.xml
@@ -166,7 +166,8 @@ https://maven.apache.org/xsd/maven-4.0.0.xsd">
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
+      <artifactId>mockito-inline</artifactId>
+      <version>${mockito-inline.version}</version>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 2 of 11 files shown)


steveloughran:
hi, welcome to contributing to hadoop.

every jar update should go in as its own pr, own tests, own update to LICENSE-binary (which does need updating, as we log versions in there). that makes it easier to isolate updates. 

Mockito has an active update; your help getting that in should avoid some of the changes here. (this is an example of how we roll back a change...)
HADOOP-18860. Upgrade mockito version to 4.11.0 (#5977)

the graal stuff is its own issue too: how does that work with older releases?

PR #6205, opened 2023-10-19
HADOOP-18946. TestErrorTranslation failure

steveloughran:
Remove invalid assertion

Rerun of failing test suite.

Not. bothered with rest of tests

### For code changes:


ayushtkn (approved):
Change LGTM.

I have like 2 questions around this test class though.

- I am curious around this

```
    if (ioe.getCause() != thrown) {
      throw new AssertionError("Cause of " + ioe + " is not " + thrown, thrown);
    }
```
Why not?
```
    Assertions.assertThat(ioe.getCause()).isEqualTo(thrown);
```

- Why throws here? I think it isn't required
```
  public void testMultiObjectExceptionFilledIn() throws Throwable {
```

steveloughran:
1. I want to throw that nested exception, as it is clearly something bad
2. just my IDE's "test" macro adds it always. and it means if we add more lines, no need to change the signature. matters more with subclasses

PR #6387, opened 2023-12-28
HDFS-17307. docker-compose.yaml sets namenode directory wrong causing datanode failures on restart

matthewrossi:
Restarting existing services using the docker-compose.yaml, causes the datanode to crash after a few seconds.

How to reproduce:

```shell
$ docker-compose up -d # everything starts ok
$ docker-compose stop  # stop services without removing containers
$ docker-compose up -d # everything starts, but datanode crashes after a few seconds
```

The log produced by the datanode suggests the issue is due to a mismatch in the clusterIDs of the namenode and the datanode:

```
datanode_1         | 2023-12-28 11:17:15 WARN  Storage:420 - Failed to add storage directory [DISK]file:/tmp/hadoop-hadoop/dfs/data
datanode_1         | java.io.IOException: Incompatible clusterIDs in /tmp/hadoop-hadoop/dfs/data: namenode clusterID = CID-250bae07-6a8a-45ce-84bb-8828b37b10b7; datanode clusterID = CID-2c1c7105-7fdf-4a19-8ef8-7cb763e5b701 
```

After some troubleshooting I found out the namenode is not reusing the clusterID of the previous run because it cannot find it in the directory set by ENSURE_NAMENODE_DIR=/tmp/hadoop-root/dfs/name. This is due to a change of the default user of the namenode, which is now "hadoop",  so the namenode is actually writing these information to /tmp/hadoop-hadoop/dfs/name.

See [https://issues.apache.org/jira/browse/HDFS-17307](https://issues.apache.org/jira/browse/HDFS-17307)

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -24,7 +24,7 @@ services:
       env_file:
         - ./config
       environment:
-          ENSURE_NAMENODE_DIR: "/tmp/hadoop-root/dfs/name"
+          ENSURE_NAMENODE_DIR: "/tmp/hadoop-hadoop/dfs/name"
    datanode:
       build: .
       command: ["hdfs", "datanode"]


matthewrossi:
Following this change, it would be nice if also the documentation located at [https://hub.docker.com/r/apache/hadoop](https://hub.docker.com/r/apache/hadoop) undergoes the same update.

ayushtkn:
do you know which commit broke this?

matthewrossi:
This is what I've found diving into the project history:

- `docker-compose.yaml` was always configured with `ENSURE_NAMENODE_DIR: "/tmp/hadoop-root/dfs/name"`
- the namenode [base image](https://github.com/apache/hadoop/blob/docker-hadoop-runner/Dockerfile) always specified the use of the `hadoop` user (so my initial assumption about the previous use of the `root` user was wrong)
- the default configurations of [Hadoop](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/resources/core-default.xml#L37) and [HDFS](https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/resources/hdfs-default.xml#L442) are the ones determining the use of the `/tmp/hadoop-${user.name}/dfs/name` directory,  but they date back before the creation of the `docker-compose.yaml`

So, it looks like the issue has always been there.

PR #4410, opened 2022-06-06
HDFS-16064. Determine when to invalidate corrupt replicas based on number of usable replicas

KevinWikant:
HDFS-16064. Determine when to invalidate corrupt replicas based on number of usable replicas

### Description of PR

Bug fix for a re-occurring HDFS bug which can result in datanodes being unable to complete decommissioning indefinitely. In short, the bug is a chicken & egg problem where:
- in order for a datanode to be decommissioned its blocks must be sufficiently replicated (which is not true if liveReplicas < "dfs.replication")
- datanode cannot sufficiently replicate some block(s) because of corrupt block replicas on target datanodes (i.e. https://issues.apache.org/jira/browse/HDFS-16064)
- corrupt block replicas will never be invalidated by the Namenode because the block(s) are not minimally replicated (because liveReplicas < "dfs.namenode.replication.min")

In this scenario, the block(s) are minimally replicated but the logic the Namenode uses to determine if a block is minimally replicated is flawed. Should be checking if usableReplicas < "dfs.namenode.replication.min" because decommissioning & entering maintenance nodes can have valid block replicas which should be replicated to the other datanodes with corrupt replicas if liveReplicas < "dfs.replication".

To understand the bug further we must first establish some background information.

#### Background Information

Givens:
- FSDataOutputStream is being used to write the HDFS file, under the hood this uses a class DataStreamer
- for the sake of example we will say the HDFS file has a replication factor of 2, though this is not a requirement to reproduce the issue
- the file is being appended to intermittently over an extended period of time (in general, this issue needs minutes/hours  to reproduce because corrupt replicas need to be generated over time)
- HDFS is configured with typical default configurations

Under certain scenarios the DataStreamer client can detect a bad link when trying to append to the block pipeline, in this case the DataStreamer client can shift the block pipeline by replacing the bad link with a new datanode. When this happens the replica on the datanode that was shifted away from becomes corrupted because it no longer has the latest generation stamp for the block. As a more concrete example:
- DataStreamer client creates block pipeline on datanodes A & B, each have a block replica with generation stamp 1
- DataStreamer client tries to append the block pipeline by sending block transfer (with generation stamp 2) to datanode A
- Datanode A succeeds in writing the block transfer & then attempts to forward the transfer to datanode B
- Datanode B fails the transfer for some reason and responds with a PipelineAck failure code
[...]

PR #6437, opened 2024-01-10
[HDFS-16064] backporting HDFS-16064. Determine when to invalidate corrupt replicas based on number of usable replicas (#4410)

zz12341:
### Description of PR

[HDFS-16064](https://github.com/apache/hadoop/pull/4410) fixed an issue where decommissioning replicas were not counted as usable replicas, which caused decom to stuck forever in the case of small clusters. We are seeing the same issue on 2.10, and thus backporting the changes. 

### For code changes:

- [ x] Does the title or this PR starts with the corresponding JIRA issue id (e.g. 'HADOOP-17799. Your PR title ...')?

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManager.java
+++ b/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManager.java
@@ -1463,8 +1463,12 @@ public class BlockManager implements BlockStatsMXBean {
         b.getReason(), b.getReasonCode());
 
     NumberReplicas numberOfReplicas = countNodes(b.getStored());
-    boolean hasEnoughLiveReplicas = numberOfReplicas.liveReplicas() >=
-        expectedReplicas;
+    final int numUsableReplicas = numberOfReplicas.liveReplicas() +
+        numberOfReplicas.decommissioning() +
+        numberOfReplicas.liveEnteringMaintenanceReplicas();
+    boolean hasEnoughLiveReplicas = numUsableReplicas >=
+        expectedRedundancies;
+
     boolean minReplicationSatisfied =
         numberOfReplicas.liveReplicas() >= minReplication;
     boolean hasMoreCorruptReplicas = minReplicationSatisfied &&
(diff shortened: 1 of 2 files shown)


hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManager.java
```diff
@@ -791,15 +791,33 @@ public short getMinReplication() {
     return minReplication;
   }
 
+  public short getMinStorageNum(BlockInfo block) {
```
shahrs87: @zz12341  Why we want these changes? I don't see it in the original patch [here](https://github.com/apache/hadoop/pull/4410/files).

zz12341: I was originally trying to make it in sync with what trunk branch is doing: https://github.com/apache/hadoop/pull/4410/files#diff-305ecf45a0f0708849b5e3c0d21a56c681db3a1497e52a19ef24939278dc99feL1922-R1926

Let me revert this change

hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManager.java
```diff
@@ -1463,21 +1463,28 @@ private void markBlockAsCorrupt(BlockToMarkCorrupt b,
...
-    boolean hasEnoughLiveReplicas = numberOfReplicas.liveReplicas() >=
-        expectedReplicas;
+    final int numUsableReplicas = numberOfReplicas.liveReplicas() +
+        numberOfReplicas.decommissioning() +
+        numberOfReplicas.liveEnteringMaintenanceReplicas();
+    boolean hasEnoughLiveReplicas = numUsableReplicas >=
+        expectedRedundancies;
+
     boolean minReplicationSatisfied =
         numberOfReplicas.liveReplicas() >= minReplication;
```
shahrs87, line 1473:
```suggestion
        numUsableReplicas >= minReplication;
```

shahrs87:
@zz12341  [branch-2.10](https://github.com/apache/hadoop/tree/branch-2.10) is the active branch for 2.10 (not [branch-2.10.0](https://github.com/apache/hadoop/tree/branch-2.10.0)). You can see all the active branches [here](https://github.com/apache/hadoop/branches).

shahrs87:
Screenshot 2024-01-11 at 9 45 49 AM
@zz12341  This change in `BlockManager#addStoredBlock` is missing from the original patch? https://github.com/apache/hadoop/pull/4410

shahrs87:
> [HDFS-16064] backporting HDFS-16064. Determine when to invalidate corrupt replicas based on number of usable replicas (#4410) #6437

@zz12341  Keep the title and the commit message same as https://github.com/apache/hadoop/pull/4410. We shouldn't add backporting keyword in the title and commit message.

PR #6785, opened 2024-04-30
MAPREDUCE-7475. Fixed non-idempotent unit tests

kaiyaok2:
## Description of PR
This PR fixes unit tests that are not idempotent and fail upon repeated execution within the same JVM instance due to self-induced state pollution. The tests shall be fixed as unit tests shall be idempotent, and it shall be good to clean the state pollution so that potential newly introduced tests do not fail in the future due to the shared state polluted by these tests.

## Overview & Proposed Fix of all non-idempotent unit tests in the MapReduce Module
The following tests try to make the directory `TEST_ROOT_DIR` and write to it. The tests do not clean up (remove) the directory after execution. Therefore, in the second execution, `TEST_ROOT_DIR` would already exist and the exception `Could not create test dir` would be thrown. 

* org.apache.hadoop.mapred.TestOldCombinerGrouping.testCombiner
* org.apache.hadoop.mapreduce.TestNewCombinerGrouping.testCombiner

## Sample Failure Message (in the 2nd run of the test):
```
java.lang.RuntimeException: Could not create test dir: /home/kaiyaok2/NIOExperiments/github.com/apache/hadoop/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/build/test/data/e4ee268d-2946-43a2-93b5-f2b4ac647279
        at org.apache.hadoop.mapreduce.TestNewCombinerGrouping.testCombiner(TestNewCombinerGrouping.java:109)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```
The fix is done by fully deleting the test directory after test execution.

-----------------------------------------------------------------------------------------------

The following two tests below do not reset `NotificationServlet.counter`, so repeated runs throw assertion failures due to accumulation. 

- org.apache.hadoop.mapred.TestClusterMRNotification#testMR
- org.apache.hadoop.mapred.TestLocalMRNotification#testMR

Fixed by resetting `NotificationServlet.counter` and `NotificationServlet.failureCounter` to 0 after test execution.

-----------------------------------------------------------------------------------------------

The following test does not remove the key `AMParams.ATTEMPT_STATE`, so repeated runs of the test will not be missing the attempt-state at all:

- org.apache.hadoop.mapreduce.v2.app.webapp.TestAppController.testAttempts

Fixed by removing `AMParams.ATTEMPT_STATE` at the end of the test.

-----------------------------------------------------------------------------------------------

[...]

PR #6790, opened 2024-05-03
MAPREDUCE-7476. Fixed more non-idempotent unit tests

kaiyaok2:
### Description of PR

Similar with https://issues.apache.org/jira/browse/MAPREDUCE-7475 , this PR fixes more non-idempotent unit tests detected.

## Overview & Proposed Fix of all remaining non-idempotent unit tests in the MapReduce Project

The following two tests below do not reset `NotificationServlet.counter`, so repeated runs throw assertion failures due to accumulation. 

- org.apache.hadoop.mapred.TestClusterMRNotification#testMR
- org.apache.hadoop.mapred.TestLocalMRNotification#testMR

Fixed by resetting `NotificationServlet.counter` and `NotificationServlet.failureCounter` to 0 after test execution.

-----------------------------------------------------------------------------------------------

The following test does not remove the key `AMParams.ATTEMPT_STATE`, so repeated runs of the test will not be missing the attempt-state at all:

- org.apache.hadoop.mapreduce.v2.app.webapp.TestAppController.testAttempts

Fixed by removing `AMParams.ATTEMPT_STATE` at the end of the test.

-----------------------------------------------------------------------------------------------

The following test fully deletes `TEST_ROOT_DIR` after execution, so repeated runs will throw a`DiskErrorException`:

- org.apache.hadoop.mapred.TestMapTask#testShufflePermissions

Fixed by checking if `TEST_ROOT_DIR` exists before test execution. Make the directory if not.

-----------------------------------------------------------------------------------------------

The following test does not restore the static variable `statusUpdateTimes` after execution, so consecutive runs throws `AssertionError`:

- org.apache.hadoop.mapred.TestTaskProgressReporter#testTaskProgress

Fixed by resetting `statusUpdateTimes` to 0 before test execution

After the patch, rerunning the tests in the same JVM does not produce any exceptions.

### DIFF
--- a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/test/java/org/apache/hadoop/mapreduce/v2/app/webapp/TestAppController.java
+++ b/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-app/src/test/java/org/apache/hadoop/mapreduce/v2/app/webapp/TestAppController.java
@@ -319,6 +319,8 @@ public class TestAppController {
     appController.attempts();
 
     assertEquals(AttemptsPage.class, appController.getClazz());
+
+    appController.getProperty().remove(AMParams.ATTEMPT_STATE);
   }
 
 }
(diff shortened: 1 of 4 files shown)


kaiyaok2:
@steveloughran PTAL

kaiyaok2:
Closing as code changes of this PR is merged to #6785

PR #6805, opened 2024-05-09
HDFS-17515. Erasure Coding: ErasureCodingWork is not effectively limi…

zhengchenyu:
### Description of PR

https://issues.apache.org/jira/browse/HDFS-17515

unit test

### For code changes:

- Add pendingECBlockReplicationWithoutTargets for DatanodeDescriptor
- When construct ErasureCodingWork, increase pendingECBlockReplicationWithoutTargets. When set targets, decrease pendingECBlockReplicationWithoutTargets.
- Because we can not decide the real source datanode when scheduleReconstruction, so update pendingECBlockReplicationWithoutTargets for all source datanode. So when we calculate getNumberOfBlocksToBeReplicated, use a factor to appropriately lower the value.


zhengchenyu:
Not ready for review! Wait for HDFS-17516!

PR #7046, opened 2024-09-16
HADOOP-19276. Create hadoop-runner based on JDK 17

adoroszlai:

Publish another version of `hadoop-runner` image with JDK 17 instead of JDK 11. It can be used for testing JDK 17 runtime compatibility.

(Creating as a draft, because the change shouldn't be merged to the Java 11-specific branch, rather pushed directly as a new branch.)

https://issues.apache.org/jira/browse/HADOOP-19276

## How was this patch tested?

Built locally, verified Java version:

```
$ docker build -t apache/hadoop-runner:dev --no-cache .
...

$ docker run -it --rm apache/hadoop-runner:dev java -version
openjdk version "17.0.11" 2024-04-16
OpenJDK Runtime Environment Temurin-17.0.11+9 (build 17.0.11+9)
OpenJDK 64-Bit Server VM Temurin-17.0.11+9 (build 17.0.11+9, mixed mode, sharing)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG JAVA_VERSION=11
+ARG JAVA_VERSION=17
 
 # Ubuntu 22.04 LTS
 FROM eclipse-temurin:${JAVA_VERSION}-jammy


steveloughran:
I agree, needs to be separate
I would like to move hadoop trunk to java11+ ASAP. which would let us drop the java8 run, just do java11 + 17.

ayushtkn:
This seems to drop JAVA-11 in favour of JAVA-17, If I am catching it right? Wouldn't it make sense to have JAVA-17 without dropping JAVA-11?

adoroszlai:
@ayushtkn PR is created only to show the diff from Java 11-based branch.  If approved, I'll push to a new branch `docker-hadoop-runner-jdk17-u2204` to create a separate image with Java 17.  That's why the PR is draft, to avoid getting it merged.

ayushtkn (approved):
In that case, I am all in

I just got confused with the description in the JIRA, instead, so I thought you want to replace it :-) 
```
Publish hadoop-runner image with JDK 17 instead of JDK 11
```

adoroszlai:
> I just got confused with the description in the JIRA, instead, so I thought you want to replace it :-)

Sorry about that, updated the description.

adoroszlai:
Thanks for the reviews.

[`apache/hadoop-runner:jdk17-u2204`](https://hub.docker.com/layers/apache/hadoop-runner/jdk17-u2204/images/sha256-c482147c40feb0953c9b83ff31852c4633c108f357e35a5b6d17722c6c580b53?context=explore) published by pushing to branch [`docker-hadoop-runner-jdk17-u2204`](https://github.com/apache/hadoop/tree/docker-hadoop-runner-jdk17-u2204).

steveloughran:
@ayushtkn I think we should
* move trunk to java 11 only
* plan for a 3.5.0 release in december/january, RCs starting nov as the "java11 release" + hbase 2 ATS, plus some changes in s3a create/rename behaviour which may require third party stores to add some new options unless they support the S3 conditional put
* also lets us declare 3.3.x EOL.

ayushtkn:
trunk would be moving to JDK-17 I believe next month, the only major roadblock was jersey & I see there was a green build on that PR 2 days back (it broke 1 or 2 tests in attempt to fix checkstyle later)
https://github.com/apache/hadoop/pull/7046#issuecomment-2360877162

So, we might move to JDK-17 directly, I think this thread even intended to do so: https://lists.apache.org/thread/fbcdfzjtj77x7hf8znjj598hrt34j6xz

PR #7795, opened 2025-07-11
HADOOP-19614. Publish Hadoop Docker Image to DockerHub for amd64 and arm64

kumaab:
### Description of PR
Update GitHub Actions workflow to publish the `apache/hadoop` Docker image to GitHub Container Registry and DockerHub for multiple architectures: `amd64` and `arm64`.

### Proposal:
The automation will work only for branches like docker-hadoop-X.Y, where the image will be tagged as X.Y.

a git tag, for ex: `git tag docker-image-3.4.1` and push will result in `tag` job being run and upon whose successful run, the image will also be tagged `3.4.1` and will be published to DockerHub.

Pull request only triggers build (for CI), but image is not published.

https://issues.apache.org/jira/browse/HADOOP-19614

- build step is executing successfully for `amd64` and `arm64` flavors: https://github.com/apache/hadoop/actions/runs/16209022572/job/45765483246

- tag step is untested

### For code changes:


smengcl:
> tag step is untested

Is it possible to test it on your own fork and Docker Hub (registry), just to verify if it works?

kumaab:
> > tag step is untested
> 
> Is it possible to test it on your own fork and Docker Hub (registry), just to verify if it works?

Thanks @smengcl for the suggestion, I tried running in my fork - it runs indeed after committing changes to trunk in my fork. However, I'm running into issues in getting the `tag` job to run, `tag` job is currently skipped and a successful `tag` push via `git push origin <tag_name>` doesn't seem to trigger the job, investigating the cause currently, thanks.

jesusch:
Hi,

I am a bit confused - with hadoop-3.4.1 there seem to be no more aarch64 binaries on the apache servers?

.github/workflows/build-and-tag-hadoop-image.yaml
```diff
@@ -0,0 +1,138 @@
...
+        uses: docker/build-push-action@4f58ea79222b3b9dc2c8bbdd6debcef730109a75
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+  tag:
+    needs: build
+    if: ${{ github.ref_type == 'tag' }}
```
adoroszlai, line 90: Hadoop images are tagged based on branch name, not by explicit Git tags.  Images built for the branch `docker-hadoop-3.4.1` get published with Docker tag `3.4.1`.  Thus the condition `ref_type == 'tag'` should be removed.

.github/workflows/build-and-tag-hadoop-image.yaml
```diff
@@ -0,0 +1,138 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate image ID
+        id: meta
+        uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81
+        with:
+          images: |
+            ghcr.io/${{ github.repository_owner }}/hadoop
+          tags: |
+            type=match,pattern=docker-hadoop-(.*),value={{branch}},group=1
```
adoroszlai, line 47: This image ID (which will be checked for existence) should be generated based on commit SHA, not branch name.  Otherwise it will not be possible to build new image for the same branch when changes are needed.

adoroszlai (commented):
Thanks @kumaab for working on this.  Instead of creating a new workflow, please update the existing one (`build-hadoop-image.yaml`).

PR #8116, opened 2025-12-02
HADOOP-19747. Switch to at.yawk.lz4:lz4-java:1.9.0 due to CVE-2025-12183

pjfanning:
### Description of PR

HADOOP-19747

### For code changes:

- [x] If applicable, have you updated the `LICENSE`, `LICENSE-binary`, `NOTICE-binary` files?


steveloughran (approved):
+1

> In the long term, it is recommended to switch to .safeDecompressor(), which is not vulnerable and provides better performance (despite the name).

pjfanning:
@steveloughran I think Hadoop uses safe decompressor already

https://github.com/apache/hadoop/blob/a01e5fd96a926a5c6eb88b003d50951010588eeb/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/io/compress/lz4/Lz4Decompressor.java#L61

slfan1989 (approved):

pjfanning:
@steveloughran do you think it is ok to backport this to the 3.4 branch?

pan3793:
Just FYI, lz4 is famous for its ultra-fast speed, the upgrade is not free, my test shows it has perf impact 

- https://github.com/apache/spark/pull/53453

PR #8415, opened 2026-04-09
HDFS-17905. Clean up pendingDNMessages on block removal

kokonguyen191:
### Description of PR

Since DNs do not report deleted blocks to NNs, in standby and observer namenodes, if not removed, `pendingDNMessages` will never drain messages from blocks that have already been deleted, leading to a leak.

image

The solution is to clean up the messages when the block is removed in `BlockManager#removeBlock()`.

UT


ZanderXu (approved):
Nice catch. LGTM + 1.

Please take a look at Jenkins

ZanderXu:
Merged. Thanks @kokonguyen191 for your contribution.

PR #8463, opened 2026-04-28
HDFS-17914. Prevent DFSInputStream from issuing 0-byte reads

lirui-apache:
### Description of PR
Avoid empty buffer read in DFSInputStream#actualGetFromOneDataNode by terminating the loop once the buffer is filled.

Added new test

### For code changes:

### AI Tooling

If an AI tool was used:

### DIFF
--- a/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/DFSInputStream.java
+++ b/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/DFSInputStream.java
@@ -1230,7 +1230,9 @@ public class DFSInputStream extends FSInputStream
         long beginReadMS = Time.monotonicNow();
         int nread = 0;
         int ret;
-        while (true) {
+        // Stop once the slice is filled; an extra read with remaining()==0
+        // can trigger wasted slow-lane I/O in SCR.
+        while (tmp.hasRemaining()) {
           ret = reader.read(tmp);
           if (ret <= 0) {
             break;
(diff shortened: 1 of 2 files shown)


hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/DFSInputStream.java
```diff
@@ -1230,7 +1230,9 @@ void actualGetFromOneDataNode(final DNAddrPair datanode, final long startInBlk,
         long beginReadMS = Time.monotonicNow();
         int nread = 0;
         int ret;
-        while (true) {
+        // Stop once the slice is filled; an extra read with remaining()==0
+        // can trigger wasted slow-lane I/O in SCR.
+        while (tmp.hasRemaining()) {
```
Hexiaoqiao, line 1235: Thanks for your contribution first. Just one nit concern, is it only one difference that return at L1235(after PR) or return at L1236(before PR)? It will perform a zero-byte transfer in-memory and return immediately if tmp.remaining()=0, So it is OK for me before PR. welcome discussion if I miss something. Thanks again.

lirui-apache: Thanks for the review. I hit the issue when I was testing short circuit read. If we call `BlockReaderLocal.read` with `tmp.remaining() == 0`, it will hit the slow lane via `BlockReaderLocal::readWithBounceBuffer::fillDataBuf`, which will actually allocate and read into the bounce buffer. But since user buffer is filled, we cannot drain into it and the read data is thrown away and wasted.

Hexiaoqiao: Great catch. Make sense to me.
Just wonder how you hit this corner case, and why `tmp.hasRemaining==0` here?

lirui-apache: I was doing aligned positional reads with a DirectByteBuffer. Ideally I should only hit the fast lane and avoid the bounce buffer. But I noticed some cost from the slow lane and found the extra read. I believe the issue is pretty easy to trigger, but whether it brings real perf regression depends on the BlockReader implementations.

PR #8469, opened 2026-05-05
HADOOP-19878. Upgrade to Netty 4.1.133.Final due to CVEs

pjfanning:
### Description of PR

https://issues.apache.org/jira/browse/HADOOP-19878

### For code changes:

- [x] If applicable, have you updated the `LICENSE`, `LICENSE-binary`, `NOTICE-binary` files?

### AI Tooling

If an AI tool was used:


steveloughran (approved):
+1 pending the build.

@pjfanning we really an upgrade marathon don't we, where everything which can be upgraded is (trunk, and 3.5) and anything which still works on java8 is done for the 3.4 branch. Then we can think of new releases

ajfabbri (approved):
+1 pending jenkins/pr-merge CI passing. We can ignore the `PR update / Build (pull_request_target) — Workflow run detection failed` check for now: We are adding new github actions-based CI which requires enabling actions in your forked repo to get past that error, but it is not a requirement yet.

LICENSE-binary
```diff
@@ -309,36 +309,36 @@ io.grpc:grpc-stub:1.69.0
...
+io.netty:netty-all:4.1.133.Final
+io.netty:netty-buffer:4.1.133.Final
+io.netty:netty-codec:4.1.133.Final
+io.netty:netty-codec-dns:4.1.133.Final
+io.netty:netty-codec-haproxy:4.1.133.Final
+io.netty:netty-codec-http:4.1.133.Final
+io.netty:netty-codec-http2:4.1.133.Final
+io.netty:netty-codec-memcache:4.1.133.Final
+io.netty:netty-codec-mqtt:4.1.133.Final
+io.netty:netty-codec-redis:4.1.133.Final
```
pan3793, line 321: are all those netty libs used by hadoop? I don't think hadoop requires `netty-codec-redis` ...

pjfanning: it feels like there is a separate task to revisit the entire LICENSE-binary file - I guarantee that this is not the only is it/isn't it correct line in the file

pan3793:
@pjfanning could you follow the instructions to ensure GHA on your forked repo?

https://github.com/apache/hadoop/pull/8469/checks?check_run_id=74443729055

pan3793:
@steveloughran do we want to land this in branch-3.4? HADOOP-19788 upgraded Netty from 4.1.127.Final to 4.1.130.Final, which does not land branch-3.4, we need both or neither.

pan3793:
merged to trunk/branch-3.5 for now. branch-3.4 backport requires making a decision first.

steveloughran:
i don't see any specific reason for holding back the 3.4 update, so go with it

pan3793:
backported to branch-3.4

PR #8524, opened 2026-06-01
HDFS-17929. Revisit retrying mechanism in DFSStripedInputStream

CapMoon:
### Description of PR
DFSStripedInputStream#createBlockReader handles reader creation failures by re-fetching the block location, because the striped internal block may have moved. However, after re-fetching it unconditionally calls addToLocalDeadNodes().

There is also a related retry issue in DFSStripedInputStream#readWithStrategy. The method retries after an IOException, but the next blockSeekTo() can still use cached LocatedBlocks. Without refreshing open file information before retrying, the retry may reuse stale block locations and hit the same moved-block condition again.

Unit tests


CapMoon:
@ZanderXu @hfutatzhanghb masters, can take a look at this mr when you are available

hfutatzhanghb:
> @ZanderXu @hfutatzhanghb masters, can take a look at this mr when you are available

Thanks for reporting and involving me here. Will review this laterly.

PR #8598, opened 2026-07-13
HDFS-16512. Improve oiv tool to parse fsimage file in parallel with XML format

YuanbenWang:
JIRA: [HDFS-16512](https://issues.apache.org/jira/browse/HDFS-16512). Improve oiv tool to parse fsimage file in parallel with XML format


Hexiaoqiao:
Thanks @YuanbenWang for your contribution. I will give my +1 for this PR source code. One concerns, if there will be improvement when parse fsimage parallel? it is unnecessary for small fsimage, and IO time cost to merge for large fsimage. Do you mind to share some bench data? Thus it will be better to evaluate. Thanks again.

PR #8606, opened 2026-07-14
HADOOP-15190. Replace Clover with JaCoCo for code coverage

ferdelyi:
Trunk supports Java 17 (https://github.com/apache/hadoop-release-support). OpenClover (the open-sourced fork of Atlassian Clover) is incompatible with Java 17+ because it relies on source-level instrumentation via a javac annotation processor that triggers split-package violations in the module system.  JaCoCo uses bytecode instrumentation via a Java agent and has no such constraint.

### Description of PR
Changes:
- pom.xml: remove com.cenqua.clover dependency management, replace clover-maven-plugin in pluginManagement with jacoco-maven-plugin 0.8.15, add the hadoop-coverage aggregate module, drop the <profile id="clover"> block entirely, and declare
  hadoop.skip-jacoco=true here so modules that inherit directly from hadoop-main (e.g. hadoop-coverage) resolve the property too
- hadoop-project/pom.xml: add hadoop.skip-jacoco=true property (opt-in by default), append @{argLine} to the surefire and failsafe argLines so the JaCoCo agent is injected into every forked test JVM, and wire in jacoco:prepare-agent with excludes for generated/proto/example classes so they are not instrumented
- per-module poms that override the surefire argLine (hadoop-common, hadoop-hdfs, hadoop-hdfs-rbf, hadoop-registry, hadoop-aws, hadoop-azure, hadoop-distcp, hadoop-federation-balance, hadoop-gcp,
  hadoop-tos, hadoop-yarn-applications-catalog-webapp): append @{argLine} so instrumentation is not dropped
- hadoop-coverage/pom.xml (new): aggregate module that depends on the test-bearing modules and runs jacoco:report-aggregate at the verify phase to produce a combined HTML/XML report under
  hadoop-coverage/target/site/jacoco-aggregate/; deploy and install are skipped so this coverage-only pom is never published
- hadoop-maven-plugins/pom.xml, hadoop-yarn-server-nodemanager/pom.xml: remove stale clover skip
  overrides that are no longer needed
- test-libhdfs.sh, hadoop-mapreduce-client-jobclient/pom.xml, hadoop-yarn-applications/pom.xml: drop remaining clover references

report-aggregate only sees modules within the same Maven reactor, so the report must be produced by a single full-reactor build (not with -pl hadoop-coverage, which would leave the other modules out of the
reactor and yield an empty report):

  mvn verify -Dhadoop.skip-jacoco=false -Dmaven.test.failure.ignore=true

The aggregate report lands at hadoop-coverage/target/site/jacoco-aggregate/index.html.

Generated-by: Claude Code

Run on my local computer coverage report for hadoop_nfs and also an aggregate coverage report.
Screenshots provided on the Jira as attachments. 

### For code changes:

### AI Tooling

If an AI tool was used:

- [x] The PR includes the phrase "Contains content generated by <tool>"

### DIFF
--- a/BUILDING.txt
+++ b/BUILDING.txt
@@ -149,9 +149,18 @@ Maven build goals:
  * Run checkstyle            : mvn compile checkstyle:checkstyle
  * Install JAR in M2 cache   : mvn install
  * Deploy JAR to Maven repo  : mvn deploy
- * Run JaCoCo coverage       : mvn verify -Dhadoop.skip-jacoco=false -Dmaven.test.failure.ignore=true
+ * Run JaCoCo coverage       : mvn verify -Dhadoop.skip-jacoco=false -Dmaven.test.failure.ignore=true --fail-at-end
                                (aggregate report: hadoop-coverage/target/site/jacoco-aggregate/index.html;
-                                must be a full-reactor build, not -pl hadoop-coverage)
+                                must be a full-reactor build, not -pl hadoop-coverage.
+                                --fail-at-end keeps the reactor building/testing the modules
+                                that don't depend on a hard-failing one (each writing its
+                                target/jacoco.exec) instead of stopping at the first failure;
+                                hadoop-coverage does not depend on the catalog-webapp war, so
+                                it still runs report-aggregate at the verify phase.
+                                -Dmaven.test.failure.ignore=true only tolerates surefire *test*
+                                failures; a plugin-level hard failure (e.g. the catalog-webapp
+                                jasmine/PhantomJS browser test in a headless container) would
+                                otherwise stop the reactor before the report is produced.)
  * Run Rat                   : mvn apache-rat:check
  * Build javadocs            : mvn javadoc:javadoc
  * Build distribution        : mvn package [-Pdist][-Pdocs][-Psrc][-Pnative][-Dtar][-Preleasedocs][-Pyarn-ui]
--- a/BUILDING.txt
+++ b/BUILDING.txt
@@ -149,18 +149,8 @@ Maven build goals:
  * Run checkstyle            : mvn compile checkstyle:checkstyle
  * Install JAR in M2 cache   : mvn install
  * Deploy JAR to Maven repo  : mvn deploy
- * Run JaCoCo coverage       : mvn verify -Dhadoop.skip-jacoco=false -Dmaven.test.failure.ignore=true --fail-at-end
-                               (aggregate report: hadoop-coverage/target/site/jacoco-aggregate/index.html;
-                                must be a full-reactor build, not -pl hadoop-coverage.
-                                --fail-at-end keeps the reactor building/testing the modules
-                                that don't depend on a hard-failing one (each writing its
-                                target/jacoco.exec) instead of stopping at the first failure;
-                                hadoop-coverage does not depend on the catalog-webapp war, so
-                                it still runs report-aggregate at the verify phase.
-                                -Dmaven.test.failure.ignore=true only tolerates surefire *test*
-                                failures; a plugin-level hard failure (e.g. the catalog-webapp
-                                jasmine/PhantomJS browser test in a headless container) would
-                                otherwise stop the reactor before the report is produced.)
+ * Run JaCoCo coverage       : mvn verify -Djacoco.skip=false -Dmaven.test.failure.ignore=true --fail-at-end
+                               (see 'Special plugins: JaCoCo code coverage' below)
  * Run Rat                   : mvn apache-rat:check
  * Build javadocs            : mvn javadoc:javadoc
  * Build distribution        : mvn package [-Pdist][-Pdocs][-Psrc][-Pnative][-Dtar][-Preleasedocs][-Pyarn-ui]
@@ -270,6 +260,24 @@ Maven build goals:
    of this project for known CVEs (security vulnerabilities against them).
    It will produce a report in target/dependency-check-report.html. To
    invoke, run 'mvn dependency-check:aggregate'.
+
+ Special plugins: JaCoCo code coverage:
+
+   Coverage is opt-in and off by default. Enable it with -Djacoco.skip=false
+   (jacoco.skip is the jacoco-maven-plugin's own skip property); the aggregate
+   report is written to hadoop-coverage/target/site/jacoco-aggregate/index.html.
+
+   It must be a full-reactor build (not -pl hadoop-coverage): report-aggregate
+   only sees modules that are part of the same reactor.
+
+   --fail-at-end keeps the reactor building/testing the modules that do not
+   depend on a hard-failing one (each writing its target/jacoco.exec) instead
+   of stopping at the first failure, so hadoop-coverage still runs
+   report-aggregate at the verify phase. -Dmaven.test.failure.ignore=true only
+   tolerates surefire *test* failures; a plugin-level hard failure (e.g. the
+   catalog-webapp jasmine/PhantomJS browser test, which cannot start a browser
+   in a headless container) would otherwise stop the reactor before the report
+   is produced.
 
  PMDK library build options:
 
--- a/BUILDING.txt
+++ b/BUILDING.txt
@@ -149,7 +149,9 @@ Maven build goals:
  * Run checkstyle            : mvn compile checkstyle:checkstyle
  * Install JAR in M2 cache   : mvn install
  * Deploy JAR to Maven repo  : mvn deploy
- * Run clover                : mvn test -Pclover
+ * Run JaCoCo coverage       : mvn verify -Dhadoop.skip-jacoco=false -Dmaven.test.failure.ignore=true
+                               (aggregate report: hadoop-coverage/target/site/jacoco-aggregate/index.html;
+                                must be a full-reactor build, not -pl hadoop-coverage)
  * Run Rat                   : mvn apache-rat:check
  * Build javadocs            : mvn javadoc:javadoc
  * Build distribution        : mvn package [-Pdist][-Pdocs][-Psrc][-Pnative][-Dtar][-Preleasedocs][-Pyarn-ui]
--- a/BUILDING.txt
+++ b/BUILDING.txt
@@ -150,7 +150,6 @@ Maven build goals:
  * Install JAR in M2 cache   : mvn install
  * Deploy JAR to Maven repo  : mvn deploy
  * Run JaCoCo coverage       : mvn verify -Djacoco.skip=false -Dmaven.test.failure.ignore=true --fail-at-end
-                               (see 'Special plugins: JaCoCo code coverage' below)
  * Run Rat                   : mvn apache-rat:check
  * Build javadocs            : mvn javadoc:javadoc
  * Build distribution        : mvn package [-Pdist][-Pdocs][-Psrc][-Pnative][-Dtar][-Preleasedocs][-Pyarn-ui]
--- a/hadoop-cloud-storage-project/hadoop-gcp/pom.xml
+++ b/hadoop-cloud-storage-project/hadoop-gcp/pom.xml
@@ -88,7 +88,7 @@
               <forkCount>${testsThreadCount}</forkCount>
               <reuseForks>false</reuseForks>
               <trimStackTrace>false</trimStackTrace>
-              <argLine>${maven-surefire-plugin.argLine} -DminiClusterDedicatedDirs=true</argLine>
+              <argLine>${maven-surefire-plugin.argLine} @{argLine} -DminiClusterDedicatedDirs=true</argLine>
               <systemPropertyVariables>
                 <testsThreadCount>${testsThreadCount}</testsThreadCount>
                 <test.build.data>${test.build.data}/${surefire.forkNumber}</test.build.data>
@@ -111,7 +111,7 @@
                 <configuration>
                   <forkCount>${testsThreadCount}</forkCount>
                   <reuseForks>false</reuseForks>
-                  <argLine>${maven-surefire-plugin.argLine} -DminiClusterDedicatedDirs=true</argLine>
+                  <argLine>${maven-surefire-plugin.argLine} @{argLine} -DminiClusterDedicatedDirs=true</argLine>
                   <forkedProcessTimeoutInSeconds>${fs.gs.scale.test.timeout}</forkedProcessTimeoutInSeconds>
                   <trimStackTrace>false</trimStackTrace>
                   <systemPropertyVariables>
--- a/hadoop-coverage/pom.xml
+++ b/hadoop-coverage/pom.xml
@@ -1,0 +1,461 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License. See accompanying LICENSE file.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                      http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.hadoop</groupId>
+    <artifactId>hadoop-main</artifactId>
+    <version>3.6.0-SNAPSHOT</version>
+    <relativePath>..</relativePath>
+  </parent>
+  <groupId>org.apache.hadoop</groupId>
+  <artifactId>hadoop-coverage</artifactId>
+  <version>3.6.0-SNAPSHOT</version>
+  <description>Hadoop JaCoCo Aggregate Coverage</description>
+  <name>Hadoop JaCoCo Aggregate Coverage</name>
+  <packaging>pom</packaging>
+
+  <properties>
+    <!-- Coverage-only helper module; never publish it to a Maven repository. -->
+    <maven.deploy.skip>true</maven.deploy.skip>
+    <maven.install.skip>true</maven.install.skip>
+  </properties>
+
+  <!--
+    This module collects JaCoCo exec data from all sub-modules and produces
+    an aggregate HTML/XML coverage report under target/site/jacoco-aggregate/.
+
+    The report-aggregate goal only sees sibling modules that are part of the
+    same Maven reactor, so it must be run over the full project (not with
+    "-pl hadoop-coverage", which would leave every other module out of the
+    reactor and produce an empty report). Run tests and the report in a
+    single reactor pass:
+
+      mvn verify -Dhadoop.skip-jacoco=false -Dmaven.test.failure.ignore=true
+
+    hadoop-coverage builds last (it depends on every module), so the aggregate
+    report is generated after all tests have run. -Dmaven.test.failure.ignore=true
+    lets the report be produced even when some tests fail.
+
+    The aggregate report will be at:
+      hadoop-coverage/target/site/jacoco-aggregate/index.html
+  -->
+
+  <dependencies>
+    <!-- hadoop-common-project -->
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-annotations</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-auth</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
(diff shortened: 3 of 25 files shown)


hadoop-coverage/pom.xml
```diff
@@ -0,0 +1,461 @@
...
+  <parent>
+    <groupId>org.apache.hadoop</groupId>
+    <artifactId>hadoop-main</artifactId>
+    <version>3.6.0-SNAPSHOT</version>
+    <relativePath>..</relativePath>
+  </parent>
+  <groupId>org.apache.hadoop</groupId>
+  <artifactId>hadoop-coverage</artifactId>
+  <version>3.6.0-SNAPSHOT</version>
+  <description>Hadoop JaCoCo Aggregate Coverage</description>
```
K0K0V0K: I think the **artifact-id** and the **description** may get some problem in future if we will have to migrate to a next coverage lib. In that case maybe the best option will be to create a new project like this like 'Hadoop xy Aggregate Coverage' and than the previous hadoop-coverage name could be confusing ...

What do you think?
- Can we remove the JaCoCo from the description?
- or can we rename the artifact to hadoop-jacoco-coverage?

hadoop-coverage/pom.xml
```diff
@@ -0,0 +1,461 @@
...
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-yarn-applications-mawo-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-yarn-services-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+  </dependencies>
```
K0K0V0K, line 440: If i see correctly now if a new artifact created in the project it should be added here also.
I think this can be a part of the required changes what can be easily missed.
Is there any other way how we can prevent to list every artifact here?

K0K0V0K (commented):
Thanks @ferdelyi for this fix!

Overall seems good, i just had some question.
Also can you please check other parts of the code where clover is mentioned?
If i see correctly the BUILDING.txt still has a line like
 * Run clover                : mvn test -Pclover
 
 UPDATE:
 
 - sry, i checked wrong code, now i see BUILDING.txt is updated also

hadoop-coverage/pom.xml
ferdelyi: Thank you for your review @K0K0V0K , removed the JaCoCo from the description.

hadoop-coverage/pom.xml
ferdelyi: Great point @K0K0V0K , extended the PR with a build-time guard.

K0K0V0K: Thanks, @ferdelyi! This looks much better.

I noticed a very similar solution in hadoop-client-check-test-invariants. I think it would be nice to move the shell script into the new module as well, so the related code is kept together.

K0K0V0K (approved):
Thanks @ferdelyi for the work!

I tested out locally in the ubuntu24 docker container, with
`mvn org.jacoco:jacoco-maven-plugin:report -pl hadoop-yarn-project/hadoop-yarn/hadoop-yarn-common`
Seems working fine.

LGTM!

ferdelyi:
Thank you for your review and approval @K0K0V0K extended the PR with the change that coverage aggregate module joins the reactor only when coverage is explicitly requested. Please see the latest commit.

brumi1024 (approved):
Thanks @ferdelyi for the patch and @K0K0V0K for the review. LGTM.

ferdelyi:
@brumi1024 thank you for the review and the approve!

BUILDING.txt
```diff
@@ -149,7 +149,18 @@ Maven build goals:
...
+                                must be a full-reactor build, not -pl hadoop-coverage.
+                                --fail-at-end keeps the reactor building/testing the modules
+                                that don't depend on a hard-failing one (each writing its
+                                target/jacoco.exec) instead of stopping at the first failure;
+                                hadoop-coverage does not depend on the catalog-webapp war, so
+                                it still runs report-aggregate at the verify phase.
+                                -Dmaven.test.failure.ignore=true only tolerates surefire *test*
+                                failures; a plugin-level hard failure (e.g. the catalog-webapp
+                                jasmine/PhantomJS browser test in a headless container) would
+                                otherwise stop the reactor before the report is produced.)
```
pan3793: can you keep the single-line command here, and move this section below? e.g., after `Special plugins: OWASP's dependency-check:` section

hadoop-project/pom.xml
```diff
@@ -2627,6 +2618,33 @@
...
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>prepare-agent</id>
+            <goals>
+              <goal>prepare-agent</goal>
+            </goals>
+            <configuration>
+              <skip>${hadoop.skip-jacoco}</skip>
```
pan3793: the plugin already exposes a property `jacoco.skip`, why do we need a custom one?

hadoop-cloud-storage-project/hadoop-gcp/pom.xml
```diff
@@ -88,7 +88,7 @@
               <forkCount>${testsThreadCount}</forkCount>
               <reuseForks>false</reuseForks>
               <trimStackTrace>false</trimStackTrace>
-              <argLine>${maven-surefire-plugin.argLine} -DminiClusterDedicatedDirs=true</argLine>
+              <argLine>${maven-surefire-plugin.argLine} @{argLine} -DminiClusterDedicatedDirs=true</argLine>
```
pan3793, line 91: how does the `@{argLine}` work? I didn't find where it is defined, and who does the replacement.

hadoop-project/pom.xml
```diff
@@ -203,7 +206,6 @@
     <maven-pdf-plugin.version>1.2</maven-pdf-plugin.version>
     <maven-remote-resources-plugin.version>1.5</maven-remote-resources-plugin.version>
     <build-helper-maven-plugin.version>1.9</build-helper-maven-plugin.version>
-    <exec-maven-plugin.version>1.3.1</exec-maven-plugin.version>
```
pan3793, line 206: Why is this change required? If it's necessary, can you do this upgrade in an independent PR first?

BUILDING.txt
ferdelyi: @pan3793 thank you for your review! The single-line command in the "Maven build goals" list is restored and moved the details into a new "Special plugins: JaCoCo code coverage" section right after the OWASP dependency-check section.

hadoop-project/pom.xml
ferdelyi: @pan3793 thank you for this question. The custom hadoop.skip-jacoco was originally meant as a single Hadoop-owned switch covering the whole coverage feature — both the plugin executions and whether the hadoop-coverage aggregate module joins the reactor — while leaving ad-hoc jacoco:report runs working without a flag. But you're right that a bespoke property isn't worth it when the plugin already exposes jacoco.skip. I've switched to jacoco.skip (default true in the root pom so hadoop-coverage resolves it too; enable with -Djacoco.skip=false) and dropped the now-redundant explicit <skip> elements. The only behavior change is that coverage is now opt-in everywhere, so a bare mvn jacoco:report is skipped unless -Djacoco.skip=false is passed.

hadoop-cloud-storage-project/hadoop-gcp/pom.xml
ferdelyi: @{...} is Surefire/Failsafe late property replacement; jacoco:prepare-agent sets the argLine property to the -javaagent:… value at runtime, and Surefire substitutes @{argLine} when it forks.

hadoop-project/pom.xml
ferdelyi: To clarify, this was never an upgrade — 1.3.1 is unchanged. The PR had relocated exec-maven-plugin's version management from hadoop-project up to the root pom: hadoop-coverage's module-list guard runs via exec-maven-plugin, and since that module inherits from hadoop-main (not hadoop-project) it couldn't see hadoop-project's management, so it needed the version at the root — which made the hadoop-project entry redundant, hence removed.

Since then I've reworked it so no shared poms are touched at all: root pom.xml and hadoop-project/pom.xml are back to trunk, and the version is now pinned locally in hadoop-coverage/pom.xml itself. So there's no longer any shared or unrelated change, and nothing that needs a separate PR.

pan3793:
Two issues, both in `hadoop-bos`. It appears to have been missed while its sibling `hadoop-tos` was handled correctly.

## Bug 1 (build-breaking): `hadoop-bos` missing from the coverage aggregate

`hadoop-bos` is a test-bearing jar module (`<packaging>jar</packaging>` + `src/test/java/`), but it is listed in neither `hadoop-coverage/pom.xml`'s `<dependencies>` (the cloud-storage section lists `hadoop-cos`, `hadoop-gcp`, `hadoop-huaweicloud`, `hadoop-tos`, but not `hadoop-bos`) nor `coverage-modules-allowlist.txt`.

Running this PR's own `check-coverage-modules.sh` against the current tree fails at `hadoop-coverage`'s `validate` phase:

```
check-coverage-modules: FAILED
The following test-bearing modules are missing from the coverage aggregate:
  - hadoop-bos
```

So the documented command (`mvn verify -Djacoco.skip=false ...`) fails at `hadoop-coverage` before `report-aggregate` ever runs, until `hadoop-bos` is added as a `<dependency>` in `hadoop-coverage/pom.xml`.

## Bug 2 (silent coverage gap): `hadoop-bos` surefire argLine missing `@{argLine}`

Even after Bug 1 is fixed, `hadoop-cloud-storage-project/hadoop-bos/pom.xml:265` overrides surefire's `<argLine>` without `@{argLine}`:

```xml
<argLine>${maven-surefire-plugin.argLine}</argLine>
```

A module-level `<argLine>` fully replaces the parent's, so the `@{argLine}` token that `jacoco:prepare-agent` substitutes to inject the agent is dropped. The forked test JVMs for `hadoop-bos` run without the JaCoCo agent, producing no exec data, so its coverage shows as empty in the aggregate report, silently. The guard script can't catch this; it only checks deps/allowlist presence.

Of the 21 surefire/failsafe `<argLine>` overrides in the repo, this PR patched 20; `hadoop-bos` is the only one missed.

Fix for both:

```xml
<argLine>${maven-surefire-plugin.argLine} @{argLine}</argLine>
```

plus a `hadoop-bos` `<dependency>` entry in `hadoop-coverage/pom.xml`.

## Minor

The reorder of `maven-resources-plugin` after `exec-maven-plugin` in `pluginManagement` (`hadoop-project/pom.xml`) is cosmetic only; `pluginManagement` order has no effect in Maven.

ferdelyi:
> Two issues, both in `hadoop-bos`. It appears to have been missed while its sibling `hadoop-tos` was handled correctly.
> 
> ## Bug 1 (build-breaking): `hadoop-bos` missing from the coverage aggregate
> `hadoop-bos` is a test-bearing jar module (`<packaging>jar</packaging>` + `src/test/java/`), but it is listed in neither `hadoop-coverage/pom.xml`'s `<dependencies>` (the cloud-storage section lists `hadoop-cos`, `hadoop-gcp`, `hadoop-huaweicloud`, `hadoop-tos`, but not `hadoop-bos`) nor `coverage-modules-allowlist.txt`.
> 
> Running this PR's own `check-coverage-modules.sh` against the current tree fails at `hadoop-coverage`'s `validate` phase:
> 
> ```
> check-coverage-modules: FAILED
> The following test-bearing modules are missing from the coverage aggregate:
>   - hadoop-bos
> ```
> 
> So the documented command (`mvn verify -Djacoco.skip=false ...`) fails at `hadoop-coverage` before `report-aggregate` ever runs, until `hadoop-bos` is added as a `<dependency>` in `hadoop-coverage/pom.xml`.
> 
> ## Bug 2 (silent coverage gap): `hadoop-bos` surefire argLine missing `@{argLine}`
> Even after Bug 1 is fixed, `hadoop-cloud-storage-project/hadoop-bos/pom.xml:265` overrides surefire's `<argLine>` without `@{argLine}`:
> 
> ```
> <argLine>${maven-surefire-plugin.argLine}</argLine>
> ```
> 
> A module-level `<argLine>` fully replaces the parent's, so the `@{argLine}` token that `jacoco:prepare-agent` substitutes to inject the agent is dropped. The forked test JVMs for `hadoop-bos` run without the JaCoCo agent, producing no exec data, so its coverage shows as empty in the aggregate report, silently. The guard script can't catch this; it only checks deps/allowlist presence.
> 
> Of the 21 surefire/failsafe `<argLine>` overrides in the repo, this PR patched 20; `hadoop-bos` is the only one missed.
> 
> Fix for both:
> 
> ```
> <argLine>${maven-surefire-plugin.argLine} @{argLine}</argLine>
> ```
> 
> plus a `hadoop-bos` `<dependency>` entry in `hadoop-coverage/pom.xml`.
> 
> ## Minor
> The reorder of `maven-resources-plugin` after `exec-maven-plugin` in `pluginManagement` (`hadoop-project/pom.xml`) is cosmetic only; `pluginManagement` order has no effect in Maven.

Thanks @pan3793 — both are real, and they only surfaced once I rebased onto current trunk (hadoop-bos landed after this branch was originally cut). All three points are addressed in the latest push, which is rebased onto current trunk:

Bug 1 (build-breaking): Added hadoop-bos as a <dependency> in hadoop-coverage/pom.xml, so it's part of the aggregate and the guard passes — check-coverage-modules: OK (72 test-bearing modules accounted for).

Bug 2 (silent coverage gap): Appended @{argLine} to hadoop-bos/pom.xml's surefire <argLine>, matching every other argLine-overriding module, so jacoco:prepare-agent injects the agent. Confirmed on a small module that coverage is off by default and -Djacoco.skip=false produces target/jacoco.exec.

Minor (reorder): Agreed, pluginManagement order is irrelevant — I've restored trunk's ordering (maven-resources-plugin before exec-maven-plugin) so hadoop-project/pom.xml carries no gratuitous diff.

Separately, I extended check-coverage-modules.sh to catch the Bug 2 class going forward: it now fails the build at the validate phase if a test-bearing, non-allowlisted module overrides <argLine> without @{argLine} (verified it flags hadoop-bos when the token is removed). The structural fix that removes the footgun entirely — centralizing the argLine so a module can't drop @{argLine} — is tracked as a follow-up in HADOOP-19948.

hadoop-yetus:
:broken_heart: **-1 overall**






| +0 :ok: |  reexec  |   0m 23s |  |  Docker mode activated.  |
|||| _ Prechecks _ |
| +0 :ok: |  detsecrets  |   0m  0s |  |  detect-secrets was not available.  |
| +0 :ok: |  xmllint  |   0m  0s |  |  xmllint was not available.  |
| +0 :ok: |  shelldocs  |   0m  0s |  |  Shelldocs was not available.  |
| +1 :green_heart: |  test4tests  |   0m  0s |  |  The patch appears to include 2 new or modified test files.  |
| +0 :ok: |  mvndep  |   2m 27s |  |  Maven dependency ordering for branch  |
| +1 :green_heart: |  mvninstall  |  28m 42s |  |  trunk passed  |
| +1 :green_heart: |  compile  |   9m 28s |  |  trunk passed with JDK Ubuntu-21.0.11+10-1-24.04.2-Ubuntu  |
| +1 :green_heart: |  compile  |   9m 36s |  |  trunk passed with JDK Ubuntu-17.0.19+10-1-24.04.2-Ubuntu  |
| +1 :green_heart: |  mvnsite  |  11m  4s |  |  trunk passed  |
| +1 :green_heart: |  javadoc  |   6m 24s |  |  trunk passed with JDK Ubuntu-21.0.11+10-1-24.04.2-Ubuntu  |
| +1 :green_heart: |  javadoc  |   6m 17s |  |  trunk passed with JDK Ubuntu-17.0.19+10-1-24.04.2-Ubuntu  |
| +1 :green_heart: |  shadedclient  |  29m 55s |  |  branch has no errors when building and testing our client artifacts.  |
| -0 :warning: |  patch  |  30m 10s |  |  Used diff version of patch file. Binary files and potentially other changes not applied. Please rebase and squash commits if necessary.  |
| +0 :ok: |  mvndep  |   0m 26s |  |  Maven dependency ordering for patch  |
| +1 :green_heart: |  mvninstall  |  37m 17s |  |  the patch passed  |
| +1 :green_heart: |  compile  |  10m 11s |  |  the patch passed with JDK Ubuntu-21.0.11+10-1-24.04.2-Ubuntu  |
| +1 :green_heart: |  javac  |  10m 11s |  |  the patch passed  |
| +1 :green_heart: |  compile  |  11m  9s |  |  the patch passed with JDK Ubuntu-17.0.19+10-1-24.04.2-Ubuntu  |
| +1 :green_heart: |  javac  |  11m  9s |  |  the patch passed  |
| +1 :green_heart: |  mvnsite  |  13m 16s |  |  the patch passed  |
| +1 :green_heart: |  shellcheck  |   0m  1s |  |  No new issues.  |
| +1 :green_heart: |  javadoc  |   6m 17s |  |  the patch passed with JDK Ubuntu-21.0.11+10-1-24.04.2-Ubuntu  |
| +1 :green_heart: |  javadoc  |   6m 11s |  |  the patch passed with JDK Ubuntu-17.0.19+10-1-24.04.2-Ubuntu  |
| +1 :green_heart: |  shadedclient  |  30m 15s |  |  patch has no errors when building and testing our client artifacts.  |
|||| _ Other Tests _ |
| -1 :x: |  unit  |  22m 42s | [/patch-unit-root.txt](https://ci-hadoop.apache.org/job/hadoop-multibranch/job/PR-8606/9/artifact/out/patch-unit-root.txt) |  root in the patch failed.  |
| +1 :green_heart: |  asflicense  |   0m 45s |  |  The patch does not generate ASF License warnings.  |
|  |   | 220m 31s |  |  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | ClientAPI=1.55 ServerAPI=1.55 base: https://ci-hadoop.apache.org/job/hadoop-multibranch/job/PR-8606/9/artifact/out/Dockerfile |
| GITHUB PR | https://github.com/apache/hadoop/pull/8606 |
| Optional Tests | dupname asflicense codespell detsecrets compile javac javadoc mvninstall mvnsite unit shadedclient xmllint shellcheck shelldocs |
| uname | Linux e1a270aaa01b 5.15.0-179-generic #189-Ubuntu SMP Tue May 5 18:20:56 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | trunk / 0917f08eeba6c3207918b87dd8e9f4fcaf599f23 |
| Default Java | Ubuntu-17.0.19+10-1-24.04.2-Ubuntu |
| Multi-JDK versions | /usr/lib/jvm/java-21-openjdk-amd64:Ubuntu-21.0.11+10-1-24.04.2-Ubuntu /usr/lib/jvm/java-17-openjdk-amd64:Ubuntu-17.0.19+10-1-24.04.2-Ubuntu |
|  Test Results | https://ci-hadoop.apache.org/job/hadoop-multibranch/job/PR-8606/9/testReport/ |
| Max. process+thread count | 1284 (vs. ulimit of 10000) |
| modules | C: hadoop-project hadoop-maven-plugins hadoop-common-project/hadoop-common hadoop-common-project/hadoop-registry hadoop-hdfs-project/hadoop-hdfs hadoop-hdfs-project/hadoop-hdfs-native-client hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient hadoop-tools/hadoop-distcp hadoop-tools/hadoop-federation-balance hadoop-hdfs-project/hadoop-hdfs-rbf hadoop-yarn-project/hadoop-yarn/hadoop-yarn-applications hadoop-yarn-project/hadoop-yarn/hadoop-yarn-applications/hadoop-yarn-applications-catalog/hadoop-yarn-applications-catalog-webapp hadoop-tools/hadoop-azure hadoop-tools/hadoop-aws hadoop-cloud-storage-project/hadoop-bos hadoop-cloud-storage-project/hadoop-tos hadoop-cloud-storage-project/hadoop-gcp . hadoop-coverage U: . |
| Console output | https://ci-hadoop.apache.org/job/hadoop-multibranch/job/PR-8606/9/console |
| versions | git=2.43.0 maven=3.9.15 shellcheck=0.9.0 |
| Powered by | Apache Yetus 0.14.1 https://yetus.apache.org |


BUILDING.txt
```diff
@@ -149,7 +149,8 @@ Maven build goals:
  * Run checkstyle            : mvn compile checkstyle:checkstyle
  * Install JAR in M2 cache   : mvn install
  * Deploy JAR to Maven repo  : mvn deploy
- * Run clover                : mvn test -Pclover
+ * Run JaCoCo coverage       : mvn verify -Djacoco.skip=false -Dmaven.test.failure.ignore=true --fail-at-end
+                               (see 'Special plugins: JaCoCo code coverage' below)
```
pan3793:
```suggestion
```

pan3793 (approved):

hadoop-yetus:
:broken_heart: **-1 overall**






| +0 :ok: |  reexec  |   0m  0s |  |  Docker mode activated.  |
| -1 :x: |  patch  |   0m 18s |  |  https://github.com/apache/hadoop/pull/8606 does not apply to trunk. Rebase required? Wrong Branch? See https://cwiki.apache.org/confluence/display/HADOOP/How+To+Contribute for help.  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| GITHUB PR | https://github.com/apache/hadoop/pull/8606 |
| Console output | https://ci-hadoop.apache.org/job/hadoop-multibranch/job/PR-8606/10/console |
| versions | git=2.34.1 |
| Powered by | Apache Yetus 0.14.1 https://yetus.apache.org |


pan3793:
thanks, merged to trunk, and backported branch-3.5 (pruned hadoop-bos)

PR #8697, opened 2026-08-20
YARN-11845. WebAppProxy add Connection close header to prevent CLOSE_WAIT leaks

sohurdc:
# YARN-11845. WebAppProxy: add `Connection: close` header to prevent CLOSE_WAIT leaks

## Description of PR

This PR fixes an accumulation of sockets stuck in `CLOSE_WAIT` state on the
ResourceManager host that runs the WebAppProxy.

`WebAppProxyServlet.proxyLink()` creates a brand-new `HttpClient` for every
proxied request (`HttpClientBuilder.create()` → `build()`), sends the request,
and only calls `base.releaseConnection()` afterward. It never closes the
`HttpClient` itself.

Apache HttpClient enables connection pooling / keep-alive by default, so the
backend (Application Master or History Server) keeps the connection open after
returning the response, waiting to reuse it. But since the proxy throws away
the `HttpClient` right after each request, that pooled connection can never be
reused. The corresponding socket on the proxy side therefore lingers in
`CLOSE_WAIT` until the `HttpClient` is finally garbage collected.

The fix is to explicitly ask the backend to close the connection once the
response has been sent, by setting a `Connection: close` request header on the
outgoing request:

```java
base.setHeader("Connection", "close");
```

Because `Connection` is not in `PASS_THROUGH_HEADERS`, a client-supplied
`Connection: keep-alive` header is never forwarded, so the proxy always sends
`close` and the backend closes the connection as soon as the response is
written.

## How was this patch tested?

- Added `testWebAppProxyConnectionCloseHeader()` in
  `TestWebAppProxyServlet`, which verifies that the proxied backend receives a
  `Connection: close` header in three cases:
  1. a plain `GET` request,
  2. a `GET` request where the client explicitly sends `Connection: keep-alive`
     (asserting the proxy overrides it with `close`),
  3. a `PUT` request (asserting the header is also set on the PUT path).
- The existing `testWebAppProxyPassThroughHeaders` assertion (9 headers
  received by the backend) is unaffected: the `Connection` header was already
  among the counted headers, only its value changes from `Keep-Alive` to
  `close`.

## For code changes

- [ ] The title of this PR accurately describes the issue.
- [ ] The code follows the project's style guidelines.
- [ ] New tests are added, and existing tests still pass.
- [ ] No user-facing API/behavior change is introduced (the proxy continues to
      return the same responses; it only instructs the backend to close the
      underlying connection).

## Additional notes

- `Connection: close` does not change the response returned to the client; it
  only affects the lifetime of the internal proxied connection, so it has no
  user-visible impact.
- This is a one-line, low-risk change scoped to the web-proxy module.

## Commit message

```text
YARN-11845. WebAppProxy: add Connection: close header to prevent CLOSE_WAIT socket buildup. Contributed by weishao <weishao@sohu-inc.com>.

WebAppProxyServlet.proxyLink() creates a new HttpClient per request and never
closes it. With keep-alive enabled by default, the backend keeps the proxied
connection open and the proxy-side socket lingers in CLOSE_WAIT until GC.
Explicitly setting Connection: close makes the backend close the connection
after each response.
```

### DIFF
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-web-proxy/src/main/java/org/apache/hadoop/yarn/server/webproxy/WebAppProxyServlet.java
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-web-proxy/src/main/java/org/apache/hadoop/yarn/server/webproxy/WebAppProxyServlet.java
@@ -312,7 +312,7 @@ public class WebAppProxyServlet extends HttpServlet {
     }
     // Tell the AM/history server to close the connection after the response
     // so that the proxied connection is not left in CLOSE_WAIT state on the
-    // proxy side (SOHU-HADOOP-11).
+    // proxy side (YARN-11845).
     base.setHeader("Connection", "close");
     String user = req.getRemoteUser();
     if (user != null && !user.isEmpty()) {
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-web-proxy/src/main/java/org/apache/hadoop/yarn/server/webproxy/WebAppProxyServlet.java
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-web-proxy/src/main/java/org/apache/hadoop/yarn/server/webproxy/WebAppProxyServlet.java
@@ -310,7 +310,10 @@ public class WebAppProxyServlet extends HttpServlet {
         base.setHeader(name, value);
       }
     }
-
+    // Tell the AM/history server to close the connection after the response
+    // so that the proxied connection is not left in CLOSE_WAIT state on the
+    // proxy side (SOHU-HADOOP-11).
+    base.setHeader("Connection", "close");
     String user = req.getRemoteUser();
     if (user != null && !user.isEmpty()) {
       base.setHeader("Cookie",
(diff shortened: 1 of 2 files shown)


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-web-proxy/src/main/java/org/apache/hadoop/yarn/server/webproxy/WebAppProxyServlet.java
```diff
@@ -310,7 +310,10 @@ private void proxyLink(final HttpServletRequest req,
         base.setHeader(name, value);
       }
     }
-
+    // Tell the AM/history server to close the connection after the response
+    // so that the proxied connection is not left in CLOSE_WAIT state on the
+    // proxy side (YARN-11845).
+    base.setHeader("Connection", "close");
```
slfan1989, line 316: This also changes the client-facing connection behavior. 

`proxyLink()` copies every header from the upstream response into `resp`, so when the AM/history server responds with `Connection: close`, that hop-by-hop header is forwarded to the browser/API client as well.

I verified this by checking `proxyConn.getHeaderField("Connection")` in the new test, and the client-facing response contains `close`. 

Consequently, every proxied request loses keep-alive between the client and WebAppProxy, contrary to the PR description that only the internal connection is affected. RFC 9110 Section 7.6.1 also requires intermediaries to remove `Connection` and other connection-specific fields before forwarding.

Could we instead close the locally created `CloseableHttpClient/CloseableHttpResponse` with try-with-resources, which directly addresses the resource lifecycle? If the request header is retained, the response path should filter hop-by-hop headers and the test should verify that the client-facing connection is not forced closed.

PR #8742, opened 2026-09-16
YARN-10422: add python script for collecting diagnostic data

Hean-Chhinling:
### Description of PR
Upon agree, the application_failed & application hanging combines into "Application Diagnostic". This endpoint will be called by YARN UI2.

The scheduler_related_issue endpoint will be called by the new YARN QueueManager UI for capacity scheduler ([PR](https://github.com/brumi1024/yarn-scheduler-ui/pull/180/changes)).

The RM_NM_startup failures have been removed as it is not feasible to implement a UI for it in real case scenario.

Manually tested in High Availability environment cluster.

### For code changes:

- [ ] Does the title of this PR start with the corresponding JIRA issue id (e.g. 'HADOOP-17799. Your PR title ...')?
      required.*

### AI Tooling

If an AI tool was used:

### DIFF
--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/resources/diagnostics/diagnostics_collector.py
+++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/resources/diagnostics/diagnostics_collector.py
@@ -1,0 +1,489 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import print_function
+
+import argparse
+import sys, os
+import subprocess
+from urllib.request import urlopen, Request
+from datetime import datetime, timedelta
+from urllib import request, error
+import xml.etree.ElementTree as ET
+import re
+import time
+
+TEMP_DIR = "/tmp"
+HADOOP_CONF_DIR = "/etc/hadoop"
+YARN_SITE_XML = "yarn-site.xml"
+MAPRED_SITE_XML = "mapred-site.xml"
+RM_ADDRESS_PROPERTY_NAME = "yarn.resourcemanager.webapp.address"
+JHS_ADDRESS_PROPERTY_NAME = "mapreduce.jobhistory.webapp.address"
+
+RM_LOG_REGEX = r"(?<=\")\/logs.+?RESOURCEMANAGER.+?(?=\")"
+NM_LOG_REGEX = r"(?<=\")\/logs.+?NODEMANAGER.+?(?=\")"
+INPUT_TIME_FORMAT = '%a %b %d %H:%M:%S %Z %Y'  # e.g. Wed May 28 07:35:39 UTC 2025
+OUTPUT_TIME_FORMAT = '%Y-%m-%d %H:%M:%S,%f'    # e.g. 2025-05-28 11:57:05,435
+OUTPUT_TIME_FORMAT_WITHOUT_SECOND = '%Y-%m-%d %H:%M'  # e.g. 2025-05-28 11:57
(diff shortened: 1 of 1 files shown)


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/resources/diagnostics/diagnostics_collector.py
```diff
@@ -0,0 +1,358 @@
...
+    Multiple Jstack of ResourceManager
+    YARN and Scheduler Configuration
+    Cluster Scheduler API /ws/v1/cluster/scheduler and Cluster Nodes API /ws/v1/cluster/nodes response
+    Scheduler Activities /ws/v1/cluster/scheduler/bulk-activities response
+    """
+    logger.info("Collecting scheduler-related diagnostics")
+    output_path = _create_output_dir(
+        os.path.join(_TEMP_DIR, "scheduler_related_issue" + str(time.time()).split(".")[0]))
+
+    rm_jstack = _create_request(_rm_url("ws/v1/node/jstack/{}".format(_NUMBER_OF_JSTACK)), False)
```
slfan1989: Could we use endpoints that exist in the target Hadoop version, or explicitly identify the prerequisite changes that introduce these endpoints?

I could not find implementations for `/ws/v1/node/jstack/{count}` or `/ws/v1/node/apps/{appId}/jstack/{count}` in the PR's base revision. 

The existing daemon thread-dump endpoint is `/stacks`.  The same issue applies to the application and NodeManager requests above.

Additionally, curl does not currently fail on HTTP error responses, so a 404
response could be saved as a successful `*_jstack` result. 

Please validate the HTTP status and report unavailable diagnostics explicitly.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/resources/diagnostics/diagnostics_collector.py
```diff
@@ -0,0 +1,358 @@
...
+    scheduler_config = _create_request(_rm_url("ws/v1/cluster/scheduler-conf"))
+    _write_output(output_path, "scheduler_configuration", scheduler_config)
+
+    yarn_conf = _run_command("cat", os.path.join(_HADOOP_CONF_DIR, _YARN_SITE_XML))
+    _write_output(output_path, "yarn_site", yarn_conf)
+
+    enable_debug_log = _set_rm_scheduler_log_level("DEBUG")
+    logger.info("Set RM scheduler log level to DEBUG: %s", enable_debug_log)
+    log_address = _get_node_log_address(_RM_ADDRESS, _RM_LOG_REGEX, _NODE_SCHEME)
+    start_time, end_time = (_format_datetime_no_seconds(datetime.now() - timedelta(seconds=120)),
```
slfan1989: This collects the previous two minutes immediately after enabling DEBUG, rather than collecting two minutes of DEBUG output.

Unless DEBUG was already enabled, most of this interval will not contain the scheduler diagnostics that this operation intends to capture.

Could we record the start time when DEBUG is enabled, wait for the collection interval, and then retrieve that interval's logs?  Please also restore the previous log level in a `finally` block instead of unconditionally setting INFO, so failures do not leave DEBUG enabled or overwrite an existing setting.

slfan1989 (commented):
Thanks for adding the diagnostics collector. I found three functional issues
that should be addressed before merging:

- The jstack requests target endpoints that are not present in the base revision.
- The scheduler DEBUG collection reads the preceding two minutes instead of
  capturing a two-minute interval after enabling DEBUG.
- Applications without an AM container cause collection to abort before the
  remaining diagnostics are saved.

Please add focused tests covering these cases, including missing AM addresses
and HTTP error responses.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/resources/diagnostics/diagnostics_collector.py
```diff
@@ -0,0 +1,358 @@
...
+    """
+    if args.arguments is None or len(args.arguments) == 0:
+        logger.error("Missing application or job id")
+        sys.exit(os.EX_USAGE)
+
+    app_id = args.arguments[0]
+    logger.info("Collecting application diagnostics for %s", app_id)
+    output_path = _create_output_dir(os.path.join(_TEMP_DIR, app_id))
+
+    nm_address = _get_nodemanager_address(app_id)
```
slfan1989: Could we handle applications that do not yet have an AM container?

For example, an application stuck in ACCEPTED while waiting for resources may not have `amHostHttpAddress`. `_get_nodemanager_address()` unconditionally accesses `.find("amHostHttpAddress").text`, so this case raises AttributeError before application information, attempts, or RM logs are collected.

This is an important case for application-hang diagnostics. 
Please save the application information first, skip NodeManager-specific collection when the
address is unavailable, and continue collecting the remaining diagnostics.

PR #8747, opened 2026-09-18
YARN-10932. Fixed deadlock when using legacy dynamic queues

p-szucs:
### Description of PR
**Fix possible deadlock when using legacy dynamic queues**

The deadlock happens when we are using ManagedParentQueues with periodically calling the scheduling monitor, that uses GuaranteedOrZeroCapacityOverTimePolicy. In this case a refresQueues operation can collide with this thread if the timing is unfortunate, because both operations lock the policy and the ManagedParentQueue as well, but in a different order.

**SchedulingMonitor:**

- locks the policy in GuaranteedOrZeroCapacityOverTimePolicy's updateLeafQueueState()
- then iterates through the child queues, locking the ManagedParentQueue

**RefreshQueues operation:**

- locks the queue by calling ManagedParentQueue.reinitialize
- locks the policy by calling GuaranteedOrZeroCapacityOverTimePolicy.computeQueueManagementChanges - updateTemplateAbsoluteCapacities

Because of the reversed order of the locks, with an unfortunate timing it can happen that both threads can acquire their first lock at the same time, and after that they are waiting on each other to release those. Any other threads that walks through the same code path will be blocked in this case, like getting the scheduler info, and also container allocations.

The fix changes the order of the locking with adding a new lock in the queue management policy, so both threads locks the queue first, then the policy, so the deadlock can not happen.

Contains content generated by Claude

Tested in a cluster with a ManagedParentQueue in the queue structure, and frequent scheduling monitoring, also requesting queue configuration changes frequently. This way I could reproduce the deadlock consistently.

During my testing the deadlock could not be reproducible with the fix.

### For code changes:

- [x] Does the title of this PR start with the corresponding JIRA issue id (e.g. 'HADOOP-17799. Your PR title ...')?
      required.*

### AI Tooling

If an AI tool was used:

- [x] The PR includes the phrase "Contains content generated by <tool>"


hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/capacity/QueueManagementDynamicEditPolicy.java
```diff
@@ -191,7 +191,12 @@ List<QueueManagementChange> manageAutoCreatedLeafQueues()
       try {
         startTime = clock.getTime();
 
-        queueManagementChanges = policyClazz.computeQueueManagementChanges();
+        parentQueue.getReadLock().lock();
+        try {
+          queueManagementChanges = policyClazz.computeQueueManagementChanges();
```
K0K0V0K, line 196: NIT:

Would not be better to apply the read lock at the beginning of  computeQueueManagementChanges method?
Cause `if (!parentQueue.shouldFailAutoCreationWhenGuaranteedCapacityExceeded()) {` called in line 186, so some unexpected behaviour can occur if that change during the `computeQueueManagementChanges` run.

K0K0V0K (commented):
Thanks @p-szucs !

Overall seems good to me, just one NIT.

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/capacity/QueueManagementDynamicEditPolicy.java
p-szucs: Thanks for the review @K0K0V0K!
I agree with your point, my intention was to lock and serialize as minimal code part as I can to minimize possible latency, but with the first version it could be possible that we are working with partly inconsistent data (however I think as the scheduler changes are done asynchronously, a later validation should catch them). But I think it is a cleaner approach to lock the queue first, and then start reading its properties/make computations. The rest of the method that is locked now is only detaching an asynchronous event to make the computed changes and debug logging

hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/scheduler/capacity/QueueManagementDynamicEditPolicy.java
```diff
@@ -183,47 +183,52 @@ List<QueueManagementChange> manageAutoCreatedLeafQueues()
...
-              new QueueManagementChangeEvent(parentQueue,
-                  queueManagementChanges);
-          scheduler.getRMContext().getDispatcher().getEventHandler().handle(
-              queueManagementChangeEvent);
-        }
-
-        if (LOG.isDebugEnabled()) {
-          LOG.debug("{} uses {} millisecond" + " to run",
-              policyClazz.getClass().getName(), clock.getTime() - startTime);
+    parentQueue.getReadLock().lock();
```
slfan1989: This read lock introduces a read-to-write lock upgrade deadlock.

`GuaranteedOrZeroCapacityOverTimePolicy.computeQueueManagementChanges()` calls
`managedParentQueue.updateTemplateCapacitiesForAbsoluteResource()`, which unconditionally acquires the same parent's write lock before checking the capacity mode.

Since `ReentrantReadWriteLock` does not support upgrading a held read lock to the write lock, the monitor thread will block on itself whenever it reaches this path, even without a concurrent queue refresh.

Please revise the locking strategy to accommodate this write-lock acquisition and add a bounded regression test exercising the real policy through `QueueManagementDynamicEditPolicy`.

